PR #101, opened 2014-07-15
Support move & resize with s-mouse-{1,3}

kanru:
I really like this way to move window around in awesome. Hold the super key then use mouse button-1 on any window to move it, button-3 to resize it.


LaloHao:
Is it possible to map the mouse on tile groups to avoid using xdotool?

PR #146, opened 2014-09-16
Split the floating group into off into its own package

jorams:
This is an attempt to commence organizing StumpWM into smaller parts, as discussed in #125.

I'm not sure if this should be merged as-is, but I'm curious to hear your thoughts. Splitting some parts off into their own package might be a good way to organize things a bit better. Maybe this will also involve finding some higher level abstractions in some places.

I chose to start with the floating group because it's a relatively uninvolved element of the system. Other things, like the tiling group, spread out a lot farther through different parts.

### DIFF
--- a/core.lisp
+++ b/core.lisp
@@ -25,6 +25,8 @@
 
 (in-package :stumpwm)
 
+(export '(grab-pointer ungrab-pointer))
+
 ;; Wow, is there an easier way to do this?
 (defmacro def-thing-attr-macro (thing hash-slot)
   (let ((attr (gensym "ATTR"))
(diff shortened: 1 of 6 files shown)


dbjergaard:
Hi @jorams,

I'd like to merge this.  I think this is an appropriate refactoring for the codebase.  Let me know if you want to resubmit a new pull request or if I should fix any merge conflicts and merge this one.  Somewhere (the reference escapes me) I made a list of things I would like re-packaged.  

There's no reason we can't slowly refactor bits of stumpwm into separate libraries that are all shipped from the main stumpwm repo.

jorams:
I think fixing any merge conflicts and then merging this one would be ok.

dbjergaard:
OK, I'll do this when I get a moment. 

Cheers,

    Dave

PR #181, opened 2015-01-27
Jorams split floating group

dbjergaard:


dbjergaard:
duplicate of #146 for travis build before pulling to master.

dbjergaard:
@jorams for some reason this PR breaks the manual generation. Did I merge something incorrectly?

jorams:
@dbjergaard Hmm, no. I think the manual generation breaks because it's looking for functions `gnew-float` and `gnewbg-float` in the `STUMPWM` package. Those functions, however, have moved. They should probably be exported from `STUMPWM` in some way.

I have started (slowly) splitting StumpWM into a package per file (which would also result in the changes introduced in this PR). While I would like to say this PR can be closed until that is done, I'm not sure if I'll finish that process anytime soon (it's a lot of work). Fixing the manual generation for this PR is probably a good idea.

Also, the changes this introduces in core.lisp shouldn't be there.

dbjergaard:
Hi @jorams,

I like the one package per file approach.  Will it be possible to refactor functionality as well? Some of the gui bits of stumpwm have broader use, especially for projects that don't want to pull in McClim. 

As far as merging the results into the stumpwm codebase, I would prefer it happen piecemeal.  If you open a PR every time you define a package in a file, then I can merge it and deal with any issues that come up then, rather than doing it all at once in one big PR.

dbjergaard:
@jorams I fixed the travis build, though its not very elegant.  Let me know what you want me to do with this (merge or close it).

PR #184, opened 2015-02-04
Add ~/.stumpwm.d/init.lisp as an init file.

etrumbo:
Now that we're using ~/.stumpwm.d for modules and data, I'd like the option to put StmpWM's init file in that directory. I've added "~/.stumpwm.d/init.lisp" as an init file StumpWM will search for. The traditional ~/.stumpwmrc continues to have the highest priority, followed by this init.lisp, then the xdg and system-wide versions.


jorams:
:+1:

etrumbo:
That was fast! Thank you!

PR #189, opened 2015-02-21
Fix #188: Missing arguments in manual when built with CCL.

### LINKED ISSUE #188: CCL builds incomplete stumpwm.info manual
At various times, I've built and run StumpWM using SBCL, CLISP, and Clozure CL (CCL), all in current 64-bit Linux versions. I've noticed a strange problem with stumpwm.info when built with CCL -- every function listing fails to list the parameters, replacing them with "(Check the code for args list)" throughout. SBCL and CLISP both pull the relevant information into the Info doc correctly.

Versions:
CCL for 64-bit Linux, Subversion revision 16352.
Texinfo 4.13 (Not current, but I've heard there are performance regressions with 5.x)
Slackware Linux, 64-bit -current

etrumbo:
Adds complete support for Clozure CL (CCL) to manual.lisp. Specifically, CCL wasn't extracting arglists from StumpWM source to fill out the manual references. It will now. ;)

### DIFF
--- a/manual.lisp
+++ b/manual.lisp
@@ -47,8 +47,9 @@
                                         name
                                         #+sbcl (sb-introspect:function-lambda-list fn)
                                         #+clisp (ext:arglist fn)
+                                        #+ccl (ccl:arglist fn)
                                         #+lispworks (lw:function-lambda-list fn)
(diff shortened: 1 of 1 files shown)


jorams:
:+1:

PR #190, opened 2015-02-22
Use >= 24 functions for sbcl version of "-dirty"

bougyman:
Current sbcl-1.2.8, packaged in voidlinux returns the string "-dirty" for lisp-implementation-version.

### DIFF
--- a/wrappers.lisp
+++ b/wrappers.lisp
@@ -279,8 +279,8 @@ they should be windows. So use this function to make a window out of them."
   ;; FIXME: seems like there ought to be a less cumbersome way to run
   ;; different code based on the version.
   #+sbcl (macrolet ((dir (p)
-                      (if (>= (parse-integer (third (split-seq (lisp-implementation-version) '(#\.))) :junk-allowed t)
-                              24)
+                      (if (or (equal (lisp-implementation-version) "-dirty") (>= (parse-integer (third (split-seq (lisp-implementation-version) '(#\.))) :junk-allowed t)
+                              24))
                           `(directory ,p :resolve-symlinks nil)
                           `(directory ,p))))
            (dir pathspec))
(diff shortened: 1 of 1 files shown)


bougyman:
Found https://twitter.com/rainerjoswig/status/561648161631260672, so it's not just me.

PuercoPop:
The logic for checking the version is wrong, it is checking if the patch number is more than 24. From the commits e37f4fc8 and e870e79d we can infer that the version in question was 1.0.24. Which imho it is so old that the keyword argument :resolve-symlinks should abe assumed to be present. and the code be simplified to:

```lisp
....
#+sbcl (directory pathspec :resolve-symlinks nil)
...
```

However if we are to continue to check the version version we should do so properly, taking heed of the fixme comment. The version should be checked with either, ```sb-ext:assert-version-<=``` or with ```uiop/utility:version<```. Something like ```(uiop/utility:version< "1.0.24" (lisp-implementation-version))``` which probably be cleaner than wrapping a handler around assert-version-<=.

my 2c,

bougyman:
Problem is both of those would still fail to build on sbcl.
(uiop/utility:version< (lisp-implementation-version) "1.0.24") returns T for the string "-dirty", which would take the wrong codepath. Removing the implementation check would be better, in my opinion, but if we want to support < 24 _and_ 1.2.8 (which has a broken generate_versions.sh that causes the "-dirty"), we need to do an exact string match of that "-dirty" and act accordingly.

PuercoPop:
And the problem with the previous solution is that is doesn't work as intended. It expands to ```(directory pathspec)``` instead of the intended ```(directory pathspec :resolve-symlinks nil)```. As all it does is check if the patch number. Having lisp-implementation-version return "dirty" instead of 1.2.8-dirty (which assert-version->= handles without problem) is a bug in SBCL, which is already fixed: https://github.com/sbcl/sbcl/commit/21147ec04d92f6b8c0f320ea6df6fc38069ac671. We shouldn't patch downstream to circumvent bugs upstream. 

But your PR has shed light on a bug in StumpWM, which we could fix by just eliminating the version check altogether as I don't think any user is using an sbcl version that is +5 years old. Support for the :resolve-symlinks keyword argument in #'directory can be safely assumed.

dbjergaard:
Addressed in #192

PR #195, opened 2015-03-28
get-frame-number-translation: Rewrite for clarity

PuercoPop:
No need to use or/and when if does the same job

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -590,10 +590,11 @@ supported. By default, the frame labels are the 36 (lower-case)
 alphanumeric characters, starting with numbers 0-9.")
 
 (defun get-frame-number-translation (frame)
-  "Given a frame return its number translation using *frame-number-map* as a char."
+  "Given a frame return its number translation using *frame-number-map* as a
+char."
   (let ((num (frame-number frame)))
-    (or (and (< num (length *frame-number-map*))
-             (char *frame-number-map* num))
+    (if (< num (length *frame-number-map*))
+        (char *frame-number-map* num)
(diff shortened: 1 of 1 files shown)


primitives.lisp
```diff
@@ -590,12 +590,13 @@ supported. By default, the frame labels are the 36 (lower-case)
...
+char."
   (let ((num (frame-number frame)))
-    (or (and (< num (length *frame-number-map*))
-             (char *frame-number-map* num))
-        ;; translate the frame number to a char. FIXME: it loops after 9
-        (char (prin1-to-string num) 0))))
+    (if  (< num (length *frame-number-map*))
+         (char *frame-number-map* num)
+         ;; translate the frame number to a char. FIXME: it loops after 9
+         (char (prin1-to-string num) 0))))
```
jorams: There's an extra space after the `if`, removing that would clean up the diff a bit.

PuercoPop: fix'd

jorams:
:+1:

jorams:
The build is currently failing because http://www.rabbitmq.com is down. I'll try to remember to restart it in a few hours.

dbjergaard:
Last I saw this commit passed the build, the latest whitespace changes "broke" it because it is down, as long as our "build passing" tag on the main page is good (and it is AFAIK) I'm happy.

jorams:
@dbjergaard rabbitmq is part of the standard Travis container, or something. I saw a bunch of other Travis builds fail because of it.

I restarted the build a few hours later and then succeeded, so everything is good.

PR #199, opened 2015-04-23
Allow non ascii chars to be typed in window-send-string.

kisp:
This simple change allows me to type German Umlauts with `window-send-string`, after doing `setxkbmap de`.


kisp:
I don't quite understand, why the clisp build fails...

dbjergaard:
very excited about this, I will look into why the build is failing

jorams:
For some reason Clisp seems to have the default package be `COMMON-LISP`, causing the load of https://github.com/sharplispers/clx/blob/ecce3edcc158cd0ad65bc04f9c7c7a8b55eba9d1/demo/clx-demos.lisp to fail because its `:export` list uses plain symbols.

I'm seeing the same issue in another build, but I haven't yet figured out why its happening.

PR #201, opened 2015-05-05
Fix fullscreen mode with floating windows

shamazmazum:
When a float-window has width or height equal to width or height of the screen, its last-width or last-height slot is not being properly updated when you go to the fullscreen mode.

So If you have, for example, a window with sizes 1280x720 on the screen with sizes 1280x1024, you will never return the window to normal size (and get BadValue error from x server).

Solution: remove unnecessary (as they seem to be) 'unless' conditions in accessor methods for window geometry slots.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -18,20 +18,16 @@
 
 ;; some book keeping functions
 (defmethod (setf window-x) :before (val (window float-window))
-  (unless (eql (window-x window) val)
-    (setf (float-window-last-x window) (window-x window))))
+  (setf (float-window-last-x window) (window-x window)))
 
 (defmethod (setf window-y) :before (val (window float-window))
-  (unless (eql (window-y window) val)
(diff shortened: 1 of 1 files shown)


dbjergaard:
Hi, I doubt the unless conditions are unnecessary.  They were likely added to fix a different issue that is getting in the way of this one.  If you can, it would be worth looking to see when the unless conditions you removed were added and see if they addressed a different bug.

shamazmazum:
These conditions were added in commit https://github.com/stumpwm/stumpwm/commit/dfe272361ac8a520aec56e0f7bc8e342ed4eaf4b

I don't know how to reproduce the error which this commit fixes (something about :configure-request event), but 'unless' conditions were added with accessors as a whole. Also I don't see if LAST-* slots are used anywhere else in the code. The only place they are, is (setf window-fullscreen). So I think, there is no side-effects of removing the conditions.

PR #206, opened 2015-06-30
Rework the API of for manipulating *module-dir*

PuercoPop:
Hi, the reasoning behind this change is that I was recently surprised when ```(setf *module-dir* ...)``` didn't work as expected in spite of ```*module-dir*``` being a public symbol in the stumpwm package. After reading module.lisp I saw the I was misusing the API and should call set-config-dir instead. However a more idiomatic API would be for a "setter/getter" and keeping ```*module-dir*``` private.

```
(module-dir) ;; getter

(setf (module-dir) #P"/path/to/contrib/dir") ;; setter
```

However I saw the module API is being reworked so I know this is probably not going to be merged. I would like to suggest a different way to handle the module API that the currently proposed upgrade path that leverages ASDF extensibilty

We can define a class stumpwm-module that inherits from asdf:system and just use asdf search path to load-modules. We could then define a new operation *load-module* that only works on *stumpwm-module*s or just use asdf's load-op (my prefered choice) and just use the stumpw-module information for ```(list-modules)```. The manual would only need to have a tldr of how to configure [asdf's search path](https://common-lisp.net/project/asdf/asdf.html#Controlling-where-ASDF-searches-for-systems)

Let me know what y'all think, If deemed a good idea I could work on a tentative implementation over the weekend.

### DIFF
--- a/module.lisp
+++ b/module.lisp
@@ -28,16 +28,13 @@
 (export '(load-module
           list-modules
           *load-path*
-          *module-dir*
+          module-dir
           init-load-path
-	  set-module-dir
           find-module
           add-to-load-path))
 
 (defvar *module-dir* (pathname-as-directory (concat (getenv "HOME") ".stumpwm.d/modules"))
   "The location of the contrib modules on your system.")
-
-
 
 (defun build-load-path (path)
   "Maps subdirectories of path, returning a list of all subdirs in the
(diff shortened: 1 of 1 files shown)


dbjergaard:
Hi,

Thanks for the pull request! As you've discovered, interacting with module-dir is not the correct way to acheive what you wanted.  In fact, module-dir isn't supposed to be called by the user at all. It set at compile time as an initial place to look for libraries. Its expected that users instead use *load-path* and its helper functions.  

If I'm following your suggestion about using asdf I like it, but it looks like it would take re-writing all of the current modules to use the stumpwm-module class.  AFAIK the current module code leverages asdf's search path in order to load the code.  

Maybe you could have a second look at module.lisp and see if there were anyway to make the code clearer? I think many of your suggestions are already implemented, they're just not exposed in an obvious way.

PuercoPop:
The code currently the code syncs `*load-path*` with `asdf:*central-registry*` which is there for backwards comparability only. It doesn't use the source-registry. 

I'm proposing we drop the whole concept of *module-dir*/load-path and just use asdf to load modules. So the resulting API would be just `load-module`, `list-modules` and `find-module`, with find-module and load-module being thin wrappers over asdf. We could offload most of the code to asdf.

From the user perspective they would just need to ensure the directory is in asdf's search path. For example by creating a file `50-stumpwm-modules.lisp` to `~/.config/common-lisp/source-registry.conf.d/` with `(:tree "/home/foo/.stumpwm.d/modules/")` to ensure that modules are loadable.

Yes, all the module system definitions would have to be updated. (I can take care of that as well).

dbjergaard:
hmm, if your willing to put in the work and it simplifies the code, go for it!  Also, did you want to merge your other PRs? It looks like you closed them all...

PuercoPop:
I closed all but #211 because they are all included in that PR.

It does simplify the code imho, but currently (list-modules) doesn't quite work*. Here is the [code](https://github.com/PuercoPop/stumpwm/tree/module-system) I worked over the weekend so you can see the main idea but I still need to figure out how to make `#'list-modules` work properly.

* It only lists 'loaded' systems found in `*defined-systems*`


P.D. A new module system definition would look like this
```lisp
(asdf:defsystem #:amixer
  :name "amixer"
  :class :stumpwm-module
  :defsystem-depends-on (:stumpwm-module-system)
  :serial t
  :description "Describe amixer here"
  :author "Amy Templeton, Jonathan Moore Liles, Ivy Foster"
  :license "GPL v3"
  :depends-on (#:stumpwm)
  :components ((:file "package")
               (:file "amixer")))
```

dbjergaard:
I'll merge this original PR, but it breaks make-image.lisp for ecl which requires set-module-dir.  It shouldn't take too much to fix this, I'm just writing this as a note to self so I don't forget this when I merge.

dbjergaard:
Please feel free to re-open if you can address the above issues and still want the code merged.

PR #210, opened 2015-07-04
Represent Newline using the character.

PuercoPop:
A small style fix. Use (string #\Newline) instead of using a literal newline inside a string

and use the default value when the default matches what was passed to `#'split-string`


PuercoPop:
On second thought it seems that split-string can be replaced with [split-sequence](http://cliki.net/SPLIT-SEQUENCE). Thoughts?

PR #211, opened 2015-07-04
Fix lookup-symbol and misc changes

PuercoPop:


dbjergaard:
Hi,
This doesn't build on my system:
```
debugger invoked on a ASDF/FIND-COMPONENT:MISSING-DEPENDENCY in thread
#<THREAD "main thread" RUNNING {10039CE723}>:
  Component #:SPLIT-SEQUENCE not found, required by #<SYSTEM "stumpwm">
```

dbjergaard:
Actually, part of stumpwm's design philosophy is minimal external dependence. If this adds an addition dependency, please rework the patch to avoid using split-sequence. 

Also, please keep PR for different issues separate, it makes it easier for me to track them mentally :)

dbjergaard:
Could you comment please? I get nervous when PRs are open for a long time :)

dbjergaard:
Hi, since I haven't heard from you I pulled the parts of this PR that don't involve split-sequence. If you want to re-work those to not use split-sequence (ie leave split-string and re-work lookup-symbol) then I'm happy to take a new PR.  For now, in stumpwm 1.0, I don't want to introduce new dependencies even if they are perfectly reasonable. I've also added your name to the authors list.  

Cheers,

David

jorams:
Making `command-docstring-warning` a `warning` instead of a `style-warning` causes ASDF compilation of anything without a docstring to fail. I don't think that's desirable, as for example the MPD module contains a ton of commands without a docstring.

I also think it makes more sense for it to be a style warning, because not having a docstring won't ever cause problems.

dbjergaard:
agreed and reverted

PR #214, opened 2015-08-04
Fix default value of *module-dir*.

alezost:
This typo wasn't noticed because `*module-dir*` is set by `make-image.lisp` to a proper value.  Also I added another commit to remove trailing whitespaces.  Is it OK?


dbjergaard:
AHH the white space noise! My eyes! Thanks for the PR!

PR #225, opened 2015-09-23
Sane behavior of the sloppy mouse focus policy in floating groups.

aoterodelaroza:
The behavior of *mouse-focus-policy* = :sloppy in floating groups is:
any window that gets the focus (e.g. by hovering over it with the
mouse) automatically raises. This is annoying and makes using floating
groups very difficult. This patch makes floating group windows grab
the focus but stay in the background.  Windows raise only when
clicked.

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -77,8 +77,8 @@
 
 (defgeneric update-decoration (window)
   (:documentation "Update the window decoration."))
-(defgeneric focus-window (window)
-  (:documentation "Give the specified window keyboard focus."))
+(defgeneric focus-window (window &optional raise)
+  (:documentation "Give the specified window keyboard focus and (optionally) raise."))
 (defgeneric raise-window (window)
   (:documentation "Bring the window to the top of the window stack."))
 (defgeneric window-visible-p (window)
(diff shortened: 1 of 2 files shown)


window.lisp
```diff
@@ -908,20 +908,21 @@ needed."
...
-  "Make the window visible and give it keyboard focus."
+(defmethod focus-window (window &optional (raise t))
+  "Make the window visible and give it keyboard focus. If raise is t, raise the window."
   (dformat 3 "focus-window: ~s~%" window)
   (let* ((group (window-group window))
          (screen (group-screen group))
          (cw (screen-focus screen))
          (xwin (window-xwin window)))
+    (when raise
+      (raise-window window))
```
jorams: Are we sure that raising a window if it's already focused doesn't cause issues anywhere?

dbjergaard: We could (should?) make (raise nil) so that the old behavior is the same, then callers in floating-group will have to call with the optional argument non-nil.

aoterodelaroza: @jorams: I've been using it since yesterday with :sloppy, and couldn't see any problems. I also checked the behavior of :click and :ignore briefly. Can you suggest a test?

@dbjergaard: do you mean making the optional argument in focus-window nil instead of t?

dbjergaard:
Is there a good reason not to flip the default condition to nil and call focus-window with t when you want to raise it as well? Checking the code base, focus-window is used in quite a few contexts and it would make me feel better if the default way of calling didn't change. If you can make that change I'll happily merge this PR.

aoterodelaroza:
The behavior of focus-window before this commit was to raise the window unless the window was already focused. Hence, the default value of t is actually closer to the previous behavior. The only difference (other than in floating groups) is that now it raises a window that is already focused. Could that be a problem? 

If I flip the default value to nil, then newly-focused windows in tiled groups won't raise. I need to test it a little bit to see if this will break things.

Another option is to leave the raise-window where they were and apply the condition only when the new window is not focused. That would reproduce the previous behavior exactly (with default value t) and would work for floating groups. But it seems to me a bit inelegant to have a raise option in the defun and then ignore for some particular cases, no?

aoterodelaroza:
After trying for a few minutes, it seems using nil as the default work. Whichever you think is more reasonable.

dbjergaard:
you were so convincing in your previous post that now I don't care either way. I'll merge this and if it breaks it for someone they'll open an issue.

Issue #187, opened 2015-02-19
Sometimes I lose control


Trevoke:
Sometimes, I'm not sure what I do, but I end up unable to call out to StumpWM.

Just now, I started emacs and tried a Ctrl+Click, and now I can't send commands to either emacs or stumpwm (as far as I can tell).

Thankfully, through clicking, I was able to enable to find emacs' menu and quit emacs that way, which restored responsiveness for stumpwm.

How do I recover from such a situation?

dbjergaard:
I chuckled at the title of this issue, but to help I'll need more info.
Can you give specific instructions to reproduce with an empty init file?
That will rule out a lot of potential sources.

Also, is this just with emacs or does it appear in other programs? Does
it happen when there are multiple frames or heads, or does it happen
when emacs is the only focused window on a screen? Do you have the mouse
focus policy set to sloppy? 

Cheers,

    Dave

Trevoke:
I'll do my best to try to reproduce it.

I have not configured StumpWM at all - I've just started my way through the manual and I'm up to the introduction to frames.
This happened on a laptop, not connected to anything. I think, both times, there were two frames split horizontally (as in they were side by side, not one on top of the other).

It happened most recently with emacs, and emacs was the left frame.

I don't know how to set the mouse focus policy, so I'd guess .. Whatever the default is.

I'm excited to use StumpWM since it'll give me a chance to dive into Common Lisp (my lisp usage has been limited to emacs-lisp so far), so I'm eager to help as much as I can. Nothing like a complete beginner to root out crazy edge cases..

dbjergaard:
Hmm, to be clear: When you have emacs in a frame and you press C-t (the
default prefix key), the cursor doesn't change shapes? C-t C-g should be
enough to change the cursor and change it back.  

If your talking about emacs focus, its possible that somehow emacs is
being raised in a frame that isn't focused and since it doesn't have
focus you won't be able to interact with emacs. Then when you clicked on
emacs, the frame was refocused and you could quit emacs.  I'm shooting
in the dark here though.

Just remember: by default you have to select the frames yourself.  If
you set the mouse policy to sloppy, the focus will follow your mouse
which may feel more familiar until you integrate with stumpwm's ideas
for handling focus.

Cheers,

    Dave

Trevoke:
When I'm in emacs, normally, I can do `C-t C-g` and see the cursor change and change back.

[...]

Issue #94, opened 2014-06-02
Display bug when switching to empty groups


ghferrari:
Hi all,

I've been experiencing the following bug for a good while (perhaps a year, perhaps longer), but it never used to happen.

How to reproduce: with two or more groups open, open an application in one group and then switch to an empty group (i.e. a group with no application windows). In the distant past, this used to result in StumpWM showing an empty desktop as I would expect. Now, instead of that, the mode-line updates to indicate the new empty group but the application window from the previous group remains displayed.

I'm using the latest stumpwm with sbcl and quicklisp on debian.

Thanks for your help,

Geoffrey

dbjergaard:
Hi, I was just re-reading this.  Are you still affected by this bug? I have some better ideas of what may be causing this.  It may be as simple as re-drawing the root window when entering a new group. Unfortunately I haven't been able to reproduce this consistently.

ghferrari:
Hi David,

Yes, this is still a problem for me. As far as know, it should be
reproducible on a stock debian testing with sbcl and quick lisp.

Hope you can get it sorted!

Geoff

dbjergaard:
Ok, I was thinking of switching back to debian anyway, so maybe I'll be able to expose it then.  As you can see this will be fixed by the next release :)

jtbm37:
This is happening to me as well (with minimal config). I installed StumpWM a week ago or so.

```
> uname -a
Linux mycomputer 3.19.0-25-generic #26-Ubuntu SMP Fri Jul 24 21:17:31 UTC 2015 x86_64 x86_64 x86_64 GNU/Linux
```

```
> sbcl --version
SBCL 1.2.4.debian
```

```
> lspci | grep VGA
01:00.0 VGA compatible controller: NVIDIA Corporation GF104GLM [Quadro 3000M] (rev a1)
```

jtbm37:
The problem goes away when setting a wallpaper with `feh`.

dbjergaard:
That would explain why I can't reproduce it since I set my wallpaper with feh in
my stumpwmwrc.  I'm guessing that the root pixmap just needs to be updated
whenever the groups are switched... Thanks for the info.

    dave

ralt:
I'm using debian with stumpwm's master, am not running `feh`, and can't reproduce. sbcl 1.2.14 (unstable).

I did the following to try to reproduce:

gnew foo
gnew bar
open window
switch to foo

dbjergaard:
[...]

PR #227, opened 2015-09-25
Do not process focus-in events of the pointer kind (fixes #196).

### LINKED ISSUE #196: Current-frame will not have a focused window in certain situations
Under certain circumstances switching to a new frame will not focus the window in that frame. Steps to reproduce:
1. From an empty desktop launch two terminals
2. split the screen and place a terminal in each frame
3. close or kill one of the windows
4. switch to the other frame
Expected behavior: other window gets the focus
Observed: the window doesn't have focus and clicking it doesn't give it focus.  The only way to focus the window is to cycle through the frames again.

This behavior is intermittant.  Sometimes the above procedure reproduces the bug, sometimes not.  I'm still looking into what exactly triggers the response.

aoterodelaroza:
Modified the focus-in event handler to ignore pointer focus-in
events. See issue #196 for a detailed explanation. It is possible that
this patch will also address the problems described in issues #187 and #94 .

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -557,9 +557,9 @@ converted to an atom is removed."
 (define-stump-event-handler :focus-out (window mode kind)
   (dformat 5 "~@{~s ~}~%" window mode kind))
 
-(define-stump-event-handler :focus-in (window mode)
+(define-stump-event-handler :focus-in (window mode kind)
   (let ((win (find-window window)))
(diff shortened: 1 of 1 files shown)


jorams:
I can confirm this appears to fix issue #196 for me.

jorams:
I ran with this patch for a day and no new problems turned up. Merging.

PR #228, opened 2015-09-27
Handle show-window-properties error when no windows are open

CodyReichert:
If you run `show-window-properties` when there are no open windows, you get some debug from this function - this just handles that and shows a message instead.


jorams:
:+1: Thanks!

PR #231, opened 2015-10-01
Command hooks

ralt:
I hope the docstrings are good enough? Would you like to have examples too in them?

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -137,10 +137,10 @@ out, an element can just be the argument type."
        (let ((%interactivep% *interactivep*)
              (*interactivep* nil))
          (declare (ignorable %interactivep%))
-         (run-hook-with-args *pre-command-hook* ',name (list ,@args))
+         (run-hook-with-args *pre-command-hook* ',name)
          (multiple-value-prog1
              (progn ,@body)
-           (run-hook-with-args *post-command-hook* ',name (list ,@args)))))
+           (run-hook-with-args *post-command-hook* ',name))))
      (export ',name)
      (setf (gethash ',name *command-hash*)
            (make-command :name ',name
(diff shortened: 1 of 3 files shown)


ralt:
Around `,@body` in the `defcommand` macro sounded like the best place to put these. I thought about `eval-command`, but it wouldn't catch all cases.

jorams:
Placing something after `,@body` breaks commands as functions as it changes the return value, so I don't think it's a good idea. Perhaps this?

```lisp
(multiple-value-prog1 (progn ,@body)
  (run-hook-with-args *post-command-hook* ',name (list ,@args)))
```

ralt:
So, it sounds like providing the arguments values is pretty hard. I either have to use an outer function (use `&rest` and `destructuring-bind` on `args`), or use something akin to `alexandria:parse-ordinary-lambda-list`.

Any option is going to add some overhead (to the code) that I'm not sure is worth it.

Is it fine if only the command name is passed to the hook?

dbjergaard:
no to alexandria (though this will change is stumpwm 2.0.0). Here's how I would guide myself if I were you:
1. Write what you need
2. If you can generalize without killing yourself, do it. The community will benefit. If not, stop at 1.  
3. Open PR

Honestly the more stumpwm is extended with modules the more powerful it becomes.  I'm happy to add hooks people need as long as they aren't re-wiring stuff deep in the core of what stumpwm does (thus introducing subtle bugs).  In short: don't kill yourself making the most general thing possible if there's a middle ground, and if there isn't, just submit what you had, I'll merge either.

ralt:
Alright, the PR is mergeable as of now then. Travis tests pass.

The hooks only pass the command name as a symbol to the listeners. (Which is enough to do some filtering based on it.)

PR #234, opened 2015-11-07
Fix (gmove-and-follow "<float>") exception

hying-caritas:
When invoke gmove-and-follow to a float group, an exception will
raised.  Because really-raise-window is defined for tile-window only.
Fix this via make it a generic function and provide method for
float-window too.


dbjergaard:
Thanks!

Issue #220, opened 2015-08-25
Search in windowlist only searches displayed titles, not full titles


ralt:
So if the application's name is hidden because the title is too long, searching for it won't work.

My current solution is to change the `*window-format*`, but then it's not shown as I'd like it to be... but as I need it to be searchable.

I guess the code to fix is around [here](https://github.com/stumpwm/stumpwm/blob/0303086e94cebd22614d06a640767b21a294cd4e/menu.lisp#L206-L237), but I couldn't make sense of it.

jorams:
The cropping of the name happens because [`SELECT-WINDOW-FROM-MENU`](https://github.com/stumpwm/stumpwm/blob/0303086e94cebd22614d06a640767b21a294cd4e/window.lisp#L951) passes the windows through a format, which defaults to `*window-format*` (in [`WINDOWLIST`](https://github.com/stumpwm/stumpwm/blob/4b6146dd8297ece3bc07127bd2bd593e41d2faeb/window.lisp#L1065)).

You could define an alternative command that uses a different format that doesn't crop titles. The searching code ([`CHECK-MENU-COMPLETE`](https://github.com/stumpwm/stumpwm/blob/0303086e94cebd22614d06a640767b21a294cd4e/menu.lisp#L150), called by `SELECT-FROM-MENU`, which you linked to) never sees the full title of the command, because the menu never sees it.

ralt:
> The searching code (CHECK-MENU-COMPLETE, called by SELECT-FROM-MENU, which you linked to) never sees the full title of the command, because the menu never sees it.

Indeed, that's the observed behavior, and I think this is a bug. Isn't it?

dbjergaard:
HI @jorams @ralt, can you check if the PR #241 looks good? I'll merge it if I don't hear that anything needs changing.

ralt:
I haven't tested it yet, but I already love it. I'll try to take some time this weekend to try it out.

PR #240, opened 2016-01-31
Helm-like menu

sebastiano-barrera:
I've made a few modifications to `menu.lisp`, so that the menu shows only the items matching all the regular expressions typed by the user (previously, the first matching item in the menu was highlighted, and that was it). Now it behaves a bit more like helm-mode for Emacs. It's been pretty useful, in particular with `pull-from-windowlist`.
The interface is exactly the same, so all other commands can take advantage of the new behavior with no further modifications. I've been using this stuff for a while now, and it hasn't shown any bug in the last few days.
Care to take a look?
(btw, this is the first time I've ever written anything in Common Lisp, any criticism/opinion is appreciated . Also, first time ever sending a PR, so, sorry if I'm not following quite the right procedure)

### DIFF
--- a/menu.lisp
+++ b/menu.lisp
@@ -177,10 +177,13 @@ backspace or F9), return it otherwise return nil"
       (vector-push-extend input-char (menu-state-current-input menu)))
     (handler-case
         (when (or input-char (not key-seq))
-          (let* ((match-regex (ppcre:create-scanner (menu-state-current-input menu)
-                                                    :case-insensitive-mode t))
-                 (match-p (lambda (element)
-                            (ppcre:scan match-regex (menu-element-name element)))))
+          (let* ((arg-line (make-argument-line :string (menu-state-current-input menu)
+                                               :start 0))
+                 (match-regexes (loop while (setf arg (argument-pop arg-line))
+                                   collect (ppcre:create-scanner arg :case-insensitive-mode t)))
+                 (match-p (lambda (item)
+                            (loop for re in match-regexes
+                               always (ppcre:scan re (menu-element-name item))))))
             (setf (menu-state-table menu) (remove-if-not match-p (menu-state-unfiltered-table menu))
                   (menu-state-selected menu) 0
                   (menu-state-view-start menu) 0
@@ -188,7 +191,7 @@ backspace or F9), return it otherwise return nil"
             (bound-check-menu menu)))
       (cl-ppcre:ppcre-syntax-error (condition)))))
 
-(defun select-from-menu (screen table &optional prompt
+(defun select-from-menu (screen table &optional (prompt "Search:")
                                         (initial-selection 0)
                                         extra-keymap)
   "Prompt the user to select from a menu on SCREEN. TABLE can be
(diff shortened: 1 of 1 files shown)


sebastiano-barrera:
Thanks! All done! The assertions were deleted (`menu-valid-view` was never called, I forgot to delete it earlier), but I've applied your suggestions to another couple of (assert) elsewhere in the same file.

menu.lisp
```diff
@@ -54,37 +54,49 @@
...
   (let ((len (length (menu-state-table menu))))
+    ;; Wrap around
     (setf (menu-state-selected menu)
           (cond ((< (menu-state-selected menu) 0) (1- len))
                 ((>= (menu-state-selected menu) len) 0)
                 (t (menu-state-selected menu))))
-    (when (and *menu-maximum-height*
-               (> len *menu-maximum-height*))  ; scrolling required
-      (let ((sel (menu-state-selected menu)))
+    (if 0
```
jorams: This doesn't make any sense. `(IF 0 <THEN> <ELSE>)` will always evaluate `<THEN>`, so  this entire `IF` is really just the `<THEN>`:

```lisp
(setf (menu-state-view-start menu) 0
      (menu-state-view-end menu) (length (menu-state-table menu)))
```

menu.lisp
```diff
@@ -149,32 +161,29 @@ backspace or F9), return it otherwise return nil"
...
-	  (setf (menu-state-selected menu) cur-pos)
-          (bound-check-menu menu)
-	  (return))))))
-
-(defun select-from-menu (screen table &optional prompt
+    (handler-case
+        (when (or input-char (not key-seq))
+          (let* ((arg-line (make-argument-line :string (menu-state-current-input menu)
+                                               :start 0))
+                 (match-regexes (loop while (setf arg (argument-pop arg-line))
```
jorams: `ARG` is an undefined variable. You probably want something like

```lisp
(loop for arg = (argument-pop arg-line)
      while arg
      ...)
```

menu.lisp
```diff
@@ -188,51 +197,45 @@ Returns the selected element in TABLE or nil if aborted. "
...
+                  :selected 0))
+           (keymap (if extra-keymap
+                       (list extra-keymap *menu-map*)
+                       (list *menu-map*))))
+      (bound-check-menu menu)
+      (catch :menu-quit
+        (unwind-protect
+             (with-focus (screen-key-window screen)
+               (loop
+                  (let* ((prompt-row (format nil "~A ~A"
```
jorams: A better format-string would be `"~@[~A ~]~A"`, because `PROMPT` can be `NIL` and if so, it shouldn't be printed.

menu.lisp
```diff
@@ -149,32 +161,29 @@ backspace or F9), return it otherwise return nil"
...
+                                   collect (ppcre:create-scanner arg :case-insensitive-mode t)))
+                 (match-p (lambda (item)
+                            (loop for re in match-regexes
+                               always (ppcre:scan re (menu-element-name item))))))
+            (setf (menu-state-table menu) (remove-if-not match-p (menu-state-unfiltered-table menu))
+                  (menu-state-selected menu) 0
+                  (menu-state-view-start menu) 0
+                  (menu-state-view-end menu) 0)
+            (bound-check-menu menu)))
+      (cl-ppcre:ppcre-syntax-error (condition)))))
```
jorams: The variable `CONDITION` isn't used, so you can just omit it.

jorams:
I'm running with this now. I like it so far and haven't hit anything problematic yet.

sebastiano-barrera:
All of that also applied. Happy to hear you like it!

dbjergaard:
Looks good to me, I'm merging it.

ralt:
Just saw this PR and thought it was somehow related to https://github.com/stumpwm/stumpwm/issues/220

Does this PR fix #220?

sebastiano-barrera:
Not really... I've implemented a similar workaround to the one proposed in #220 (a custom `*window-format*`). However, I've had that exact same problem lately and plan to implement the following design changes: change `select-from-menu` as to accept a custom filter predicate as additional argument; the filter predicate receives the object represented in the menu item and the current user input, and tells whether the item should be visible or not; if the predicate is not specified, the current behavior is retained. If this is useful to somebody, then it's coming soon!

PR #241, opened 2016-02-05
Menu filter predicate

sebastiano-barrera:
This aims to solve #220.

The changes are, basically:

1. an additional argument to `select-from-menu` specifies a filter predicate, which is a function that says whether a given menu item should be visible or not, based on the user-visible string, the object corresponding to the item, and user input;

2. `select-window-from-menu` specifies a filter predicate that also checks the user-input regexp against the full window title (not just the portion shown in the menu).

It's also possible to override `select-window-from-menu`'s default by setting the `*window-menu-filter*` variable.

(I still consider myself a CL beginner; any criticism/opinion on the code or the general design is appreciated. Also: should I squash some commits?)

### DIFF
--- a/menu.lisp
+++ b/menu.lisp
@@ -177,9 +177,7 @@ backspace or F9), return it otherwise return nil"
 predicate, an item is visible when it matches all of the regular
 expressions in USER-INPUT (multiple regexps are separated by one or
 more spaces; ARGUMENT-POP is used to split the string)."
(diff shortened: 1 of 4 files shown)


dbjergaard:
This looks good to me.

jorams:
Works handsomely, really nice.

sebastiano-barrera:
I'm happy this is useful!

PR #251, opened 2016-05-04
Sans invert rect

luka-n:
Workaround for this - https://github.com/stumpwm/stumpwm-contrib/issues/21

### DIFF
--- a/color.lisp
+++ b/color.lisp
@@ -380,13 +380,15 @@ rendered width."
                            (xlib:drawable-height px) t))
     (loop for parts in strings
           for row from 0 to (length strings)
-          for line-height = (render-string parts cc
-                                           (+ padx 0)
-                                           (+ pady y))
-          when (find row highlights :test 'eql)
-            do (invert-rect screen px 0 (+ pady y)
-                            (xlib:drawable-width px)
-                            line-height)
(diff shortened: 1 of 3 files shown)


color.lisp
```diff
@@ -380,13 +380,15 @@ rendered width."
...
-                                           (+ padx 0)
-                                           (+ pady y))
-          when (find row highlights :test 'eql)
-            do (invert-rect screen px 0 (+ pady y)
-                            (xlib:drawable-width px)
-                            line-height)
+          for line-height = (max-font-height parts cc)
+          if (find row highlights :test 'eql)
+            do (xlib:draw-rectangle px gc 0 y (xlib:drawable-width px) line-height t)
+            and do (xlib:with-gcontext (gc :foreground (xlib:gcontext-background gc)
```
PuercoPop: The `and do` is unnecessary do will evaluate multiple forms. ie

```lisp
...
if (find row highlights :test 'eql)
        do (xlib:draw-rectangle px gc 0 y (xlib:drawable-width px) line-height t)
           (xlib:with-gcontext (gc :foreground (xlib:gcontext-background gc)
                                   :background (xlib:gcontext-foreground gc))
             (render-string parts cc (+ padx 0) (+ pady y)))
      else
...
```

PR #255, opened 2016-05-16
Fix "y-or-n" prompt

alezost:
Hello, try this command:

```lisp
(defcommand bool-test (bool) ((:y-or-n ""))
  (if bool
      (echo "yes")
      (echo "no")))
```

Now call it interactively and you always get "yes" message no matter if
you press `y` or `n`.

Alternatively, try this:

```lisp
(defcommand bool-test (bool) ((:y-or-n ""))
  (echo bool))
```

It will display either `(T)` or `(NIL)`.  As you can see it wraps the
boolean value into a list. ``(define-stumpwm-type :y-or-n ...)`` does
this.  I also don't see a reason to use `values` and `when` there, or do
I miss anything?

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -300,8 +300,7 @@ then describes the symbol."
 (define-stumpwm-type :y-or-n (input prompt)
   (let ((s (or (argument-pop input)
                (read-one-line (current-screen) (concat prompt "(y/n): ")))))
-    (when s
-      (values (list (equal s "y"))))))
+    (equal s "y")))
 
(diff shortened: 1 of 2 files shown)


jorams:
Note that this `:y-or-n` is used in `tile-window.lisp`, in the command `remember`, where the value is actually used with `(first ...)`. I don't see the point either. If you change those two instances I'll merge this.

jorams:
:+1: Looks good to me!

alezost:
Thanks!

PR #260, opened 2016-05-29
Remove CONC1

PuercoPop:
The function is not called from anywhere else nor is it
exported. Furthermore PUSH accomplishes the same as CONC1


jorams:
`PUSH` doesn't actually do the same thing as `CONC1`, but yes, `CONC1` should go. I'll merge this later today.

dbjergaard:
sorry @jorams beat you to it ;)  Thanks for the PR @PuercoPop

Issue #265, opened 2016-07-10
restart-hard doesn't execute *quit-hook*


eadmund:
`restart-hard` leaves open file descriptors open; this means that if a StumpWM instance is e.g. running swank then the new StumpWM won't be able to open the swank port again.

I think that it'd be good to run the `*quit-hook*`, even on a hard restart, in order to cleanly shut down anything which is running — or maybe there should be a `*restart-hook*` instead?

dbjergaard:
Was this fixed by your PR? If not, feel free to open a new one, or let me know what I need to do to make restart-hard close the open file descriptors.  I'm not sure how to handle those.

eadmund:
Yup, this can be closed.

I figure that if there's a restart hook then anyone who opens a file descriptor can close it himself. There's probably a way for StumpWM itself to take advantage of this too, but that can be a PR for a different day.

PR #266, opened 2016-07-10
Add *restart-hook*

eadmund:
It is called during both soft & hard restarts.

Addresses #265 

(This also cleans up some whitespace at the end of lines)

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -259,6 +259,9 @@ window group and frame")
 (defvar *quit-hook* '()
   "A hook called when stumpwm quits.")
 
+(defvar *restart-hook* '()
+  "A hook called when stumpwm restarts.")
+
 (defvar *internal-loop-hook* '()
   "A hook called inside stumpwm's inner loop.")
 
--- a/stumpwm.lisp
+++ b/stumpwm.lisp
@@ -61,7 +61,7 @@ further up. "
 
 (defun error-handler (display error-key &rest key-vals &key asynchronous &allow-other-keys)
   "Handle X errors"
-  (cond 
+  (cond
     ;; ignore asynchronous window errors
     ((and asynchronous
           (find error-key '(xlib:window-error xlib:drawable-error xlib:match-error)))
@@ -185,7 +185,7 @@ The action is to call FUNCTION with arguments ARGS."
   "Parse an X11 DISPLAY string and return the host and display from it."
   (ppcre:register-groups-bind (protocol host ('parse-integer display screen))
 			      ("^(?:(.*?)/)?(.*?)?:(\\d+)(?:\\.(\\d+))?" display :sharedp t)
-    (values 
+    (values
      ;; clx doesn't like (vector character *)
      (coerce (or host "")
 	     '(simple-array character (*)))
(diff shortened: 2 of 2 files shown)


eadmund:
Any comments? My thought in adding a new hook is that restart & quit are subtly different enough that it's wise to handle them differently. Another option would have been to pass an argument to `*quit-hook*`, but if it were an optional argument it'd be a little ugly, and if it were a required argument then it'd be a breaking change.  In this case I think that being explicit is better than being implicit.

dbjergaard:
I gotta be honest, I was upset about the whitespace changes and so I tabled this until later in the week.  Not that I'm opposed to whitespace changes, just that I would prefer that they come in separate PR so there isn't a bunch of visual noise.

I agree, a restart hook probably makes the most sense, though the argument could be made that restart should be the same as cold-start.  Is there a way to make sure quit-hook also gets ran, we can leave the restart-hook for users who want to trigger something on restart that wouldn't get triggered otherwise.

eadmund:
Sorry, the whitespace changes were due to emacs — it stripped the trailing whitespace from the end of three lines. I didn't think that there was any way that trailing whitespace could be intentional, so I didn't bother opening up the file in vi and re-inserting those spaces. I've just now pushed a commit which adds them back.

Regarding `*quit-hook*`, it's certainly possible to run it as well — but then there's no way for the quit hooks to know whether it's a restart or a quit. I think that there's a useful distinction between 'I'm `exec(3)`-ing a new process, but my children may still run' and 'I'm exiting now,' e.g. say one wants to prompt emacs to save buffers and quit, or somesuch — one wouldn't want to do that on a window manager restart, but would on WM quit.

I think the big difference between a restart and a cold start is that child processes, fds &c. are still around. One can restart StumpWM without restarting one's terminal, browser, editor &c. That does raise an issue, though — is there any need to expose to a new StumpWM that it was restarted rather than cold-started (and if so, is there any reliable way to detect that)?

dbjergaard:
I see your logic now.  I'm OK with this as is.

I apologize for being anal about whitespace.  The codebase is pretty crufty, and sometimes best practice formatting introduces whitespace changes that create a lot of visual noise.  I should be more tolerant, but sometimes I have to read the code a few times in order to make sure I'm parsing functional vs ws changes. If you want to fix that, you are more than welcome to submit a whitespace only PR and I'll merge it.

I don't ever restart stumpwm since my stumpwmrc/xinitrc spawn a number of helper daemons.  I suppose the correct way is to put those daemons exclusively in the xinitrc and let the stumpwmrc handle lispy only things. Then stumpwm restarting makes sense to me.  

As far as reliably detecting cold vs restarted StumpWM, anything I can think of would be hacky and probably not worth the extra complexity.

PuercoPop:
Pro tip @dbjergaard You can add `?w=1` to a diff URL to remove the whitespace noise when reading the changes ([ej](https://github.com/stumpwm/stumpwm/pull/266/commits/4ea0fb2da533b3ecc85d6380eea15c72a6103966?w=1)). I agree that it is good practice to not modify lines unrelated to your change in the same commit though.

dbjergaard:
@PuercoPop thanks for that! I will make sure to use it in the future.

eadmund:
> If you want to fix that, you are more than welcome to submit a whitespace only PR and I'll merge it.

I may get a round tuit …

> I don't ever restart stumpwm since my stumpwmrc/xinitrc spawn a number of helper daemons. I suppose the correct way is to put those daemons exclusively in the xinitrc and let the stumpwmrc handle lispy only things. Then stumpwm restarting makes sense to me.

I'm actually working on a StumpWM-based desktop environment (tentatively named `branch`, to pun off of 'stump'), so I restart pretty frequently. If stumpwm spawns a process or opens a file, I think it makes sense for the `exec`-ed `stumpwm` to still have those procs running and those files open. There's nothing wrong in POSIX with forking off children and then execing oneself, although it is maybe a bit unusual.

PR #270, opened 2016-08-05
Ioloop

dolda2000:
This is a patch-set for a multiplexing I/O loop, for StumpWM to be able to handle events from multiple I/O channels, rather than only watching the X11 display for events. I plan to use this in order to display continuous output from long-running child processes in a StumpWM message window.

It is admittedly wholly undocumented. I'm perfectly willing to document the code better if you want me to, but I thought I'd check in with it first. The I/O loop code is technically independent of StumpWM structure and could possibly be useful in other CL projects too.

The code only contains an "actual" implementation for SBCL, as that's what I use. For other Lisps, there's a "dummy" loop that attempts to emulate what the current StumpWM top-level loop does (but consequently doesn't actually support multiple I/O channels). The SBCL implementation has the side-effect that timers can now trigger with sub-second precision. It is unfortunately based on select(2), since that seems to be the only thing SBCL supports from what I can see, but given support for the relevant syscalls, it should be perfectly portable to poll, epoll or kqueue as well.

### DIFF
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -143,6 +143,11 @@
   "The default I/O loop implementation. Should be generically optimal
   for the given LISP implementation and operating system.")
 
+(defvar *current-io-channel* nil
+  "While processing an I/O channel, this variable is dynamically bound
+  to the channel in question. This is provided primarily for
+  error-handling code.")
+
 ;; Default methods for the above
 (defmethod io-channel-handle (channel event &key &allow-other-keys)
   (declare (ignore channel event)))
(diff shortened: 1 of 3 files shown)


PuercoPop:
Hey, this is really cool!

Ideally the io-loop used would be extendable and configurable by the user. I'm thinkong of a [cl-async](https://github.com/orthecreedence/cl-async implementation). The only issue preventing this I see is that `make-io-loop` is called directly. Just adding a layer of indirection storing the io-loop constructor in a global variable would be good enough

```lisp
(defun stumpwm-internal-loop ()
  (let ((io (funcall *io-loop-ctor*))) ...))

```
I see you are using sb-unix:unix-select, IOLIB has a similar API. It wouldn't be to hard to port it. Although StumpWM has a no extra dependencies policy so that would have to be off-repo

Feedback

- Instead of having the SBCL specific and the fallback implementation in the same file using a `#+sbcl(progn ...)` and `#-sbcl(progn ...)` it would be better to use [ASDF's if-feature feature](https://common-lisp.net/project/asdf/asdf.html#if_002dfeature-option)  to split the implementation specific parts

```lisp
(:file "module")
(:module "io-loop"
 :components ((:file "protocol")
              (:file "sbcl" :if-feature :sbcl)
              (:file "fallback" :if-feature (:not :sbcl))))
(:file "stumpwm")
```

- Better leave *toplevel-io* unbounded by default. If somehow the code tries to access it before the io-loop is set it would signal an error

 ie.
```lisp
(defvar *toplevel-io*)
(setf (documentation '*toplevel-io* 'variable)
      "Top-level I/O loop")
```

### Style Nitpicks

- Use a newline between function and method definitions
- No need to quote the empty list. Ej. (let ((rest ())) ...) would work just fine. Although I prefer nil in such cases
- Add documentation strings at the very list to the generic functions.

P.D. Not I'm not a StumpWM developer, to take my feedback accordingly

dolda2000:
Good point about I/O loop extensibility; there is certainly very little point to not simply having IO-LOOP itself be a generic function, so I pushed a change for this. It was also nice in the way that the dummy I/O loop didn't have to be cordoned off, so it is now always compiled instead. It did require some changes to IO-CHANNEL-FD, however, since different implementations need different representations, but that should be all OK. It is now called IO-CHANNEL-IOPORT instead.

>[...] it would be better to use ASDF's if-feature feature to split the implementation specific parts

I would probably do that if this were a stand-alone library, but as just part of StumpWM it felt nicer to not let it throw around more weight than necessary in terms of the number of files. Just my opinion, though, and if the StumpWM developers feel differently, I've no objections.

>Better leave toplevel-io unbounded by default.

Maybe it's just me, but I have somewhat of an aversion to unbound variables. Leaving it as NIL also throws fairly immediate errors, though, since there are no methods for a NIL I/O-loop.

>No need to quote the empty list.

I know, but I like quoting my empty lists when used as data, and that's the style used elsewhere in StumpWM too, so I think it's consistent. :)

>Add documentation strings at the very list to the generic functions.

Yes. As long as the StumpWM developers don't have any major objections to the feature as such, I'll certainly add such.

dbjergaard:
Hi, 

Before this gets merged it has to be tested with ccl and clisp.  Ideally ecl as well, but that takes some doing.  If this doesn't work for ccl or clisp, it can't be merged into stumpwm, but we can use it in paulownia (the codename for stumpm 2.0).  In fact, I would/will try to merge this method there ASAP since it completely restructures the event loop for the WM.

Please document everything! Especially some commentary in ioloop.lisp so people old and new understand what it is for and why its important.  

WRT to packaging: if we can't merge this into stumpwm now if it breaks ccl/clisp support, then I prefer that it gets its on asdf package that paulownia depends on.  In fact, we could do that now and have it live in the stumpwm organization. 

Thanks for the code! I'm glad you took the time to open the discussion now so that we can get it cleanly integrated.  It is a significant change to how stumpwm operates.

PuercoPop:
I've done a cursory test run it works with CCL. The CLISP bundled with ubuntu broke recently so I can't test it with clisp though, but the fallback should work

dolda2000:
I just added some documentation. Feedback is welcome.

dolda2000:
>If this doesn't work for ccl or clisp

While I certainly agree that it should be tested, I should like to add that there should be no reason to assume that it wouldn't work with CCL or CLisp. The dummy loop should do exactly what StumpWM used to do, and I've tested it on SBCL where it seemed to work (though I haven't tested it for any extended time). There may be bugs, of course, but I can't see that there could be any architectural reasons that it shouldn't work with any Lisp that has been supported thus far.

I tried to test with CLisp just now, but for some reason I couldn't get ASDF to work on it. Will see if I can find out why.

dolda2000:
If only as an example of using the I/O loop, here's my current implementation for the long-running command execution with message-window output that I mentioned in the OP: https://gist.github.com/dolda2000/08433fffc15c35658dec35ee262391d6

It very much only supports SBCL and the "nonblocking read-sequence" implementation is horribly inefficient, but it works pretty well.

dbjergaard:
I'm happy to merge this, we'll deal with bugs as they arise :)

dolda2000:
Cool! I'll likely be back later with some patches relating to recovering from errors gracefully in I/O channels.

dbjergaard:
Hi, #272 is directly related to this PR.  Can you please look into it?

jorams:
Maybe we should consider reverting this until we can get it fully working. Right now the StumpWM in Quicklisp is broken for many people.

dbjergaard:
I've partially addressed the glaring issues with the window indicators.  Do you think the current head should still be reverted?

jorams:
Reading #272, it doesn't seem fixed enough. Timeouts should work after a certain amount of time not require any manual actions. If I understand the issue correctly, it would also impact things like mode line updates when you're not actively doing anything.

That said, I am personally not having any issues, so I don't really know how bad it is.

jorams:
And now that I've said that, I notice that I *am* experiencing the issue since this last reboot. It's terrible, my MPD connection keeps timing out.

dbjergaard:
Ok, lets roll it back.  My only worry is my impetus to fix it will go way down.

dbjergaard:
OK, well we're gonna have to manually revert it since I've made other changes that break this PR.

jorams:
@dbjergaard See my comment in #272. The changes you made to fix it are what's causing the breakage for me.

Things are interacting in weird ways.

dbjergaard:
I admit I don't understand the IO loop code well enough to implement timers within the new paradigm.

I don't think the original author understood the timer code well enough to implement them either.  

The original timer code ran in the main loop and checked a list with time stamps and ran those whose time-stamps were older than the current time.  

The new timer code got hooked into the ioloop framework and wasn't getting called properly (resulting in expired timers getting removed but not executed afaik).

My fix was to re-add the old timer logic to the part of ioloop that matched where the timer code was run previously.  This partially addressed the problem, but the new behavior of timers being run only when new events were processed showed up.

Based on this evidence I conclude (and your fix of registering an empty timer):
1. Previously the assumption that every iteration of the internal loop was an appropriate time to check the current time to a list of future times was valid
2. After changing the IO so that it can handle multiple streams this assumption was invalidated
3. By adding the timer code back to the event processing part of the loop, timers only get processed when X11 events that stumpwm handles fires
4. Registering an empty timer (after reverting 3) has the effect of 1 with the new code.

If all these pieces hang together then it may be that your hack is the only way to get timers working properly.  The alternative is to find where in the ioloop logic the assumption for 1 still holds and insert the timer code there, or to rework the timer code completely to properly utilize ioloop.

I hope this makes sense...

jorams:
As noted in #272 and #276, the issue is fixed in 13ff02bf6a02daf6f81b9f311371a1e64e657621.

PR #271, opened 2016-08-21
Top-level condition handling

dolda2000:
I wanted to add more proper error-handling for I/O channels that cause errors, but the top-level condition handling was already quite complex, so I tried to clean it up a bit by moving its functionality to a generic function. Then I added I/O channel error handling.

One advantage of handling errors this way is also that new methods can be defined/redefined on HANDLE-TOP-LEVEL-CONDITION without having to restart the STUMPWM-INTERNAL-LOOP, making it easier to patch error-handling at runtime.

The resulting functionality should be similar to previously, but as I was at it, I did change so that non-warning, non-serious conditions are ignored, rather than handled as if they were errors. I would like to think that is the correct way of handling them, as they are likely to be "informational" conditions rather than things to be handled.

Also, this allows for the possibility to move the whole XLIB:LOOKUP-ERROR handling to workarounds.lisp, where the rest of it is, but I wasn't sure whether I should do so.

### DIFF
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -66,7 +66,8 @@
 ;;;; multiplexer to actually run it.
 
 (export '(io-channel-ioport io-channel-events io-channel-handle
-          *default-io-loop* io-loop io-loop-add io-loop-remove io-loop-update
+          *default-io-loop* *current-io-loop* *current-io-channel*
+          io-loop io-loop-add io-loop-remove io-loop-update
           callback-channel callback-channel-stream callback-channel-events))
 
 ;;; General interface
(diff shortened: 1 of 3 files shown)


dolda2000:
Oh, and the 592065a commit is technically unrelated, actually; but it was a small change so I just included it.

dbjergaard:
Go ahead and move the code to workarounds.lisp, better to have it consolidated than fragmented.

dolda2000:
Oh, I see in fact that the lookup-error workaround had been complete removed (I'm normally working on a different branch, where it's still present). Since that was the case, I went ahead and removed all the supporting code. I don't assume anyone is going to miss it, right?

Issue #276, opened 2016-09-10
*mode-line-timeout* is ignored


Jehops:
The  mode-line no longer updates after commit 56303adf44876ef71571e4745521bbccacae477a.  After reverting back to 9d854fda0b11815930468a27da280ebbf2073971, the mode-line updates again.

dbjergaard:
This is related to #270 and #272, if you can see how to get the timers to execute properly please let me know/open a PR.

jorams:
Fixed by 13ff02bf6a02daf6f81b9f311371a1e64e657621

PR #279, opened 2016-09-16
Refactor RUN-EXPIRED-TIMERS

PuercoPop:
This is a minor change intended to make `run-expired-timers` easier to understand. `sort-timers` could benefit from moving the re-scheduling logic out of it. Maybe returning a third value, `to-reschedule` and a separate function `reschedule-timers`.

One thing I'm unsure of is if `execute-timers` is necessary or if `(map nil 'execute-timer pending)` is clear enough.

Another thing is that expired or expired-timers may be a better name than pending in `sort-timers`

Thoughts?

### DIFF
--- a/stumpwm.lisp
+++ b/stumpwm.lisp
@@ -112,21 +112,35 @@ The action is to call FUNCTION with arguments ARGS."
   (setf (timer-time timer) (+ (get-internal-real-time)
                               (* when internal-time-units-per-second))))
 
-(defun run-expired-timers ()
+(defun sort-timers (timers)
   (let ((now (get-internal-real-time))
-	(timers *timer-list*)
-	(pending '())
-	(remaining '()))
-    (setf *timer-list*
-	  (dolist (timer timers (sort remaining #'< :key #'timer-time))
-	    (if (<= (timer-time timer) now)
-		(progn (push timer pending)
-		       (when (timer-repeat timer)
-			 (schedule-timer timer (timer-repeat timer))
-			 (push timer remaining)))
-		(push timer remaining))))
-    (dolist (timer pending)
-      (apply (timer-function timer) (timer-args timer)))))
+        (pending ())
+        (remaining ()))
+    (dolist (timer timers)
+      (if (<= (timer-time timer) now)
+          (progn (push timer pending)
+                 (when (timer-repeat timer)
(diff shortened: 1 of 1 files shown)


dbjergaard:
wrt run-with-timer: the merge ensures that the new timer is inserted in the correct chronological order so that you can schedule a timer that may run at a different time than when it is inserted into the list of timers.  I am leery of changing it to the simpler cons or append. 

Overall: I'm a little worried about the number of extra function calls that are now called on each iteration of the internal loop.  I'm not familiar enough with the performance of common lisp idioms, so I don't know if my fears are unfounded.  I agree that this PR makes the functionality of the timers more manifest.

PuercoPop:
The chronological order is also ensure by update-timer-list. When inserting only one atom to a list the (cons atom <known-list>) is a common idiom in CL ime. The correct solution would be to do like SBCL's timer module does and use a priority-queue instead of a list that preserves orders at insertion.

If extra function calls are a concern, one can always insert inline declarations, but imho StumpWM's code tends favor performance instead of readability a little too much, w/o measurements. I'm using this code on my day to day w/o any performance issue. But then again, I only use SBCL and I understand the concern of 'if it aint broke don't fix it'. Plus a mid term goal to improve performance would be to have a module that uses epoll through iolib, which would be easier w/ more understandable code (This is something I want to work on).

Thanks for the prompt feedback!

dbjergaard:
There are parts of StumpWMs codebase that make me cringe in their low-level nature and lack of regard for whether it could be re-expressed more clearly.  

OTOH we have a diverse community of people using stumpwm, and sometimes a change on one piece of hardware makes no difference, and on the next it makes it unusable.  

I'll merge this in, just because I like it, and because I have a feeling that the timer related bugs #276 #272 are not resolved, so we will be touching related code anyway.

Cons or append, I don't mind.  I think append is the "modern" way, but everyone uses cons anyway.  I prefer append.

PuercoPop:
Append only works in list so cons is more readable for a single item imho. But it is a style choice so other people may prefer append. I'll add the append version to the PR when I get home. Or do you want to do it? I've checked the new option to allow maintainers to have write access to the PR

Yes there are still more changes to be made to the timer code, I refactored this code when reading the code to look for hypothesis if what could be happening on #272 but I'm haven't gotten to it yet.

PR #280, opened 2016-09-20
Remove *RESIZE-BACKUP*

PuercoPop:
It was never used anywhere. The variable was introduced alongside the
initial implementation of iresize, in the commit https://github.com/stumpwm/stumpwm/commit/a71e89. It was
probably used for for debugging purposes.


jorams:
Good catch!

PR #281, opened 2016-09-22
wrappers: Use PRINT-BACKTRACE instead of backtrace

PuercoPop:
SB-DEBUG:BACKTRACE has been deprecated in favor of PRINT-BACKTRACE since
SBCL version 1.2.15


jorams:
And `print-backtrace` has been available sine SBCL 1.1.5. I think this is a good idea, merging.

PR #282, opened 2016-09-22
Remove unused screen argument from RENDER-STRINGS

PuercoPop:
Its use in the function was removed in #251. Remove it as it is no
longer necessary

### DIFF
--- a/color.lisp
+++ b/color.lisp
@@ -352,7 +352,7 @@ rendered width."
               (apply #'apply-color cc (first part) (rest part))))
     (values height draw-x)))
 
-(defun render-strings (screen cc padx pady strings highlights)
+(defun render-strings (cc padx pady strings highlights)
   (let* ((gc (ccontext-gc cc))
          (xwin (ccontext-win cc))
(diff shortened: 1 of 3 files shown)


jorams:
I think this should also remove it from the two callers of this function:

- `mode-line.lisp:415` in `redraw-mode-line`
- `message-window.lisp:223` in `echo-string-list`

PuercoPop:
Right, sorry I'll fix this PR

dbjergaard:
@PuercoPop I was looking over your PR history.  Would you like commit rights? Jorams is in a similar situation.  We agreed the best way to handle it is to continue opening PRs for code review for larger commits/changes, but you'll be able to merge other's PRs and have commit access for smaller changes. Let me know what you think.

PuercoPop:
@jorams done

@dbjergaard Sure I'd like them. But I would probably not use them for myself, even for small changes. As this PR shows I sometimes forget about some details so I prefer if my proposed changes go through an extra set of eyes.

jorams:
Thanks!

PR #283, opened 2016-10-18
Register systems as immutable in a built image

jorams:
If a contrib module depends on the `stumpwm` system, ASDF will look for it in the build directory. When that directory no longer exists (which is fairly likely), loading of the module will fail.

By registering StumpWM and dependencies as ["immutable"](https://common-lisp.net/project/asdf/asdf.html#index-register_002dimmutable_002dsystem-1), we tell ASDF to treat the system as already loaded. It will not look for the system definition nor the source code again, unless you explicitly request it to be loaded.

Fixes stumpwm/stumpwm-contrib#17 (already marked as fixed, but not generally)
Fixes stumpwm/stumpwm-contrib#53

Some more testing would be appreciated.

### DIFF
--- a/make-image.lisp.in
+++ b/make-image.lisp.in
@@ -4,6 +4,13 @@
 
 #-ecl (stumpwm:set-module-dir "@MODULE_DIR@")
 
+(when (uiop:version<= "3.1.5" (asdf:asdf-version))
+  ;; We register StumpWM and its dependencies as immutable, to stop ASDF from
+  ;; looking for their source code when loading modules.
+  (asdf:register-immutable-system :stumpwm)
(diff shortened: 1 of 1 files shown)


dbjergaard:
will this have any consequences to rebuilding stumpwm? It sounds like it could prevent recompiling parts of the code if they change because we've explicitly told asdf that it shouldn't search for files.

jorams:
It shouldn't.

- If you're just using `make`, the new image will reload everything just fine because this setting is applied happens *after* loading StumpWM.
- If you're reloading into an existing built image, you will use something like `(asdf:load-system :stumpwm)`, which is not affected by the immutable system setting. It *will* then cause CL-PPCRE and the other dependencies to not be reloaded unless you explicitly load them individually as well.

PR #285, opened 2016-10-25
Clean up a variety of things

jorams:
I don't think any of these changes will be controversial (except maybe the screen initialization), but there are enough changes that I think a pull request is more appropriate than pushing directly.

None of these changes should be externally noticeable, but some more testing is in order to make sure I didn't miss anything.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -107,18 +107,16 @@
   (dformat 4 "CONFIGURE NOTIFY ~@{~S ~}~%" stack-mode window x y width height border-width value-mask)
   (let ((screen (find-screen window)))
     (when screen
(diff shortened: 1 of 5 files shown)


PuercoPop:
:+1: On everything except the screen initialization, which because it modifies a lot of code I haven't thoroughly checked out. But from skimming it seems ok as well.

jorams:
I improved the screen initialization a bit more and rebased onto master. The build succeeded, so I think it's ready to merge.

dbjergaard:
It won't be the first time we merged something to master and triaged unforseen bugs :)

Issue #284, opened 2016-10-23
"Value NIL is not of the expected type REAL" error (stumpwm compiled with ccl on FreeBSD11)


pbgc:
I tried to compile StumpWM on FreeBSD11 with ccl.
Running it I always get (immediately after start and pressing any key or mouse button):

```
Value NIL is not of the expected type REAL.
 (821C19858) : 0 (PRINT-CALL-HISTORY :CONTEXT NIL :PROCESS NIL :ORIGIN NIL :DETAILED-P NIL :COUNT 100 :START-FRAME-NUMBER 0 :STREAM #<STRING-OUTPUT-STREAM  #x302000D2BACD> :PRINT-LEVEL 2 :PRINT-LENGTH 5 :SHOW-INTERNAL-FRAMES NIL :FORMAT :TRADITIONAL) 853
 (821C199C0) : 1 (PRINT-BACKTRACE 100) 149
 (821C199E0) : 2 (BACKTRACE-STRING) 189
 (821C19A38) : 3 (FUNCALL #'#<#<STANDARD-METHOD STUMPWM::HANDLE-TOP-LEVEL-CONDITION (SERIOUS-CONDITION)>> #<TYPE-ERROR #x302000D2BBAD>) 533
 (821C19A68) : 4 (FUNCALL #'#<(:INTERNAL STUMPWM::STUMPWM-INTERNAL-LOOP)> #<TYPE-ERROR #x302000D2BBAD>) 53
 (821C19A88) : 5 (SIGNAL #<TYPE-ERROR #x302000D2BBAD>) 989
 (821C19AE0) : 6 (%ERROR #<TYPE-ERROR #x302000D2BBAD> NIL 4365759348) 117
 (821C19B08) : 7 (FUNCALL #'#<CCL::DEFAULT-REQUIRE-TYPE-RESTARTS> 4365759348 NIL REAL) 1093
 (821C19BA0) : 8 (<-2 NIL 268245972) 2301
 (821C19BD8) : 9 (FUNCALL #'#<#<STANDARD-METHOD IO-LOOP (STUMPWM::XLIB-IO-LOOP)>> #<STUMPWM::XLIB-IO-LOOP #x302000D0328D> :DESCRIPTION "StumpWM") 2173
 (821C19C88) : 10 (STUMPWM-INTERNAL-LOOP) 837
 (821C19CE8) : 11 (STUMPWM-INTERNAL ":0") 3629
[...]

PR #286, opened 2016-10-28
Implement CALL-IN-MAIN-THREAD

lokedhs:
This function allows a multi-threaded module to perform actions on the main
event handler thread and should be used whenever a different thread
needs to invoke any stumpwm functions.

Currently, only SBCL is supported.

### DIFF
--- a/stumpwm.lisp
+++ b/stumpwm.lisp
@@ -189,41 +189,6 @@ The action is to call FUNCTION with arguments ARGS."
 (defmethod io-channel-handle ((channel stumpwm-timer-channel) (event (eql :loop)) &key)
   (run-expired-timers))
 
-(defun open-pipe (&key (element-type '(unsigned-byte 8)))
-  #+sbcl
-  (multiple-value-bind (in-fd out-fd)
-      (sb-posix:pipe)
-    (let ((in-stream (sb-sys:make-fd-stream in-fd :input t :element-type element-type))
-          (out-stream (sb-sys:make-fd-stream out-fd :output t :element-type element-type)))
-      (values in-stream out-stream)))
-  #+ccl
-  (multiple-value-bind (in-fd out-fd)
-      (ccl::pipe)
-    (let ((in-stream (ccl::make-fd-stream in-fd :direction :input :element-type element-type))
-          (out-stream (ccl::make-fd-stream out-fd :direction :output :element-type element-type)))
-      (values in-stream out-stream)))
-  #-(or sbcl ccl)
-  (error "Unsupported CL implementation"))
-
-(defun make-lock ()
-  #+sbcl
-  (sb-thread:make-mutex)
-  #+ccl
-  (ccl:make-lock "Anonymous lock")
-  #-(or sbcl ccl)
-  nil)
-
-(defmacro with-lock-held ((lock) &body body)
-  #+sbcl
-  `(sb-thread:with-mutex (,lock)
-     ,@body)
-  #+ccl
-  `(ccl:with-lock-grabbed (,lock)
-     ,@body)
-  #-(or sbcl ccl)
-  `(progn
-     ,@body))
-
 (defclass request-channel ()
   ((in    :initarg :in
           :reader request-channel-in)
@@ -271,7 +236,7 @@ The action is to call FUNCTION with arguments ARGS."
 
 #-(or sbcl ccl)
 (defun call-in-main-thread (fn)
-  (funcall fn))
+  (run-with-timer 0 nil fn))
 
 (defclass display-channel ()
   ((display :initarg :display)))
--- a/stumpwm.lisp
+++ b/stumpwm.lisp
@@ -189,6 +189,41 @@ The action is to call FUNCTION with arguments ARGS."
 (defmethod io-channel-handle ((channel stumpwm-timer-channel) (event (eql :loop)) &key)
   (run-expired-timers))
 
+(defun open-pipe (&key (element-type '(unsigned-byte 8)))
+  #+sbcl
+  (multiple-value-bind (in-fd out-fd)
+      (sb-posix:pipe)
+    (let ((in-stream (sb-sys:make-fd-stream in-fd :input t :element-type element-type))
+          (out-stream (sb-sys:make-fd-stream out-fd :output t :element-type element-type)))
+      (values in-stream out-stream)))
+  #+ccl
+  (multiple-value-bind (in-fd out-fd)
+      (ccl::pipe)
+    (let ((in-stream (ccl::make-fd-stream in-fd :direction :input :element-type element-type))
+          (out-stream (ccl::make-fd-stream out-fd :direction :output :element-type element-type)))
+      (values in-stream out-stream)))
+  #-(or sbcl ccl)
+  (error "Unsupported CL implementation"))
+
+(defun make-lock ()
+  #+sbcl
+  (sb-thread:make-mutex)
+  #+ccl
+  (ccl:make-lock "Anonymous lock")
+  #-(or sbcl ccl)
+  nil)
+
+(defmacro with-lock-held ((lock) &body body)
+  #+sbcl
+  `(sb-thread:with-mutex (,lock)
+     ,@body)
+  #+ccl
+  `(ccl:with-lock-grabbed (,lock)
+     ,@body)
+  #-(or sbcl ccl)
+  `(progn
+     ,@body))
+
 (defclass request-channel ()
   ((in    :initarg :in
           :reader request-channel-in)
@@ -196,22 +231,10 @@ The action is to call FUNCTION with arguments ARGS."
           :reader request-channel-out)
    (queue :initform nil
           :accessor request-channel-queue)
-   (lock  :initform (bordeaux-threads:make-lock)
+   (lock  :initform (make-lock)
           :reader request-channel-lock)))
 
 (defvar *request-channel* nil)
-
-#+sbcl
-(defun open-pipe (&key (element-type '(unsigned-byte 8)))
-  (multiple-value-bind (in-fd out-fd)
-      (sb-posix:pipe)
-    (let ((in-stream (sb-sys:make-fd-stream in-fd :input t :element-type element-type))
-          (out-stream (sb-sys:make-fd-stream out-fd :output t :element-type element-type)))
-      (values in-stream out-stream))))
-
-#-sbcl
-(defun open-pipe (&key (element-type '(unsigned-byte 8)))
-  (error "Unsupported CL implementation"))
 
 (defmethod io-channel-ioport (io-loop (channel request-channel))
   (io-channel-ioport io-loop (request-channel-in channel)))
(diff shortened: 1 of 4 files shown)


PuercoPop:
To provide further context I'll paste the conversation on `#stumpwm`

<details>
  <summary> Click to Expand </summary>
```
15:23 <loke> [03:06:56] Is there a way to force the repaint of the status
             bar?
15:23 <mood> [03:50:38] loke: Perhaps (update-all-mode-lines) ?
15:23 <loke> [03:51:27] mood: Can I call that from a different thread?
15:23 <loke> [03:51:42] In general, how much of stump is thread-safe?
15:23 <mood> [04:35:51] loke: I wouldn't assume anything is thread-safe, to
             be honest
15:23 <loke> [04:37:06] mood: I thought so.
15:23 <loke> [04:37:24] Is there a method I can use to queue a function to
             be called in the event handler thread?
15:23 <mood> [04:37:36] I'm not sure if CLX is even thread-safe
15:23 <loke> [04:38:28] mood: That's the point. So I'd need a
             CALL-IN_MAIN-THREAD function or something.
15:23 <mood> [04:39:06] loke: You could probably work something out with the
             io loop, though I'm not sure how hard it would be
15:23 <mood> [04:39:39] Yet another area to work on. I should make myself a
             real todo list for this
15:23 <loke> [04:39:52] mood: If something like that doesn't exist, I'd be
             happy to see if I can implement it.
15:23 <loke> [04:40:03] It's definitely needed in my dbus module.
15:23 <loke> [04:40:14] Since there are lots of asynchrnous events
             happening.

15:44 <PuercoPop> loke: the io-loop has a 'job queue', the entry point is
                  RUN-WITH-TIMER but the API is not very good. (Maybe we
                  should copy sb-timer?). Any feedback would be
                  appreciated. https://github.com/stumpwm/stumpwm/blob/master/stumpwm.lisp#L90
22:43 <loke> PuercoPop: I'm looking at it, and it's not very reliable.
22:43 <loke> I create a timer for 1 second and it sometimes takes 5 seconds
             for it to trigger
22:43 <loke> (or more)
22:44 <loke> Also, RUN-WITH-TIMER isn't thread-safe in itself.
22:45 <loke> Ah, I see why that happens. It's because the io-channel timeout
             is only updated when the main thread does something. Makes
             sense.
22:47 <loke> I would suggest creating a pipe, and then an io channel
             listeners on that pipe. Jobs would be scheduled by adding
             closures to a list, and then send a message on the pipe,
             indicating that there are waiting jobs. The main thread can
             then drain that list by funcalling the closures.
23:03 [PuercoPop nods]
23:06 <PuercoPop> So a 'job channel'
23:13 <loke> PuercoPop: Yes. Exactly.
23:14 <loke> The job queue management would be the only thing that would
             need proper synchronisation. Everything else can be called
             though the queue.
23:27 <loke> PuercoPop: I started working on an implementation of this idea.
23:27 <loke> How, I need to open a pipe. The question is, I can do it using
             SBCL-specific code, but there is no generic way of doing it.
23:28 <loke> Adding a dependency on iolib would not be an option, I guess.
23:31 <PuercoPop> loke: ideally the iolib code should be pluggable so we
                  could have a contrib to use iolib
23:31 <PuercoPop> I've been meaning to suggest using cl-launch so we can use
                  quicklisp so we can clean-up some code.
23:32 <PuercoPop> But having iolib as a dependency in the main repo would be
                  a no go
23:33 <loke> Well, the job queue would be a central part of the
             application. After all, it needs to be the main way to run
             things, so it can't be an optional thing.
23:33 <loke> This means, that if the solution will be built on pipes, a way
             to create that pipe needsw to be found.
23:34 <loke> One altnerative would be to use platform-specific code, but I
             only know how to do it in SBCL
23:34 <loke> (which is SB-POSIX:PIPE)
```
</details>

jorams:
I'll have to look more closely at this later, but some quick initial ideas:

- We might not want bordeaux-threads as a dependency right now.
- CCL has `ccl::pipe`, which does what we want but sadly isn't exported.
- xcvb uses `linux:pipe` for this on CLISP, but no `linux` package exists in my CLISP.
- This appears to currently break StumpWM on all implementations other than SBCL (because of the `open-pipe` call in `stumpwm-internal-loop`), which we don't want.
- Maybe on implementations that don't support this functionality, `call-in-main-thread` could use `run-with-timer`.

lokedhs:
The most recent commit addresses the concerns raised above. While it's not actually been tested, it should work on CCL now. There is also a fallback which is no worse than what we had before. It calls the functions directory rather than using the timer. The timer should be a bit more safe, but if used the function will not invoked until the next event, which could take some time. Should the timer still be used?

dbjergaard:
Hi,

We can't add dependencies to stumpwm (but paulownia is a different story).  Also, we can't add something that only works in sbcl, and has untested solutions in ccl and no solution in ecl and gcl.  

I know this is a frustrating development constraint (which is why paulownia will only support sbcl) but a number of users run stumpwm from esoteric lisp flavors, and so doing non-generic things is usually a non-starter.

So, the bar for getting merged is "must not break the build on ccl and sbcl" which this passes, but in reality the bar is a little higher because I don't want to break ecl or gcl compatability.

jorams:
I did some testing:

- CCL currently crashes during startup because `io-channel-ioport` can't handle CCL file descriptors

    ```
    > Error: There is no applicable method for the generic function:
    >          #<STANDARD-GENERIC-FUNCTION IO-CHANNEL-IOPORT #x18F9347E>
    >        when called with arguments:
    >          (#<STUMPWM::XLIB-IO-LOOP #x19312816> #<FD-BINARY-INPUT-STREAM :ISO-8859-1 (PIPE/5) #x19311FEE>)
    > While executing: #<CCL::STANDARD-KERNEL-METHOD NO-APPLICABLE-METHOD (T)>, in process listener(1).
    > Type :GO to continue, :POP to abort, :R for a list of available restarts.
    > If continued: Try calling it again
    > Type :? for other options.
    1 > :b
     (F6FFED70) : 0 (FUNCALL #'#<#<CCL::STANDARD-KERNEL-METHOD NO-APPLICABLE-METHOD (T)>> #<STANDARD-GENERIC-FUNCTION IO-CHANNEL-IOPORT #x18F9347E> #<STUMPWM::XLIB-IO-LOOP #x19312816> #<FD-BINARY-INPUT-STREAM :ISO-8859-1 (PIPE/5) #x19311FEE>) 135
     (F6FFED98) : 1 (FUNCALL #'#<#<STANDARD-METHOD IO-CHANNEL-IOPORT (T STUMPWM::REQUEST-CHANNEL)>> #<STUMPWM::XLIB-IO-LOOP #x19312816> #<STUMPWM::REQUEST-CHANNEL #x193115DE>) 79
     (F6FFEDAC) : 2 (FUNCALL #'#<#<STANDARD-METHOD IO-LOOP-ADD (STUMPWM::XLIB-IO-LOOP T)>> #<STUMPWM::XLIB-IO-LOOP #x19312816> #<STUMPWM::REQUEST-CHANNEL #x193115DE>) 39
     (F6FFEDBC) : 3 (STUMPWM-INTERNAL-LOOP) 647
     (F6FFEDE8) : 4 (STUMPWM-INTERNAL ":1") 3119
     (F6FFEE3C) : 5 (STUMPWM ":1") 383
     (F6FFEE4C) : 6 (CALL-CHECK-REGS STUMPWM) 247
     (F6FFEE68) : 7 (CHEAP-EVAL (STUMPWM)) 87
     (F6FFEE84) : 8 (FUNCALL #'#<(:INTERNAL CCL::EVAL-STRING CCL::STARTUP-CCL)> "(stumpwm:stumpwm)") 431
     (F6FFEEA8) : 9 (STARTUP-CCL ("home:ccl-init" "home:\\.ccl-init")) 1431
     (F6FFEED8) : 10 (FUNCALL #'#<(:INTERNAL (CCL:TOPLEVEL-FUNCTION (CCL::LISP-DEVELOPMENT-SYSTEM T)))>) 55
     (F6FFEEE8) : 11 (FUNCALL #'#<(:INTERNAL CCL::MAKE-MCL-LISTENER-PROCESS)>) 551
     (F6FFEF34) : 12 (RUN-PROCESS-INITIAL-FORM #<TTY-LISTENER listener(1) [Active] #x18203636> (#<CCL:COMPILED-LEXICAL-CLOSURE # #x1820336E>)) 639
     (F6FFEF78) : 13 (FUNCALL #'#<(:INTERNAL (CCL::%PROCESS-PRESET-INTERNAL (CCL:PROCESS)))> #<TTY-LISTENER listener(1) [Active] #x18203636> (#<CCL:COMPILED-LEXICAL-CLOSURE # #x1820336E>)) 519
     (F6FFEFCC) : 14 (FUNCALL #'#<(:INTERNAL CCL::THREAD-MAKE-STARTUP-FUNCTION)>) 255
    1 >   
    ```

- CLISP starts up fine, but any command requiring an input window errors with `NIL isn't a INT16`. That error also occurs on `master` though, so this PR is not at fault. I opened issue #287 for this. Note that I had to rebase this code onto `master` to even get there, because of #284.

I think using timers on "unsupported" implementations would be better. Just calling the function and completely killing thread safety doesn't seem sane to me, because the entire purpose of using `call-in-main-thread` is that you want some level of safety. Timers can provide a little more safety, and functions like this by definition can't really guarantee *when* your code will get execute, so it seems to be the better option.

lokedhs:
Fair enough. I'll make the appropriate changes to address these two points.

lokedhs:
The timer workaround is now used on platforms that don't have pipes support implemented. I also moved the platform-specific code to `wrappers.lisp` where this code actually seems to belong.

dbjergaard:
Is this ready to merge?

lokedhs:
It should be. I'm using it already and it is stable on SBCL.

wrappers.lisp
```diff
@@ -383,4 +383,56 @@ regarding files in sysfs. Data is read in chunks of BLOCKSIZE bytes."
   #-(or sbcl clisp)
   (error "Unimplemented"))
 
+(defun open-pipe (&key (element-type '(unsigned-byte 8)))
+  "Create a pipe and return two streams. The first value is the input
+stream, and the second valus is the output stream."
```
dbjergaard: valus -> value

stumpwm.lisp
```diff
@@ -219,6 +273,16 @@ The action is to call FUNCTION with arguments ARGS."
        (let ((io (make-instance *default-io-loop*)))
          (io-loop-add io (make-instance 'stumpwm-timer-channel))
          (io-loop-add io (make-instance 'display-channel :display *display*))
+
+         ;; If we have no implementation for the current CL, then
+         ;; don't register the channel.
+         #+sbcl
```
jorams: This should be `#+(or sbcl ccl)` right?

jorams: The answer is *no*. I forgot about the fact that the non-dummy IO loop is currently SBCL only.

jorams:
I tried running with this on SBCL, CCL, ECL and CLISP. It does not (appear to) introduce any new issues, so I'm merging.

PR #288, opened 2016-10-31
Reorganize and clean up the mode line

jorams:
The commit messages should be enough to explain what's happening here.

The external interface didn't change, so this is backwards compatible in that regard.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -166,8 +166,8 @@
       (if win
           (destroy-window win)
           (progn
-            (let ((ml (find-mode-line-window window)))
-              (when ml (destroy-mode-line-window ml))))))))
+            (let ((ml (find-mode-line-by-window window)))
+              (when ml (destroy-mode-line ml))))))))
 
 (defun read-from-keymap (kmaps &optional update-fn)
   "Read a sequence of keys from the user, guided by the keymaps,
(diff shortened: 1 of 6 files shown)


PuercoPop:
Having O(n) lookups where n is the number of monitors (2-4) is ok

:+1:

dbjergaard:
Hi, while this wasn't a controversial PR, I would like a chance to read them over before they're merged.  I get a little nervous when things get re-factored a lot and I miss the changes. 

I don't want to stifle contributions or slow down the development process.  I try to review pull requests within a few days of them being posted.  Sometimes they sit in the queue a while, so if this policy isn't working we can revisit it.

PuercoPop:
@dbjergaard I apologize if I overstepped my bounds. I don't think waiting a few days before merging is a bad policy.

The strategy I mentioned @jorams on IRC was that when one submits a PR we wait for another one to merges the PR (after reviewing it). That way every part of the code was reviewed at least two persons.

dbjergaard:
OK, so I think the best policy:

Let the PR sit for a few days (3?) so I have a chance to review, and if I'm radio silent and one of the other people reviews the PR, it can be merged.  If the change is controversial, I hope that we have a discussion before it gets merged, and in cases like this they can be merged with no friction.

PR #289, opened 2016-11-06
[WIP] StumpUI

jorams:
I've been working on a simplification of all UI in StumpWM, and I've now arrived at a point where getting some feedback on what I'm doing seems like a good idea. It is not ready for a merge.

What I want this to become is a small UI framework for StumpWM, enabling all of the places in the codebase that do something of the sort to be simplified. Documentation is currently incredibly sparse, but I hope the code itself is clear enough to get the picture. More documentation will have to be added before I can consider this ready. The commit messages also document some changes or issues relative to current behaviour.

The CI build is currently failing because the Makefile can't handle ASDF modules. Loading the `stumpwm` system through ASDF should work fine.

Note that I will probably be force-pushing to this branch, so keep that in mind when testing locally.


PuercoPop:
@jorams One thing that should be defined is the scope of StumpUI. Besides windows, should it take care of popup dialogs, or repl input?

The code currently handles popup dialogs, but it appears to name them timed-windows. Although that name is descriptive, it may be a good idea to not chose baroque names. Although GTK uses the term [dialog](https://developer.gnome.org/gtk3/stable/GtkDialog.html) and QT uses the term [MessageBox](https://doc.qt.io/qt-5/qmessagebox.html) so there doesn't appear to be any consensus.

jorams:
@PuercoPop I had some trouble coming up with a good name for `timed-window`, but I now realize that's because I wasn't thinking about it correctly. It is basically a mixin that causes the window to disappear after a specified number of seconds. `window-timeout-mixin` fits much better. I'll remove the superclass and rename the protocol functions to `window-schedule-timeout` and `window-cancel-timeout`. I'm not sure what to do about `timed-window-p`, maybe remove it?

Now for scope, which is tricky but important. When I started working on this I just wanted a common denominator for every graphical window in StumpWM. After the basics worked I gave it a cleaner API, and that's what you're seeing right now. What is currently in StumpUI works at a lower level than "popup dialogs", but a popup dialog is really just a window. The primary thing missing to support a general version of them is being able to handle user input, which is also what's needed to port the input window and the mode line's click hook.

In terms of capabilities I currently think we should draw the line somewhere around the point where StumpUI is able to support the mode line features [listed in our README](https://github.com/stumpwm/stumpwm/blob/15fbc62698c20829754b7206ef602031f05d3c2a/README.md#wishlist). I don't want to create a full-fledged graphics toolkit, because at that point integrating McCLIM would probably be a better idea. That's also why I'm hesitant to include more concrete things, like popup dialogs, as concepts in StumpUI.

So in summary I'm not entirely sure. Input on where to draw the line is greatly appreciated.

message-window.lisp
```diff
@@ -150,47 +46,53 @@ function expects to be wrapped in a with-state for win."
...
+                                  stumpui:timed-window)
+  ()
+  (:default-initargs :event-mask '(:exposure)
+                     :padding 0))
+
 (defun show-frame-indicator (group &optional force)
   (show-frame-outline group)
-  ;; FIXME: Arg, these tests are already done in show-frame-outline
-  (when (find group (mapcar 'screen-current-group *screen-list*))
+  ;; FIXME: Arg, this test is already done in show-frame-outline
```
dbjergaard: can we actually fix this instead of updating the comment?

gravity.lisp
```diff
@@ -0,0 +1,70 @@
+;; Copyright (C) 2016 Joram Schrijver
```
dbjergaard: While I think its important to refactor into separate files, I think this is a bit too granular.  This functionality is really a user UI preference, and should get lumped in with similar preference type code.

package.lisp
```diff
@@ -23,3 +23,35 @@
 
 (defpackage :stumpwm-user
   (:use :cl :stumpwm))
+
+(defpackage :stumpui
+  (:use :cl)
+  (:export
+   ;; Window protocol
```
dbjergaard: When you say protocol, do you mean a stumpwm protocol, or something like icccm?

stumpui/windows.lisp
```diff
@@ -0,0 +1,215 @@
+;;;; Copyright (C) 2016 Joram Schrijver
```
dbjergaard: What happened to window.lisp? Can we find a name that represents the content of this file better? window-protocol.lisp maybe?

stumpui/windows.lisp
```diff
@@ -0,0 +1,215 @@
...
+(defmethod (setf text-window-border-color) (color (window text-window))
+  (setf (xlib:window-border (window-xwin window))
+        color))
+
+(defmethod (setf text-window-border-width) (width (window text-window))
+  (setf (xlib:drawable-border-width (window-xwin window))
+        width))
+
+;;; Timed window --------------------------------------------------------------
+;;; The TIMED-WINDOW class implements the Timed window protocol.
```
dbjergaard: If these are pop-ups lets call them that: 
https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.10

dbjergaard:
You can see some of my comments in the code-review.  I think that abstracting the ui into generic functions is an important first step to making it possible to experiment with different GUI toolkits.  

The community really wants to support non-ascii keybindings and input.  This means utf-8, which stumpwm doesn't support at all. Right now everything is home-grown, and its showing its age.  If we abstract the parts of stumpwm that handle the ui, then it may be possible to use a toolkit that natively supports utf8.

stumpui/windows.lisp
jorams: As noted in my comment above, they aren't popups. I'm going to update it to be called `window-timeout-mixin`, to reflect what it actually is.

message-window.lisp
jorams: We can, by making `show-frame-outline` return whether or not it actually did anything. I updated the comment because "these tests" was confusing in the presence of other tests that aren't done in `show-frame-outline`.

gravity.lisp
jorams: I moved it to a separate file because it's being used in various places, not message-window specific, and because I was going to be using it in StumpUI. After everything has been ported over to StumpUI, I think this should be moved into there.

dbjergaard:
I'm on irc if you want to chat

package.lisp
jorams: I mean protocol in the sense that it is a set of generic functions defining what needs to be implemented. This is fairly common usage in CL, think of the MetaObject Protocol as another example.

stumpui/windows.lisp
jorams: This is `/stumpui/windows.lisp`, not `/window.lisp`. I agree we need a better name, but `window-protocol.lisp` isn't it. This file contains implementations of the protocol, not the actual protocol.

dbjergaard:
ok, I gave this some thought last night and here's what I propose:
1. We create a paulownia branch
2. Merge this there
3. Merge the changes I already made in the paulownia repository
4. The master branch gets put in maintenance mode, only PRs that fix broken builds on non-standard flavors get merged

Thoughts?

jorams:
@dbjergaard What are the relevant changes in the Paulownia repository? Is it usable? I haven't kept up to date with development there.

I'm in favor of branching development, keeping a version of the current codebase—with its restrictions on dependencies—in maintenance mode. I wonder, however, if it might be a better idea to branch that off (call it "1.0") while moving forward with more drastic changes on the master branch. It would more clearly show where the focus of the project lies. (That could also partly be done by changing the *Default branch* in the GitHub settings, of course.)

dbjergaard:
I've been busy with work, but I plan on branching/releasing 1.0.0 and instituting the paulownia changes in the master branch of stumpwm.  Then, we can get this code integrated and move forward.

dbjergaard:
Sorry to spam, but I forgot to address the changes in paulownia that I've implemented:
- Builds image using roswell
- Unit tests using prove
They can be re-integrated with stumpwm. The idea is to have enough unit tests that we can experiment with re-structuring the code without fear of breaking code. I don't know how well that fits into the lisp build cycle though.

dbjergaard:
Can we reorganize this onto its own branch? I'm a little OCD with having PRs open too long.

jorams:
I'm just going to close it until I find time to work on it again.

PR #290, opened 2016-11-08
Thread-safety improvements to the timer code

lokedhs:


stumpwm.lisp
```diff
@@ -227,21 +245,24 @@ The action is to call FUNCTION with arguments ARGS."
...
-
-#-sbcl
-(defun call-in-main-thread (fn)
-  (run-with-timer 0 nil fn))
+  (if *in-main-thread*
+      (funcall fn)
+      ;; ELSE: Schedule the job in the main thread
+      (progn
+        #+call-in-main-thread
+        (progn
```
jorams: I don't think merging these two definitions, ending up with an `if` containing two nested `progn`s, is a good idea. Keeping the two separate definitions, both with the same (simple) condition, and/or using `cond`, seems like a better idea.

lokedhs:
I have split it into two functions now, as suggested.

jorams:
:+1: LGTM

PR #294, opened 2016-12-13
Add which-key-mode

dmh43:
This replaces: https://github.com/stumpwm/stumpwm-contrib/pull/58
out
I've attached a GIF demonstrating the behavior of `which-key-mode`


dmh43:
I've simplified this PR so that it only touches `help.lisp`

PR #295, opened 2016-12-13
Add: `remove-all-hooks`

AeroNotix:
Removes all functions from a hook.

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -128,6 +128,7 @@
           define-frame-preference
           redirect-all-output
           remove-hook
+          remove-all-hooks
           run-hook
           run-hook-with-args
           command-mode-start-message
@@ -682,6 +683,10 @@ display a message whenever you switch frames:
 (defmacro remove-hook (hook fn)
 "Remove the specified function from the hook."
   `(setf ,hook (remove ,fn ,hook)))
+
+(defmacro remove-all-hooks (hook)
+"Remove all functions from a hook"
+  `(setf ,hook NIL))
 
 ;; Misc. utility functions
 


dbjergaard:
Why can't a user just write (setf hook-name nil)?

AeroNotix:
They can, but then why can't a user just write `(setf hook (remove fn hook))`?

dbjergaard:
I really appreciate the code contribution, but to me this adds a level of indirection that isn't needed.  (remove-hook fn hook) is a well defined operation.  It expands to a relatively simple piece of code, but it doesn't require mental gynmastics to operate it.  On the other hand:
(setf hook (remove fn hook))
does.  Clearing a hook by setting it nil is one step (setf hook nil) and is easily mentally parsed.  

Can we compromise and update the manual to explain how to remove hooks using setf?

PuercoPop:
@dbjergaard I think @AeroNotix was being cryptic by asking a rhetorical question. Their point is, we provide APIs to abstract over implementation details. What happens if we change how we store hooks from a list to a custom set implementation (fset or otherwise)? Users would have to rewrite calls to `(setf hook-name nil)`.

We provide some abstractions in that spirit such as `add-screen-mode-line-formatter`. But then again most of the code in StumpWM-contribs didn't used it and instead pushed to `*screen-mode-line-formatter*` instead.

dbjergaard:
@PuercoPop makes a good point.  I'm all for abstraction, but I don't want it at the expense of increasing maintenance overhead. @AeroNotix if you want to re-open this PR, I'll merge it.

dbjergaard:
@AeroNotix I look forward to more PRs :+1:

AeroNotix:
:+1:

PR #297, opened 2016-12-28
Always send WM_TAKE_FOCUS when a window is focused.

lokedhs:
Previously the message was only sent when *CURRENT-EVENT-TIME* was
non-NIL. This caused problems in IntelliJ IDEA when a dialog window
was dismissed using the ESC key which then resulted in the cursor not
being active in the main edit window in the application.


jorams:
LGTM 👍

PR #298, opened 2017-01-05
Floating tile mix

dmh43:
Implements support for float windows in tile groups!
I haven't tested with multiple heads yet.
I also have plans to reorganize some of code between `tile-group.lisp` `floating-group.lisp` and `group.lisp` so that I don't have to reach into `stumpwm.floating-group::float-window`.


dbjergaard:
This is awesome! Give me a bit to review the code and I'll merge it ASAP.

dmh43:
@jorams, I agree for the most part, but I believe that a purely floating group is still useful occasionally, so I'd like to keep that around. I'll spend some time tomorrow cleaning up the organization.

floating-group.lisp
```diff
@@ -135,12 +135,17 @@
   (float-window-align window)
   (focus-window window))
 
+(defmethod group-add-window (group (window float-window) &key &allow-other-keys)
```
dmh43: I'd like to refactor some of the code to remove the use of multimethods where it doesn't seem entirely appropriate. What do you think? See: https://github.com/stumpwm/stumpwm/issues/300

dmh43:
@jorams I've cleaned up the code a bit. I'm not entirely satisfied with the organization so I've made https://github.com/stumpwm/stumpwm/issues/301 to address the code organization in the project in general. Are you alright addressing the code organization concerns in a separate PR?

jorams:
I responded to the issues you created. In summary, I'm definitely alright adressing the code organization in a separate PR, but removal of the generic functions less so. At least `group-add-window` should definitely stay a generic function.

floating-group.lisp
dmh43: ^ this has been resolved. Organization will be addressed in a separate PR.

dbjergaard:
is this good to go?

dmh43:
@dbjergaard @jorams @PuercoPop I've been using this daily and haven't run into any issues so far. If there aren't any more comments on the code then I think it's good to go.

Also, I haven't tested this with multiple heads, but just by looking at the code I dont expect any issues.

dbjergaard:
@jorams please merge this when you are happy.

jorams:
Alright, merging. The one thing that still stands out is that the floating group should be merged back into the `stumpwm` package, but that can wait.

group.lisp
```diff
@@ -101,6 +101,14 @@ needs to redraw anything on it, this is where it should do it."))
   (:documentation "When a head or its usable area is resized, this is
 called. When the modeline size changes, this is called."))
 
+(defmethod group-current-window (group)
+  (screen-focus (group-screen group)))
```
jorams: This is causing an issue. The "current window" in a group can now be a window that's not actually in that group, which goes against expectations in other parts of the codebase. (Currently I'm looking at `group-forward-with-window`)

dmh43: Ah, thanks for catching this! I think the best way to address this would be to keep track of the current window on the group object itself. What do you think?

jorams: I think so too, and that's why I opened #324 :)

PR #299, opened 2017-01-06
Stop ignoring nonexistent modifiers in bindings

### LINKED ISSUE #296: "x not bound"
I have been running into a very frustrating problem for some time. When I plug in an external keyboard and start typing,

I get an "x not bound" message for every key press x for which there exists a H-x binding (H is hyper)  

This doesn't seem to happen with the laptop's built-in keyboard

I've run xev, and pressed "x" on both keyboards. Both show the same keysym, but only on the 2nd keyboard I get the "x not bound" message.

I've checked the code, it seems that "x not found" should only happen on describe-key:

```lisp
(defcommand describe-key (keys) ((:key-seq "Describe Key: "))
"Either interactively type the key sequence or supply it as text. This
command prints the command bound to the specified key sequence."
  (let ((cmd (loop for map in (top-maps)
                   for cmd = (lookup-key-sequence map keys)
                   when cmd return cmd)))
    (if cmd
        (message "~{~a~^ ~} is bound to \"~a\"." (mapcar 'print-key keys)  cmd)
        (message "~{~a~^ ~} is not bound." (mapcar 'print-key keys)))))
```

I also sometimes notice "x not found" even on the laptop keyboard, especially when the cpu is so busy that it lags behind the X input (keypresses, etc).

Could this be a threading issue?

jorams:
When a binding requires a modifier, but there is no actual key for that
modifier, it should simply not be bound. Previously the key without that
modifier would be bound instead, causing unexpected behaviour.

Fixes #296 

It would be great for this to receive a little more testing to make sure it doesn't break anything unexpected.

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -694,19 +694,29 @@ and bottom_end_x."
                                  (eql (key-keysym key) (xlib:keycode->keysym *display* code 1)))
                             (add-shift-modifier key)
                             key)))
-                   (xlib:grab-key w code
-                                  :modifiers (x11-mods key) :owner-p t
-                                  :sync-pointer-p nil :sync-keyboard-p nil)
-                   ;; Ignore capslock and numlock by also grabbing the
-                   ;; keycombos with them on.
-                   (xlib:grab-key w code :modifiers (x11-mods key nil t) :owner-p t
-                                  :sync-keyboard-p nil :sync-keyboard-p nil)
-                   (when (modifiers-numlock *modifiers*)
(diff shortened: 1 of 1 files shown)


jorams:
I just woke up and realized that this wouldn't solve the problem anyway. It would only stop bindings that don't have any modifiers that *do* exist. For example, `C-H-a` would still be bound to `C-a`.

dbjergaard:
Can we close this if we aren't goign to merge it?

PR #303, opened 2017-01-10
Refactor ADD-GROUP

PuercoPop:
Use ASSERT instead of IF to check for conditions to be met. Split the retrieval
or creation of the group into a separate function.

Another alternative would be to use flet to define ensure-group as a local function, but in my opinion local functions tend to clutter the function body.

Another thing to consider is if it would be a good idae to move the group creation to its own MAKE-GROUP constructor. It appears that add-group is the only place in the code base where groups are instantiated.

Something I'm left wondering is, would it be wise to allow a new group name to be provided when the assertion fails?

### DIFF
--- a/group.lisp
+++ b/group.lisp
@@ -288,6 +288,20 @@ Groups are known as \"virtual desktops\" in the NETWM standard."
       (netwm-update-groups screen)
       (netwm-set-group-properties screen))))
 
+(defun %ensure-group (group-name group-type screen)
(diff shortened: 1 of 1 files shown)


dbjergaard:
If the assert triggers does it hang stumpwm?

dbjergaard:
Nevermined they are semantically similar. I just worry about the information propagating to the user properly, but  I don't think the old code addressed that issue either.

PR #304, opened 2017-01-12
ensure a key's modifiers exist before grabbing it

### LINKED ISSUE #296: "x not bound"

erjoalgo:
fix grabbing wrong key when modifier is missing
fixes #296

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -683,15 +683,17 @@ and bottom_end_x."
              (let ((key (copy-structure key)))
                (setf (key-shift key) t)
                key))
+           (key-modifiers-exist-p (key)
+             (and
+              (or (not (key-meta key)) (modifiers-meta *modifiers*))
+              (or (not (key-alt key)) (modifiers-alt *modifiers*))
+              (or (not (key-hyper key)) (modifiers-hyper *modifiers*))
+              (or (not (key-super key)) (modifiers-super *modifiers*))))
            (grabit (w key)
              (loop for code in (multiple-value-list (xlib:keysym->keycodes *display* (key-keysym key))) do
                ;; some keysyms aren't mapped to keycodes so just ignore them.
-               (when (and code
-                             ;;temporary hack to make sure H-a
-                             ;;doesn't cause "a" to be grabbed
-                             ;;when there is no H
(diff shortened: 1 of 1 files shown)


window.lisp
```diff
@@ -683,10 +683,17 @@ and bottom_end_x."
...
                (setf (key-shift key) t)
                key))
+           (key-modifiers-exist-p (key)
+             (and
+              (or (not (key-meta key)) (modifiers-meta *modifiers*))
+              (or (not (key-alt key)) (modifiers-alt *modifiers*))
+              (or (not (key-hyper key)) (modifiers-hyper *modifiers*))
+              (or (not (key-super key)) (modifiers-super *modifiers*))))
            (grabit (w key)
              (loop for code in (multiple-value-list (xlib:keysym->keycodes *display* (key-keysym key))) do
```
PuercoPop: We could rewrite the loop to use a when clause which would make clear that the when applies to the whole of the loop 'body' ie.
```lisp
(loop for code in ...
      when (and code (key-modifiers-exist-p key))
        do
```

Issue #292, opened 2016-11-24
Add interactive command macro


caioaao:
Implemented it to navigate between frames without having to type the prefix again (eg.: to move focus to two frames to the left, I have to type `C-q o l l` instead of `C-q l C-q l`), just like `resize` does.

My implementation is [here](https://github.com/caioaao/stumpwm.d/tree/master/modules/icommand), but I'd like some input if it's worth integrating it with stumpwm's project or just leave it as a standalone module (also I'm very new to CL, so I wouldn't know how to do any of the two options properly).

dbjergaard:
I think this is an important abstraction, and I'm interesting in putting it in stumpwm proper.  Here's what I would want in PR to merge:
1. This goes in its own file in `stumpwm/icommand.lisp` 
2. You can merge the file into stumpwm instead of having its own package, just modify stumpwm.asd as necessary
3. Restructure iresize to use your new macro instead of the current code
4. Update the manual to include documentation describing icommand and pointing to iresize as an example of how users can implement the macro in practice

I know this seems like a tall order, but I can see the power of what you've implemented and I want to make sure it is properly exposed to everyone who uses stumpwm.
We can clean up any non-lispy idioms in the code review of the PR.

caioaao:
Sorry for the long delay, I've been too busy during December. I'll try doing it during next weeks. I still have to figure how to bootstrap and share state between commands when an interactive command "session" starts, but I think I have a simple solution to that.

caioaao:
What I've done so far: https://github.com/stumpwm/stumpwm/commit/9615e3d84b4390f647e0897ba1bddb611a7d1235

Dos are still missing, but you can see how this works on the new `iresize` implementation.

dbjergaard:
I'm going to close this since the PR is now open and being reviewed.

PR #306, opened 2017-01-14
Initial interactive command implementation

caioaao:
As discussed in #292

### DIFF
--- a/iresize.lisp
+++ b/iresize.lisp
@@ -27,60 +27,44 @@
 
 (export '(*resize-increment*
           iresize
-          abort-iresize
-          exit-iresize))
+          setup-iresize))
 
 (defvar *resize-increment* 10
   "Number of pixels to increment by when interactively resizing frames.")
 
 (defun set-resize-increment (val)
-  (setf *resize-increment* val)
-  (update-resize-map))
+  (setf *resize-increment* val))
 
-(defun update-resize-map ()
-  (let ((m (or *resize-map* (setf *resize-map* (make-sparse-keymap)))))
-    (let ((i *resize-increment*))
-      (labels ((dk (m k c)
-                 (define-key m k (format nil c i))))
-        (dk m (kbd "Up") "resize 0 -~D")
-        (dk m (kbd "C-p") "resize 0 -~D")
-        (dk m (kbd "p") "resize 0 -~D")
-        (dk m (kbd "k") "resize 0 -~D")
-
-        (dk m (kbd "Down") "resize 0 ~D")
-        (dk m (kbd "C-n") "resize 0 ~D")
-        (dk m (kbd "n") "resize 0 ~D")
-        (dk m (kbd "j") "resize 0 ~D")
-
-        (dk m (kbd "Left") "resize -~D 0")
-        (dk m (kbd "C-b") "resize -~D 0")
-        (dk m (kbd "b") "resize -~D 0")
-        (dk m (kbd "h") "resize -~D 0")
-
-        (dk m (kbd "Right") "resize ~D 0")
-        (dk m (kbd "C-f") "resize ~D 0")
-        (dk m (kbd "f") "resize ~D 0")
-        (dk m (kbd "l") "resize ~D 0")
-        (define-key m (kbd "RET") "exit-iresize")
-        (define-key m (kbd "C-g") "abort-iresize")
-        (define-key m (kbd "ESC") "abort-iresize")))))
-
-(update-resize-map)
-
-(defcommand (iresize tile-group) () ()
-  "Start the interactive resize mode. A new keymap specific to
-resizing the current frame is loaded. Hit @key{C-g}, @key{RET}, or
-@key{ESC} to exit."
+(defun single-frame-p ()
+  "Checks if there's only one frame."
   (let ((frame (tile-group-current-frame (current-group))))
-    (if (atom (tile-group-frame-head (current-group) (frame-head (current-group) frame)))
-        (message "There's only 1 frame!")
-        (progn
-          (when *resize-hides-windows*
-            (dolist (f (head-frames (current-group) (current-head)))
-              (clear-frame f (current-group))))
-          (message "Resize Frame")
-          (push-top-map *resize-map*)
-          (draw-frame-outlines (current-group) (current-head))))))
+    (atom (tile-group-frame-head (current-group)
+                                 (frame-head (current-group)
+                                             frame)))))
+
+(defun abort-resize-p ()
+  "Resize is only available if there's more than one frame."
+  (when (single-frame-p)
+    (message "There's only 1 frame!")
+    t))
+
+(defun setup-iresize ()
+  "Start the interactive resize mode."
+  (when *resize-hides-windows*
+    (dolist (f (head-frames (current-group) (current-head)))
+      (clear-frame f (current-group))))
+  (draw-frame-outlines (current-group) (current-head)))
+
+(defcommand resize-direction (d)
+  (:direction)
+  (let* ((formats '((:up . "0 -~D")
+                    (:down . "0 ~D")
+                    (:left . "-~D 0")
+                    (:right . "~D 0")))
+         (deltas (format nil (cdr (assoc (princ d) formats))
+                         *resize-increment*))
+         (to-be-run (concatenate 'string "resize " deltas)))
+    (run-commands to-be-run)))
 
 (defun resize-unhide ()
   (clear-frame-outlines (current-group))
@@ -95,15 +79,26 @@ resizing the current frame is loaded. Hit @key{C-g}, @key{RET}, or
       (when (current-window)
         (focus-window (current-window))))))
 
-(defcommand (abort-iresize tile-group) () ()
-  "Exit from the interactive resize mode."
-  (resize-unhide)
-  (message "Abort resize")
-  ;; TODO: actually revert the frames
-  (pop-top-map))
+(define-interactive-keymap (iresize tile-group) (:on-enter #'setup-iresize
+                                                 :on-exit #'resize-unhide
+                                                 :abort-if #'abort-resize-p)
 
-(defcommand (exit-iresize tile-group) () ()
-  "Exit from the interactive resize mode."
-  (resize-unhide)
-  (message "Resize Complete")
-  (pop-top-map))
+  ((kbd "Up") "resize-direction up")
+  ((kbd "C-p") "resize-direction up")
+  ((kbd "p") "resize-direction up")
+  ((kbd "k") "resize-direction up")
+
+  ((kbd "Down") "resize-direction down")
+  ((kbd "C-n") "resize-direction down")
+  ((kbd "n") "resize-direction down")
+  ((kbd "j") "resize-direction down")
+
+  ((kbd "Left") "resize-direction left")
+  ((kbd "C-b") "resize-direction left")
+  ((kbd "b") "resize-direction left")
+  ((kbd "h") "resize-direction left")
+
+  ((kbd "Right") "resize-direction right")
+  ((kbd "C-f") "resize-direction right")
+  ((kbd "f") "resize-direction right")
+  ((kbd "l") "resize-direction right"))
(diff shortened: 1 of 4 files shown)


icommand.lisp
```diff
@@ -0,0 +1,61 @@
...
+         (exit-command (concatenate 'string "exit-" (symbol-name command)))
+         (m-name (gensym "m")))
+    `(let ((,m-name (make-sparse-keymap)))
+       ,@(loop for keyb in key-bindings
+            collect `(define-key ,m-name ,@keyb))
+       (define-key ,m-name (kbd "RET") ,exit-command)
+       (define-key ,m-name (kbd "C-g") ,exit-command)
+       (define-key ,m-name (kbd "ESC") ,exit-command)
+       (defcommand ,name () ()
+         ,(concatenate 'string "Starts interactive command \"" ,(symbol-name command) "\"") ;; TODO Add docstring support
```
PuercoPop: There is an extra comma before symbol-name. The code fails to compile due to that.

icommand.lisp
```diff
@@ -38,22 +38,22 @@
...
   (pop-top-map))
 
-(defmacro deficommand (name key-bindings &key pre-start-fn cleanup-fn)
+(defmacro deficommand (name (&key bootstrap-fn cleanup-fn) &body key-bindings)
   "Declare an interactive command." ;; TODO
   (let* ((command (if (listp name) (car name) name))
-         (exit-command (concatenate 'string "exit-" (symbol-name command)))
-         (m-name (gensym "m")))
-    `(let ((,m-name (make-sparse-keymap)))
+         (exit-command (concatenate 'string "EXIT-" (symbol-name command)))
```
PuercoPop: Here we can use format instead`(format nil "EXIT-~A" command)`, which doesn't need thecall to symbol-name. (Or alexandria:symbolicate which also saves as the call to intern below)

iresize.lisp
```diff
@@ -71,24 +72,25 @@
...
-     ((kbd "b") "resize-direction left")
-     ((kbd "h") "resize-direction left")
-
-     ((kbd "Right") "resize-direction right")
-     ((kbd "C-f") "resize-direction right")
-     ((kbd "f") "resize-direction right")
-     ((kbd "l") "resize-direction right"))
-  :pre-start-fn #'setup-iresize
-  :cleanup-fn #'resize-unhide)
+
```
PuercoPop: Below I've pasted a more idiomatic indentation style (If you are using Emacs, just C-M-Space to select the whole form and then M-\ to indent)
```lisp
(deficommand (iresize tile-group) (:bootstrap-fn #'setup-iresize
                                   :cleanup-fn #'resize-unhide)
  ((kbd "Up") "resize-direction up")
  ((kbd "Up") "resize-direction up")
  ((kbd "C-p") "resize-direction up")
  ((kbd "p") "resize-direction up")
  ((kbd "k") "resize-direction up")

  ((kbd "Down") "resize-direction down")
  ((kbd "C-n") "resize-direction down")
  ((kbd "n") "resize-direction down")
  ((kbd "j") "resize-direction down")

  ((kbd "Left") "resize-direction left")
  ((kbd "C-b") "resize-direction left")
  ((kbd "b") "resize-direction left")
  ((kbd "h") "resize-direction left")

  ((kbd "Right") "resize-direction right")
  ((kbd "C-f") "resize-direction right")
  ((kbd "f") "resize-direction right")
  ((kbd "l") "resize-direction right"))
```

icommand.lisp
```diff
@@ -38,22 +38,22 @@
   (message "~s finished" name)
   (pop-top-map))
 
-(defmacro deficommand (name key-bindings &key pre-start-fn cleanup-fn)
+(defmacro deficommand (name (&key bootstrap-fn cleanup-fn) &body key-bindings)
```
PuercoPop: Some minor things, As joshe mentioned on IRC, `defcommand` has interactive-args which makes the name confusing. Also having only one different letter in the middle of the word makes it hard to differentiate from `defcommand` when skimming. Other options would be `defhydra`, or `define-interactive-keymap` but I'm not convinced on either of them. @dbjergaard What do you think?  

Another, nitpick, How about the names, on-entry and on-exit instead of bootstrap-fn and cleanup-fn

icommand.lisp
```diff
@@ -38,22 +38,22 @@
...
   "Declare an interactive command." ;; TODO
```
PuercoPop: What does this TODO refer to?

icommand.lisp
```diff
@@ -38,22 +38,22 @@
...
+            collect `(define-key ,keymap ,@keyb))
+       (define-key ,keymap (kbd "RET") ,exit-command)
+       (define-key ,keymap (kbd "C-g") ,exit-command)
+       (define-key ,keymap (kbd "ESC") ,exit-command)
        (defcommand ,name () ()
-         ,(concatenate 'string "Starts interactive command \"" ,(symbol-name command) "\"") ;; TODO Add docstring support
-         ,(let ((start-code `(icommand-enter-interactive-mode ,m-name ,(symbol-name command))))
-            (if pre-start-fn
-                `(when (funcall ,pre-start-fn)
+         ,(concatenate 'string "Starts interactive command \"" (symbol-name command) "\"") ;; TODO Add docstring support
```
PuercoPop: The easiest way to add docstring support would be to use alexandria:parse-body, but we could do that in a subsequent PR.

PuercoPop: Format would more  succinct here `(format nil "Starts interactive command \"~A\"" command) `

icommand.lisp
```diff
@@ -0,0 +1,61 @@
...
+
+(in-package #:stumpwm)
+
+(export '(icommand-enter-interactive-mode
+          icommand-exit-interactive-mode
+          deficommand))
+
+(defun icommand-enter-interactive-mode (kmap name)
+  "Enters interactive mode" ;; TODO
+  (message (format nil "~s started" name))
```
PuercoPop: message calls format nil underneath so `(message "~A started." name` would suffice. It is common lisp style to uppercase format directives

icommand.lisp
```diff
@@ -0,0 +1,61 @@
...
+         (keymap (gensym "m")))
+    `(let ((,keymap (make-sparse-keymap)))
+       ,@(loop for keyb in key-bindings
+            collect `(define-key ,keymap ,@keyb))
+       (define-key ,keymap (kbd "RET") ,exit-command)
+       (define-key ,keymap (kbd "C-g") ,exit-command)
+       (define-key ,keymap (kbd "ESC") ,exit-command)
+       (defcommand ,name () ()
+         ,(concatenate 'string "Starts interactive command \"" (symbol-name command) "\"") ;; TODO Add docstring support
+         ,(let ((start-code `(icommand-enter-interactive-mode ,keymap ,(symbol-name command))))
```
PuercoPop: you can pass the symbol directly to `icommand-enter-interactive-mdoe`, no need to call symbol-name here

PuercoPop:
Another thing to simply the implementation. When one wants to conditionally insert code depending on conditions known at compile time, like if a keyword argument was provided, we can use  ,@(when ...) with the code we would like to insert in a list. This way if the condition is met the code we want to insert is returned (the outer list was 'spliced') and if the condition is not met the when returns nil which when spliced disappears. ie.

```lisp
,@(when <condition>
  (list <code>)
```

When using this in the implementation of deficommand it becomes

```lisp
`(let ...

(defcommand ,name () ()
  ,(format nil "Starts interactive command \"~A\"" command)
  ,@(when bootstrap-fn
      `((funcall bootstrap-fn)))
  `(icommand-enter-interactive-mode ,keymap ,command))

(defcommand ,exit-command () ()
  ,@(when cleanup-fn
      `((funcall ,cleanup-fn)))
  (icommand-exit-interactive-mode ,command)))
```

As you can see, there is no need for `(let ((start-code )`. Also do you intend that if the boostrap-fn returns nil the command would do nothing? (I assumed not so I 'fixed' it in the example above.

PuercoPop:
Btw @caioaao I've pushed a branch (https://github.com/PuercoPop/stumpwm/tree/feedback-interactive-commands) with most of the feedback in case you prefer to read it as whole instead of by snippets. around with the code and macro-expand it to see how to `,@(when ..)` idiom works

caioaao:
@PuercoPop about the `bootstrap-fn` making the command do nothing when it returns nil: I did that on purpose, so I could abort `iresize` when there's only one frame, but I don't think this is a good way of doing it. Suggestions?

I think calling the command `defhydra` is not good as well: `hydra` is a well known emacs package that works very differently, and we only share the same motivation. 

I'll try and fix the rest of the comments during this week. Thanks a lot!

PuercoPop:
@caioaao How about  adding a new keyword argument `:abort-if`. Regarding the name, how about define-interactive-keymap/defikeymap? Also do we need to export {enter,exit}-interactive-mode?

dbjergaard:
I vote for define-interactive-keymap and I agree with *not* naming it hydra even if this is inspired by that package.

dbjergaard:
I'm ok with merging this when @PuercoPop is.

dbjergaard:
With the caveat that I want to see documentation in the manual!

PuercoPop:
@dbjergaard Ok, I'm waiting for @caioaao to push the renaming :nail_care:. I'll wait until Sunday and then fix any unaddress remaning issue (including documentation) then merge. Is that ok with you?

dbjergaard:
@PuercoPop I'm happy with this as it is now, merge when you are ready.

PuercoPop:
There are a couple of details. @caioaao I'm sorry for the repeated request for changes, thanks for your patience :tada: 

There are still a few things to changes:

- Drop remove the calls to symbol-name, but you'll have to add a quote. ie `,(symbol-name command)` becomes `,command`
- Add a line between the `defcommand`s in `define-interactive-keymap` for readability.
- Instead of nesting when forms for abort-if use have the when for the abort-if occur earlier and exit with return-from. That way the expansion is cleaner, i.e. no `(when t ...)`
- `resize-p` did two things. Decided if iresize should abort and presented a message explaining why it had aborted. I would prefer we decouple this into `:abort-if` and `:on-abort`. With abort-if taking predicate function and on-abort taking forms to execute.
- Instead of resize-p, the name only-frame-p would be more apt, also it was lacking a docstring. 

I've addressed all these issues on this [branch](https://github.com/stumpwm/stumpwm/commit/8d31d9b35d5ccc46b5b1b75d420a609b7793106c). However even though I only used amend and rebase (reword and fixup) I'm listed as the under Commit: and @caioaao under Author:. I don't want to steal attribution from @caioaao. So @caioaao could you add the modifications as shown in the branch and squash all the commits into one so we can merge?

I thank you again for your work and patience

PuercoPop:
One last thing, the copyright at the start of interactive-keymap.lisp should read 2016, 2017

caioaao:
Also, thank you @PuercoPop for walking me through the issues on my code! About splitting `abort-if`: idk if it's  a good idea, but I don't really like calling `message` inside a predicate either. The problem is when there's more than one reason to abort the command. The client would have to work around the two functions (sharing state, maybe?) just so it could show a more ellaborate message. I also think it's overkill refactoring the code to do some sort of error piping from `abort-if` to `on-abort` In the end I thought a little "impure" predicate would be the best option. wdyt?

PuercoPop:
@caioaao Yes I'm not sold on `:on-abort`. Especially because we already have an `:on-exit` option; and isn't aborting exiting? One way to approach this situation is to have the on-exit function accept one argument which would be a flag signaling if it was called from abort-if or not. But lets not delay the merge any longer. But we should avoid the nested on either way, just simply `define-interactive-keymap` as following and I'll merge. We can work on improving the API afterwards 

```lisp
(defcommand ,name () ()
  ,(format nil "Starts interactive command \"~A\"" command) ;; TODO Add docstring support
  ,@(when abort-if `((when (funcall ,abort-if)
                       (return-from ,command))))

  ,@(when on-enter `((funcall ,on-enter)))
  (enter-interactive-keymap ,keymap ',command))
```

PR #307, opened 2017-01-18
Don't assume ASDF:SYSTEM-DEPENDS-ON exists

PuercoPop:
On ASDF versions prior to 3.15 we can't assume that the symbol
SYSTEM-DEPENDS-ON is exported in the ASDF package. If it is not present the
reader signals an error even though the code won't be executed.

As reported by user jmccabe on IRC. They were using SBCL 1.1.14 with ASDF 3.0.3 which are 3 years old on Linux Mint 17.


jorams:
👍

PR #316, opened 2017-01-22
toggle-always-show for float windows

dmh43:
This PR is branched off https://github.com/stumpwm/stumpwm/pull/298
This allows one to display a float window on all groups. Combined with https://github.com/stumpwm/stumpwm/pull/302 should give some really nice support for floating windows. I watched lord of the rings using this functionality to create picture in picture viewer with Xephyr :) 
img

I considered adding support for always showing tile windows as well, but that is much more complex, and I'm not too sure that it's a very common use case at all.

### DIFF
--- a/group.lisp
+++ b/group.lisp
@@ -39,6 +39,8 @@
 (defvar *default-group-type* 'tile-group
   "The type of group that should be created by default.")
 
+(defvar *always-show-windows* ())
+
 (defclass group ()
   ((screen :initarg :screen :accessor group-screen)
    (windows :initform nil :accessor group-windows)
@@ -109,6 +111,11 @@ called. When the modeline size changes, this is called."))
       (window-head (group-current-window group))
       (first (screen-heads (group-screen group)))))
 
+(defmethod group-delete-window (group window)
+  (when (find window *always-show-windows*)
+    (disable-always-show-window window (current-screen)))
+  (call-next-method))
+
 (defun current-group (&optional (screen (current-screen)))
   "Return the current group for the current screen, unless
 otherwise specified."
@@ -161,8 +168,6 @@ at 0. Return a netwm compliant group id."
   (let ((screen (group-screen group)))
     (position group (sort-groups screen))))
 
-(defvar always-show-windows ())
-
 (defun switch-to-group (new-group)
   (let* ((screen (group-screen new-group))
          (old-group (screen-current-group screen)))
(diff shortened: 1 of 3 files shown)


PuercoPop:
A minor note around renaming the accesor window-group*. When you want to define more than straightforward accessors it is best practice to don't include them in the slot definition. And you should use slot value when defining them. However for the functionality you need for window-group would be trivially implemented with an around method. 

I'm sorry I haven't done a proper review as I'm not as familiar with this part of the code base.

dmh43:
Thanks for the feedback! I was wondering about a better way to achieve that.

PuercoPop:
The code would look something like
```lisp
(defmethod window-group :around (window)
  (if (find window *always-shown-windows*)
      (stumpwm:current-group)
      (call-next-method)))
```

It may be useful to specialize on a class for the window parameter.

<hr>

I'm not sold on `*always-shown-windows*`, how about `*pegged-windows*`? Don't forget to add a docstring to the variable as well.

dmh43:
Oh wow, that's really cool, I hadn't realized that the slot accessors were generic functions. Latest commit uses something similar to the code snippet in your last comment. I'll wait on other suggestions for naming of the list before doing the rename. I am open to suggestions.

dbjergaard:
@PuercoPop @dmh43 when this is ready let me know and I'll merge it.

dmh43:
I'd like some suggestions on renaming: `*always-show-windows*` which is a list containing the windows to show on all desktops. I like `*pegged-windows*`, `*pinned-windows*` or `*always-shown-windows*

PuercoPop:
@dbjergaard Normally I don't look into code regarding floating windows because @jorams is more familiar with that part of the codebase. But I'll do a proper review on the weekend.

@dmh43 Sorry for taking so long to review your PR

group.lisp
```diff
@@ -179,9 +187,48 @@ at 0. Return a netwm compliant group id."
...
 
+(defun copy-window-to-group (group window)
+  (setf (window-number window) (find-free-window-number group))
+  (push window (group-windows group))
+  (group-add-window group window))
+
+(defun always-show-window (window screen)
+  (let ((groups-to-add-w-to (remove (current-group) (screen-groups screen))))
+    (mapc (lambda (group) (copy-window-to-group group window))
+          groups-to-add-w-to))
```
PuercoPop: `other-groups` (or `other-groups-in-screen` if one feels like the extra specificity is worth the verbosity) would be a better name than `groups-to-a-w-to`. Also we can use alexandria's rcurry to avoid the lambda clutter. Also `(map nil ...)` communicates that we only care about the side-effects more clearly.

```
(map nil (alexandria:rcurry #'copy-window-to-group window)
      other-groups))
```

or because the variable is only used in one place we can do w/o the let.

```
(map nil (alexandria:rcurry #'copy-window-to-group window)
      (remove (current-group) (screen-groups screen)))
```


(mapc (alexandria:rcurry #'copy-window-to-group window)
      (remove (current-group) (screen-groups screen)))

dbjergaard:
@PuercoPop I usually defer to whoever actively replied first ;) feel free to tag @jorams if he would be a better fit to review the code.

dbjergaard:
I'm happy with this so its getting merged, I prefer mapc in this context so I'm letting it stay. @PuercoPop, @dmh43 if I missed something you want addressed then lets address it with another PR.

PR #318, opened 2017-01-27
Move MAKE-XLIB-WINDOW out wrappers.lisp

PuercoPop:
Remove support for other implementations other than SBCL. Use XLIB:WINDOW-ID
instead of accessing the direct slot of the DRAWABLE.

Moved the function to where it is used.
I would like to remove the function altogether but I have to investigate if the issue it is working around of (the cache returning a pixmap instead of a window) is still present in CLX.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -585,6 +585,12 @@ the window in it's frame."
   ;; Pass click to client
   (xlib:allow-events *display* :replay-pointer time))
 
+(defun make-xlib-window (xobject)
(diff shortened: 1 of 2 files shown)


jorams:
It looks like this commit: sharplispers/clx@86dea89fe6f1fee4720ed8d8fea34ea6c2e06d7b should have fixed the cache issue. See also <https://sourceforge.net/p/sbcl/mailman/message/19823960/>

PuercoPop:
Sweat, nice find! We can delete the code altogether then

PR #319, opened 2017-01-27
Wrappers cleanup

PuercoPop:
Starting to remove code for non-sbcl implementations. In some cases removed the functions altoghther, in others left the function but removed all but the SBCL implementation

### DIFF
--- a/wrappers.lisp
+++ b/wrappers.lisp
@@ -401,21 +401,5 @@ stream, and the second value is the output stream."
   #-(or sbcl ccl)
   (error "Unsupported CL implementation"))
 
-(defun make-lock ()
-  #+sbcl
-  (sb-thread:make-mutex)
-  #+ccl
-  (ccl:make-lock "Anonymous lock")
-  #+(and clisp mt)
-  (mt:make-mutex)
-  #+lispworks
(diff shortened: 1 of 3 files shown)


wrappers.lisp
```diff
@@ -184,24 +89,9 @@
...
-  #+lispworks (let ((dbg::*debugger-stack*
-                      (dbg::grab-stack nil :how-many frames))
-                    (*debug-io* *standard-output*)
-                    (dbg:*debug-print-level* nil)
-                    (dbg:*debug-print-length* nil))
-                (dbg:bug-backtrace nil))
-  #-(or sbcl clisp ccl lispworks) (write-line "Sorry, no backtrace for you."))
+  (sb-debug:print-backtrace :count frames :stream *standard-output*))
 
 (defun bytes-to-string (data)
```
dmh43: What about for these string manipulation functions? Did you leave those unmodified intentionally?

PuercoPop: Yes. Because it is not as straight forward how to clean them up. It may be possible to only leave to functions utf8-to-octets and octets-to-utf8, but I haven't gone through their uses yet. See issue #320

PR #322, opened 2017-02-01
Introduce the DEFINE-FRAME-PREFERENCES macro

lokedhs:
Redefine all window placement rules. This macro removes all defined
placement rules and replaces them with the set of rules defined in
GROUPS. The benefit to using this function instead of a set of calls
to DEFINE-FRAME-PREFERENCE is that reloading the configuration will
not result in duplicate entries in *WINDOW-PLACEMENT-RULES*.

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -125,6 +125,7 @@
           data-dir-file
           dformat
           define-frame-preference
+          define-frame-preferences
           redirect-all-output
           remove-hook
           remove-all-hooks
@@ -1117,6 +1118,27 @@ The window's title must match @var{title}.
          (declare (ignore create restore class instance type role title))
          (push (list* ,target-group frame-number raise lock keys)
                *window-placement-rules*)))))
+
+(defmacro define-frame-preferences (&body groups)
+  "Redefine all window placement rules. This macro removes all defined
+placement rules and replaces them with the set of rules defined in
+GROUPS. The benefit to using this function instead of a set of calls
+to DEFINE-FRAME-PREFERENCE is that reloading the configuration will
+not result in duplicate entries in *WINDOW-PLACEMENT-RULES*.
+
+Each rule takes the following form:
+
+@example
+\(target-group frame-rule...)
+@end example
+
+Where each FRAME-RULE has the same form as for DEFINE-FRAME-PREFERENCE."
+  `(progn
+     (clear-window-placement-rules)
+     ,@(loop
+         for (target-group . rules) in groups
+         collect `(define-frame-preference ,target-group
+                    ,@rules))))
 
 (defun clear-window-placement-rules ()
   "Clear all window placement rules."


scottjad:
I recommend not merging this pull request.

The name define-frame-preferences is too similar to the name define-frame-preference, and a user would not have any idea by the name that there is a big difference like one clears the placement rules and the other doesn't.

Since this behavior is easy for the user to implement on their own, I don't think it warrants the potential confusion of two similarly named but subtly different functions.

People who want this behavior can do this:

``
(progn
  (clear-window-placement-rules)
  (define-frame-preference ...)
  (define-frame-preference ...))
``

I am sympathetic to the notion that define-frame-preference could check if a rule already exists before adding it, but I don't know what consequences that might have and I think it should be considered separately.

PuercoPop:
How about REDEFINE-FRAME-PREFERENCES?

dbjergaard:
Could we address the main thrust of this PR some other way? Like adding a check to define-frame-preference to make sure that the preference is unique? Otherwise I tend to agree that this confuses the API (its a one letter difference function that does more than it says on the tin). At the very least it needs to be renamed to something else.

lokedhs:
The main reason for me proposing this fix is that `DEFINE-FRAME-PREFERENCE` has a name which is not accurate on several levels. First and foremost, it's not actually defining anything, but rather adds a rule to `*WINDOW-PLACEMENT-RULES*`. Secondly, there is the problem that adding the same rules multiple times causes duplicate entries, where the previous entries can interfere with other entries.

The fact that the current name is, in my opinion, not very good, means that a proper solution would necessarily not be backwards compatible. If so, then my proposal is to eliminate `DEFINE-FRAME-PREFERENCE` and only keep `DEFINE-FRAME-PREFERENCES`. A replacement for the former could be added, but having more reasonable semantics.

If we want to stay backwards compatible, then I agree that my proposed macro isn't really necessary, and thus shouldn't be included (as least not with the current name).

dbjergaard:
ok so we can rename define-frame-preference to something better and the have the current definition call the new one with a "warning this function is deprecated please use blah" and after a few versions we'll remove define-frame-preference.  As far as a name: "add-window-placement-rule" sounds like it captures the function better, then we could call your macro "define-window-placement-rules" which are separate enough for me.

scottjad:
I prefer existing APIs like this not change unless there's a really good reason, which I haven't heard. The reasons mentioned have been quite minor, IMO. Yes the name is not good, but it has the advantage of being known to existing users and already in their configs. I think deprecation warnings like this might be an issue in stumpwm/sbcl since it can output a ton of text (style warnings), and a long-time user probably doesn't look at it once their config is working.

Having duplicate individual rules is not the issue afaict. The first matching rule is all that's used, so if there are old rules in the list identical to the new ones, I don't think that will ever impact behavior.

Evaluating more than one form for the same group, and wanting only the latest to be active, is the issue, I think. Or in other words, the issue is having old rules that one no longer wants around because one is interactively developing their rules. In that case I think it's OK to require users to manually clear-window-placement-rules while developing, until they get a set of rules they like. It's not ideal, but better than requiring all existing users to change their configs, IMO.

I prefer the status quo, but if a change had to be made I prefer adding redefine-frame-preferences to the latest proposal by @dbjergaard, but that's because I'm extremely conservative with programs I use everyday and no longer change much, and I hate them breaking/changing names of things for little benefit to me the long-time user. I realize there's a trade-off and others will prefer the cleaner/better code improvement. Just wanted to share my opinion/preference.

dbjergaard:
@scottjad I agree with your sentiments completely.  When I started maintaining stumpwm there were a few cases where there were deprecated functions.  The warning I propose should happen as an X11 message, not as console output.  I never look at the console output unless I'm debugging something so its not fair to expect regular users to check this for deprecation warnings.

@lokedhs Thanks so much for the code contribution! Right now there are some unresolved issues with merging this as is.  Can you address the points raised above? If not I won't be able to merge this.  Please don't let this discourage future contributions though.

lokedhs:
@dbjergaard I think the only thing that could possibly be done is to rename the macro, but that wouldn't do much to improve the actual problems (unfortunate names of existing functions, etc). With that in mind, it's probably best not to do anything at this point.

PR #324, opened 2017-02-05
Clean up current window tracking for groups

jorams:
The first commit fixes the issue that a group's current window would not actually be a window in that group. The issue has always been present in the floating group, but was recently also introduced in the tiling group when it gained the ability to contain floating windows.

The second commit allows focusing behavior to properly be managed by the group the window is in.

It introduces one possible change in behavior: The floating group's GROUP-FOCUS-WINDOW method now always raises the window. I am not sure this actually changes anything user-visible.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -134,7 +134,7 @@
          (declare (ignore group))
          (change-class window 'float-window)
          (float-window-align window)
-         (focus-window window)))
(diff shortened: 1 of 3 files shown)


asalminen:
Thanks. Merging this fixed gnext-with-window for me.

PR #328, opened 2017-02-27
Add support for XRandR head queries if Xinerama is not available

rayslava:


head.lisp
```diff
@@ -56,6 +68,18 @@
                  :test #'frames-overlap-p)
                 do (setf (head-number h) i)
                 collect h)))
+      (and (xlib:query-extension *display* "RANDR")
+             (loop
+                for i = 0 then (1+ i)
+                for h in
+                (delete-duplicates
```
PuercoPop: Consider factoring this (delete-duplicates ...) into (list-heads-xrandr) (or something similar

head.lisp
```diff
@@ -56,6 +68,18 @@
                  :test #'frames-overlap-p)
                 do (setf (head-number h) i)
                 collect h)))
+      (and (xlib:query-extension *display* "RANDR")
+             (loop
+                for i = 0 then (1+ i)
```
PuercoPop: I know the previous loop is written this way but it is simpler to write this clause as `for i from 0`

head.lisp
```diff
@@ -56,6 +68,18 @@
...
+                for h in
+                (delete-duplicates
+                 (loop for i in (split-string (run-shell-command "xrandr -q" t))
```
PuercoPop: name the variable line instead of i

head.lisp
```diff
@@ -40,6 +40,18 @@
                                 (parse-error ()
                                   nil))))
 
+(defun parse-xrandr-head (line)
+  (let ((number 0))
+  (ppcre:register-groups-bind (('parse-integer width height x y))
```
PuercoPop: Is quoting parse-integer necessary?

PuercoPop:
Hi, Thanks for submitting the PR. 

Although I don't have multiple monitors to test it the code seems similar enough to the xinerama version that I'm pretty sure it works. 

Beside the comments left in the code (which are mostly due to taking as a template the code dealing with xinerama), there are some indentation issues. The form following the let form in parse-xrandr-head should be indented, the clause following the when clause in loop should be indented. Just doing X-h a, <tab> in slime should be enough. If you are not using slime don't worry about it, we can fix it ourselves before merging.

rayslava:
I thought that this loop construction was required. If it's not, I'd better do all the work in functional style without all these loop variables at all.

dbjergaard:
@rayslava can you replace the xinerama code completely with xrandr?

PuercoPop:
@rayslava I don't think that the issue is loop vs map but to break the code in smaller functions.

dbjergaard:
@rayslava I would prefer that this replace xinerama rather than fall back to xrandr.  Is it possible to address this? If not, I can merge it as is. My worry is that it will never be properly addressed if it gets merged now.

rayslava:
oh, I'm sorry, missed your comment. 
Okay, I'll replace the xinerama.

dbjergaard:
is this ready to go? I don't like PRs to sit in the queue so long

PuercoPop:
@dbjergaard We have to extract the regexp for xrandr from this PR (if we don't want to the CLX route).

PuercoPop:
Closing this PR as it appears that the XRandr extension implements the xinerama protocol so the using the xinerama extension covers both cases.  

 ref: https://github.com/stumpwm/stumpwm/pull/348#issuecomment-302901250

@rayslava Although the code didn't make it to master this time, it helped us clean up the code. Thanks you for taking the time to sumbit the PR!

PR #329, opened 2017-02-28
Simplify 'focus-last-frame'

alezost:
Currently `focus-last-frame` (used by `fother` command) calls
``(tile-group-last-frame group)`` 3 times!  I think it's not
good.


alezost:
Wow, that was very fast, thanks!

PR #335, opened 2017-03-13
Only try to handle events with XLIB:WINDOW windows

### LINKED ISSUE #334: Crash when switching groups
When switching groups, I get an error that a value `#<XLIB:DRAWABLE …>` is not of type `XLIB:WINDOW` in `XLIB:WINDOW-EQUAL` during a `HANDLE=EVENT`.

Some investigation shows that the `:window` slot is sometimes a drawable instead of a window. The simplest solution would be to simply ignore events in drawables.

zvbuhl:
For some reason, XLIB:PROCESS-EVENT sometimes sends events for
drawables as well as for windows; this causes crashes later on.

Fixes #334.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -597,7 +597,7 @@ they should be windows. So use this function to make a window out of XOBJECT."
   (let ((eventfn (gethash event-key *event-fn-table*))
         (win (getf event-slots :window))
         (*current-event-time* (getf event-slots :time)))
-    (when eventfn
+    (when (and eventfn (typep win 'xlib:window))
       (handler-case
           (progn
             ;; This is not the stumpwm top level, but if the restart
--- a/events.lisp
+++ b/events.lisp
@@ -585,11 +585,11 @@ the window in it's frame."
   ;; Pass click to client
   (xlib:allow-events *display* :replay-pointer time))
 
-(defun make-xlib-window (xobject)
+(defun make-xlib-window (drawable)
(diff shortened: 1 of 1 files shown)


PuercoPop:
Hi!
does this happen on SBCL? IIRC I removed code to handle this edge cause because the comment said it was CLISP specific. (I'm looking for the commit in question)

PuercoPop:
This is the commit in question e8ae004597bd5fc9fc41113a559246562831f2a2 
Could you confirm it is the same issue or that reverting that commit fixes the issue for you?

Btw, do you have a way to reproduce so we could file this issue upstream to CLX in addition to including the workaround?

jorams:
The relevant commit is actually 29eeb1fdf28af220937984970f9bf75b86b55781, and it contains links to more motivation. (See #318 and #321)

(Something seems to have gone wrong by merging both #318 and #321, because there's actually a stray definition of `make-xlib-window` in `events.lisp`.)

PuercoPop:
Thanks @jorams! You are right, that was the commit I was looking for

zvbuhl:
Yup, I'm seeing it happen in SBCL — very consistently.

zvbuhl:
Looking at 29eeb1f I think the older way was a much better approach than my own lazy just-drop-the-event fix. I'll update my PR.

zvbuhl:
The new commit should use the better way of handling things.

Also, it correctly short-circuits if the event window slot is neither a window nor a drawable.

Issue #325, opened 2017-02-19
"C-t !" displays /bin/sh even when *shell-program* is non-nil


sydii:
Currently /bin/sh is hardcoded into the definition of `run-shell-command` which causes it to be displayed instead of the user's preferred shell when `*shell-program*` is non-nil. What should occur is upon setting `*shell-program*`, the user-set shell should be displayed on screen when `C-t !` is typed.

The code currently in user.lisp, line 157 to 164:
```
(defcommand run-shell-command (cmd &optional collect-output-p) ((:shell "/bin/sh -c "))
  "Run the specified shell command. If @var{collect-output-p} is @code{T}
then run the command synchonously and collect the output. Be
careful. If the shell command doesn't return, it will hang StumpWM. In
such a case, kill the shell command to resume StumpWM."
  (if collect-output-p
      (run-prog-collect-output *shell-program* "-c" cmd)
      (run-prog *shell-program* :args (list "-c" cmd) :wait nil)))
```

dbjergaard:
Fixed with #336

PR #336, opened 2017-03-17
Let shell to be run be set by user

tpine:
Change define-stumpwm-type :shell to allow the setting of the shell which the user wishes to run.
This is done by the addition of the variable *shell-to-use*.
When *shell-to-use* is nil the default "/bin/sh -c" is used otherwise the value of *shell-to-use* is used. 

Fixes issue #325

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -451,14 +451,11 @@ then describes the symbol."
     (or (choose-frame-by-number (current-group))
         (throw 'error :abort))))
 
-(defvar *shell-to-use* nil)
-
 (define-stumpwm-type :shell (input prompt)
-  (if (eq *shell-to-use* nil)
-      (or (argument-pop-rest input)
-          (completing-read (current-screen) prompt 'complete-program))
-      (or (argument-pop-rest input)
-          (completing-read (current-screen) *shell-to-use* 'complete-program))))
+  (declare (ignore prompt))
+  (let ((prompt (format nil "~A -c " *shell-program*)))
+    (or (argument-pop-rest input)
+        (completing-read (current-screen) prompt 'complete-program))))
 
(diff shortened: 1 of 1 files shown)


PuercoPop:
Hi! Thanks for following on the issue with a PR.

There are a couple of problems with the proposed solution. First `*shell-to-use*` and `*shell-program*` point to the same concept but the first only affects the prompt and the second only the program. Also you don't use the prompt variable in both branches. Using `(declare (ignorable prompt))` would give warning to future users about this.

The proper solution would be for interactive args to able able to accept a variable (or at least a symbol?) as a prompt argument use its bound value as the prompt argument. However this would, unless I'm mistaken, require a not minor change to how commands are stored internally.

A better stop-gap measure would be
```lisp
(define-stumpwm-type :shell (input prompt)
  (declare (ignore prompt))
  (let ((prompt (format nil "~A -c" *shell-program*)))
    (or (argument-pop-rest input)
        (completing-read (current-screen) prompt 'complete-program))))
```

@dbjergaard what do you think?

tpine:
I do prefer your method. My option is quite silly and was more a solution to a simple bug I saw to get started with the stumpwm source and common lisp.
Is the (declare (ignore prompt)) required? The original version doesn't use it and only uses prompt for one of the branches.

PuercoPop:
It is not required. At most it will avoid the compiler emitting an unused variable warning. Btw don't sell yourself short, as 'silly' as your solution was it a step in the right direction. I was stuck trying it to address the issue at the define-command level instead of define-type.

PuercoPop:
Looks good to me 👍 (Although I prefer if you rebase the branch to only include the last commit, but don't worry about it)

PR #338, opened 2017-03-19
Revert "Remove un-needed SBCL_HOME reference"

dbjergaard:
Reverts stumpwm/stumpwm#333


jorams:
I think we should handle this nicer. I now have to manually remove these lines from `make-image.lisp`, because on my system, like in #333, `SBCL_HOME` is not set.

PR #344, opened 2017-04-04
Refactor make-screen-heads

PuercoPop:
This is just a refactor or the current head implemenatation to make it more readable. It is a first stop towards #328, which seems to have gone cold

### DIFF
--- a/head.lisp
+++ b/head.lisp
@@ -40,22 +40,29 @@
                                 (parse-error ()
                                   nil))))
 
+(defun list-heads ()
+  "Return a list of HEAD structures without duplicates"
+  (delete-duplicates
+   (loop for line in (split-string (run-shell-command "xdpyinfo -ext XINERAMA" t))
+         for head = (parse-xinerama-head line)
+         when head
+           collect head)
+   :test #'frames-overlap-p))
+
+(defun number-heads (heads)
+  "Number each head according to its position in the list."
+  (let ((head-index -1))
+    (flet ((set-head-number (head)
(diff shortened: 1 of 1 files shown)


head.lisp
```diff
@@ -40,22 +40,29 @@
...
+   :test #'frames-overlap-p))
+
+(defun number-heads (heads)
+  "Number each head according to its position in the list."
+  (let ((head-index -1))
+    (flet ((set-head-number (head)
+             (setf (head-number head)
+                   (incf head-index))))
+      (mapc #'set-head-number heads))))
+
```
Harleqin: (loop for head in heads
          for i upfrom 0
          do (setf (head-number head) i))

PuercoPop: @Harleqin Yes, that is the code I replaced with a mapc, because it has a higher information density. Ideally I'd use misc-extensions's rflet so the flet would be at the end instead of cluttering the relevant code. Do you think the loop version is more readable in this case?

Harleqin: Yes, absolutely, it also feels more declarative to me.

PuercoPop: @Harleqin I'm on the fence on this own. @jorams @dbjergaard Any preference?

dbjergaard: Honestly the flet+incf+mapc seems like scratching your nose behind your head to keep it functional.  The operation is inherently declarative and its easier to mentally parse the loop version.

PuercoPop: K, I'll go back to using loop

PR #348, opened 2017-04-17
Use CLX's Xinerama extesion to query screen heads

PuercoPop:
I've done some light testing with Xephyr, `Xephyr -screen 640x480 -screen 640x480 +xinerama :1 &`

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,11 +112,6 @@ Hopefully that will put you in X running stumpwm! See [StartUp on the
 wiki](https://github.com/sabetts/stumpwm/wiki/StartUp) for more
 examples.
 
-## Requirements for multiple monitor setups
-
-For stumpwm to work as intended with multiple monitors setups the 
-`xdpyinfo` utility is needed.
-
 # Contributing
 
 Pull requests are always welcome! Here are some guidelines to ensure
--- a/head.lisp
+++ b/head.lisp
@@ -29,44 +29,25 @@
 (defun head-by-number (screen n)
   (find n (screen-heads screen) :key 'head-number))
 
-(defun parse-xinerama-head (line)
-  (ppcre:register-groups-bind (('parse-integer number width height x y))
-                              ("^ +head #([0-9]+): ([0-9]+)x([0-9]+) @ ([0-9]+),([0-9]+)" line :sharedp t)
-                              (handler-case
-                                  (make-head :number number
-                                             :x x :y y
-                                             :width width
-                                             :height height)
-                                (parse-error ()
-                                  nil))))
-
-(defun list-heads ()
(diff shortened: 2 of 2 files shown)


PuercoPop:
Michael Raskin noted that this PR did away with the case where xinerama wasn't present, I'll amend it

PuercoPop:
There are still some cases to debug with checking for the presence of Xinerama

jorams:
@PuercoPop What kind of cases are those? My testing suggests this works and can be merged.

PuercoPop:
I don't have a multi-monitor setup so I used Xephyr for testing.

One problem is that it seems to detect xinerama as active even when it isn't. cc @38a938c2 

For example using the following Xephyr configuration
```
$ Xephyr -screen 640x480 -xinerama :1 &
```
The result of `(xlib/xinerama:xinerama-is-active *display*)` is true.

jorams:
That is probably related to this [bug report](https://bugs.freedesktop.org/show_bug.cgi?id=11505). In short, the Xinerama extension can be disabled, but the protocol support is still implemented otherwise. The RandR extension, for example, implements the Xinerama protocol. ([This](https://bugs.freedesktop.org/show_bug.cgi?id=11118) bug report reports the same thing explicitly about Xephyr).

I'm not sure if it's even possible, these days, to run an X server that doesn't report Xinerama support.

jorams:
Ok, I did some more testing. It *is* possible to disable Xinerama in Xephyr, by adding `-extension RANDR` to the line you posted above.

Then starting StumpWM results in the following exception:

```
debugger invoked on a XLIB::ABSENT-EXTENSION in thread
#<THREAD "main thread" RUNNING {10031C6753}>:
  Extension XINERAMA isn't defined for display #<XLIB:DISPLAY :1 (The X.Org Foundation R11903000)>

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [QUIT-STUMPWM] Quit Stumpwm

(XLIB::X-ERROR XLIB::ABSENT-EXTENSION :NAME :XINERAMA :DISPLAY #<XLIB:DISPLAY :1 (The X.Org Foundation R11903000)>)
   source: (APPLY #'ERROR CONDITION KEYARGS)
0] 
```

jorams:
It's quite easy to fix: Just make sure `(xlib:query-extension *display* "XINERAMA")` doesn't return `NIL`, before calling `xinerama-is-active`.

jorams:
This should now work correctly. Before merging, it's probably a good idea to remove the `xdpyinfo` mention from the README, because it's no longer necessary.

jorams:
Alright, I think this is ready to merge.

PR #351, opened 2017-05-15
Remove conditionals for non sbcl common lisp implementations

tpine:
Work on removing all non sbcl conditionals from stumpwm and cleaning up the remaining code where possible if the conditional only specified a few lines.

Currently this is a lot of small commits but they can be rebased into more logical groupings when they occur before merge.


PuercoPop:
@tpine Thanks for the taking the time to submit this PR. There are some commits which I'd like to merge, like the directory-no-deref one. There are a couple of issues with other commits in this PR though, for example you are replacing all calls to getenv to the sb-posix equivalent. However getenv is an exported symbol of the StumpWM package, so the approach you are taking would break API compatibility. Maybe it would be best to import the symbol from the sb-posix package? (I'd have to check how that plays with importing the same symbol in a different package that is also using stumpwm ej
`(defpackge #:foo (:use #:sb-posix #:stumpwm))`

tpine:
In regards to sb-posix:getenv I was thinking of doing something along those lines but wouldn't that approach break the api still?

I was also wondering while I did this if we could replace execv in some way since the same result as the execv syscall can be done with sb-ext:run-program?

PuercoPop:
> In this case would it really be worth not breaking the api simply to redefine getenv to sb-posix:getenv though?

I'm leaning towards yes.

1. When customizing StumpWM we expected  to evaluate their code inside the `stumpwm-user` pacakge. Accessing environment variables is a common task so it seems logically to provide the functionality. Or at the very least not remove it if we have been providing it. If editing from Emac/SLIME I can see the extra 9 characters not being an issue. But if the user wants to eval some code through StumpWM's prompt?

2.  Also the first place a newcomer will look for how to access environment variables would be in the StumpWM manual, not in SBCL. We have getenv listed in the manual https://stumpwm.github.io/git/stumpwm-git_10.html#index-getenv

3. The API exported by StumpWM is different that getenv/setenv. We use a more lispy approach,`(setf getenv ..)`

Note we would not be redefining, just re-exporting.

tpine:
I've reverted the getenv changes in the last commit.

tpine:
All conditionals for other lisp implementations have now been removed.
The only conditionals left are:
- clx-ext-test
- asdf
- linux

dbjergaard:
I'm happy as long as getenv stays the same. Don't break the API unless you have a *really* good reason.

dbjergaard:
@PuercoPop @jorams thanks for all the help you guys have been doing to keep the PRs clean and merged in a timely manner! @PuercoPop please merge this one when you are happy with it.

PuercoPop:
@dbjergaard It looks good to me, thought a little bit uneasy about merging so many changes in one go.

@tpine Thanks for taking the time to do this!

PR #352, opened 2017-05-15
Remove non sbcl lisp implementation references from README

tpine:
These are left behind from before the usage of non SBCL lisp implementations was removed so remove them.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -61,3 +61,4 @@ Joram Schrijver     i at joram io
 Yu Changyuan        reivzy at gmail com
 Edward Trumbo       trumboe at comcast net
 Javier Olaechea     pirata at gmail com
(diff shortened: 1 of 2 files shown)


PuercoPop:
Some minor issues, StumpWM doesn't seem to depend on `cl-xembed` any longer. Also the name of the system is [`clx-xembed`](http://quickdocs.org/clx-xembed/).

Thanks for the recent contributions btw!

tpine:
I checked by seeing if I could use `(in-package :xembed)` on my system and the package didn't exist so I've removed it from the README.

jorams:
I'm pretty sure `xembed` is still there because Stumptray requires it. Stumptray is now an entirely separate system, so it is indeed unneeded.

PR #353, opened 2017-05-18
Add more split functions

caioaao:
So we can split a frame in n rows/columns of equal sizes.

I also reorganized the next command defined to fit in 80 columns, but I don't know if I'm breaking some style rule.

### DIFF
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -931,7 +931,7 @@ windows used to draw the numbers in. The caller must destroy them."
   (split-frame-in-dir (current-group) :row (read-from-string ratio)))
 
 (defun split-frame-eql-parts* (group dir amt)
-  (when (not (eq amt 1))
+  (when (> amt 1)
     (when-let ((new-frame (split-frame group dir (/ (- amt 1) amt))))
       (cons new-frame (split-frame-eql-parts* group dir (- amt 1))))))
 
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -930,9 +930,46 @@ windows used to draw the numbers in. The caller must destroy them."
 "Split the current frame into 2 frames, one on top of the other."
   (split-frame-in-dir (current-group) :row (read-from-string ratio)))
 
-(defcommand (remove-split tile-group) (&optional (group (current-group)) (frame (tile-group-current-frame group))) ()
-"Remove the specified frame in the specified group (defaults to current
-group, current frame). Windows in the frame are migrated to the frame taking up its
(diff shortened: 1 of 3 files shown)


tile-group.lisp
```diff
@@ -930,9 +930,46 @@ windows used to draw the numbers in. The caller must destroy them."
 "Split the current frame into 2 frames, one on top of the other."
   (split-frame-in-dir (current-group) :row (read-from-string ratio)))
 
-(defcommand (remove-split tile-group) (&optional (group (current-group)) (frame (tile-group-current-frame group))) ()
-"Remove the specified frame in the specified group (defaults to current
-group, current frame). Windows in the frame are migrated to the frame taking up its
+(defun split-frame-eql-parts* (group dir amt)
+  (when (not (eq amt 1))
```
Harleqin: Compare numbers with `=`. In this case, use `/=` instead of `(not (= ...))`.  I'd actually think about using `>`.

tile-group.lisp
```diff
@@ -930,9 +930,46 @@ windows used to draw the numbers in. The caller must destroy them."
...
+    (when-let ((new-frame (split-frame group dir (/ (- amt 1) amt))))
+      (cons new-frame (split-frame-eql-parts* group dir (- amt 1))))))
+
+(defun split-frame-eql-parts (group dir amt)
+  "Splits frame in equal parts defined by amt."
+  (assert (> amt 1))
+  (let ((f (tile-group-current-frame group))
+        (new-frame-numbers (split-frame-eql-parts* group dir amt)))
+    (if (= (list-length new-frame-numbers) (- amt 1))
+        (progn
```
PuercoPop: When has an implicit progn, remove this progn form

tile-group.lisp
```diff
@@ -930,9 +930,46 @@ windows used to draw the numbers in. The caller must destroy them."
...
+          (when (frame-window f)
+            (update-decoration (frame-window f)))
+          (show-frame-indicator group))
+        (progn
```
PuercoPop: Is there a reason why the message cannot go inside the let so we can remove the progn here as well?

tile-group.lisp
caioaao: `show-frame...` is outside `when`, that's why I needed the `progn`. not sure why it's outside as I just copied what's done here: https://github.com/stumpwm/stumpwm/blob/master/tile-group.lisp#L919

PuercoPop: You are right. My bad

dbjergaard:
I'm happy with this as is.

PR #354, opened 2017-05-21
Allows keys to be used when loading modules

tpine:
When interacting with systems support is normally allowed to specify
the system by using a key instead of a string. Add this support to load-module.

### DIFF
--- a/module.lisp
+++ b/module.lisp
@@ -117,6 +117,7 @@ an asdf system, and if so add it to the central registry"
 
 (defcommand load-module (name) ((:module "Load Module: "))
   "Loads the contributed module with the given NAME."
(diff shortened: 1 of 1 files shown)


module.lisp
```diff
@@ -117,6 +117,7 @@ an asdf system, and if so add it to the central registry"
 
 (defcommand load-module (name) ((:module "Load Module: "))
   "Loads the contributed module with the given NAME."
+  (setf name (string-downcase name))
   (let ((module (find-module name)))
```
jorams: Since `name` is only used here, you can just put `(string-downcase name)` in its place. No need to `setf`.

PR #358, opened 2017-06-01
Manual fixes

impaktor:
Hi!

After many years as a stump user, I thought I should probably actually read the manual. Eight pages in I find I've underlined a bunch of stuff that's wrong, or could be improved. I thought I'd share them here, you can cherry pick the stuff you want if not all is good.

Let me know if I should keep going as I read pages 10-70 as well.


PuercoPop:
Thanks for taking the time to submit this! <3

PR #359, opened 2017-06-01
Fix typos

ecraven:
Fix a few typos and disable line breaks for command lambda lists (right now if they are too long, they are broken across a few lines and texi renders them incorrectly).
Also add all missing commands to the documentation, and fix a few functions to actually be documented as functions, not commands.


PuercoPop:
@ecraven Thanks for taking the time to improve the manual!

PR #361, opened 2017-06-02
Fix doc generation failure

zvbuhl:
This fixes a typo in the doc source, and adds a somewhat-more-useful error message.


PuercoPop:
@zvbuhl Thanks for catching this. Should have been more thorough. I've merged this PR partially. I went a different way handling the case when the symbol is not found.

PR #362, opened 2017-06-03
More documentation fixes 2

impaktor:
Some more typo fixes. (Note to self: I'm now on page 33 of 70 on pdf-version of manual).

### DIFF
--- a/bindings.lisp
+++ b/bindings.lisp
@@ -237,7 +237,7 @@ from most specific groups to most general groups.")
   (kbd "w") "where-is")
 
 (defcommand command-mode () ()
-"Command mode allows you to type ratpoison commands without needing the
+"Command mode allows you to type StumpWM commands without needing the
 @key{C-t} prefix. Keys not bound in StumpWM will still get sent to the
 current window. To exit command mode, type @key{C-g}."
   (run-hook *command-mode-start-hook*)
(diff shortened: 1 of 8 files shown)


PuercoPop:
@dbjergaard @jorams How do y'all feel about ba87203d7d82ced07a3639143813ed30b5520136?
I prefer using uppercase for all symbols that refer to functions/variables/etc, but I'm aware it is an archaism at this point.

@impaktor @ecraven Thanks for proof-reading the manual!

impaktor:
I have  no opinion one way or another, but my impression is that lowercase is by far the most frequent in the manual, thus I opted for the path of least resistance.

In general, I don't think there is a standard in the manual for how to typeset key words, functions, and variable names?

PuercoPop:
@impaktor Don't worry, I would have chosen the same approach as you in your place.

The [convention is more a common lispone](https://google.github.io/styleguide/lispguide.xml?showone=Document_everything#Document_everything), but is becoming increasingly uncommon. And we embed texinfo in our docstrings so we are already unconventional.

impaktor:
I've found some more typos, some fitted nicely in the other commits, so squashed those changes into there (like capitalize new sentences), and I also removed the controversial "is it `NIL` or `nil`" as further in, e.g. chapter 12, most variables are uppercase, so I myself can't be consistent, thus I removed it.

PuercoPop:
@impaktor Thanks 🎉

PR #363, opened 2017-06-07
Allow define-stumpwm-type :function to throw error

tpine:
define-stumpwm-type :function uses symbol-function to check if a
symbol is a function except symbol-function throws a error if the
symbol isn't a function instead of nil. Common Lisp the Language
provides a function to address this issue. Implement the logic to
allow our error handling to work.

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -323,7 +323,7 @@ then describes the symbol."
 (define-stumpwm-type :function (input prompt)
   (multiple-value-bind (sym pkg var)
       (lookup-symbol (argument-pop-or-read input prompt))
-    (if (symbol-function sym)
+    (if (fboundp sym)
         (symbol-function sym)
         (throw 'error (format nil "the symbol ~a::~a has no function."
                               (package-name pkg) var)))))


PuercoPop:
@tpine Is there a reason why `FBOUNDP` is not enough (besides the obvious semantic reason that macros and special operators are not functions)? Where do we use the function type?

tpine:
There is no reason except this was the solution tk this problem I found in Common Lisp the Language when I was checking what symbol-function did. We can just reduce the check to that if you like.

I have no idea where we use the function type myself Im just cleaning up the warnings which come up with building stumpwm.

jorams:
The `:function` type appears to only be used for the command `describe-function`. I also checked the stumpwm-contrib sources. That command mirrors the `describe-function` in Emacs, which accepts every kind of function, so there just checking `fboundp` seems to make most sense. Looking at it more generally, I would expect `:function` to result in a function I can actually call.

Because it's the only use of `:function`, I'm inclined to say we should just only check `fboundp`despite it being slightly unintuitive.

PuercoPop:
ok then lets use `fbound` then. @tpine would you update the PR to `(if (fboundp ..) ...)`

tpine:
Fixed. Sorry for the late reply.

PuercoPop:
Merged in 16b2810bf04f22ab062cc0bd512f2b4dd914d2b4 with some minor changes

PR #364, opened 2017-06-07
Remove useless (declare (ignore group))

tpine:
group is used in function so remove declaration.


PuercoPop:
Merged on 1d8c6140eb1d5e0af037c770888890dd47e3df76

Thanks for spotting it :tada:

PR #367, opened 2017-06-10
Remove clisp window urgency flags workaround

tpine:
Clisp bundles its version of clx which changed to using a list to
represent a windows flags instead of a integer.

Remove the workaround so we cold use both versions of clx as we now
only support one clx impementation.


jorams:
Thanks a lot! Small improvements improving clarity and removing unnecessary edge-cases are always good.

PR #368, opened 2017-06-10
Find current head from pointer in floating group

jorams:
In the tiling group the pointer is usually ignored, but in the floating
group it is essential. Thus it makes sense to use its position to
determine the current head.

I think this makes sense, do you (@PuercoPop , @dbjergaard) agree?

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -150,6 +150,13 @@
 
 (defmethod group-suspend ((group float-group)))
 
+(defmethod group-current-head ((group float-group))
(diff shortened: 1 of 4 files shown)


floating-group.lisp
```diff
@@ -150,6 +150,13 @@
 
 (defmethod group-suspend ((group float-group)))
 
+(defmethod group-current-head ((group float-group))
+  (if (group-current-window group)
```
PuercoPop: We can use `alexandria:if-let` here.

PR #370, opened 2017-07-02
More documentation fixes, part 3

impaktor:
I've finished reading the manual, and thus the trilogy has come to its grand finale, with this last PR.

Have a look, and comment, and I'll rebase/modify as needed.

### DIFF
--- a/color.lisp
+++ b/color.lisp
@@ -42,7 +42,7 @@
 ;;    - :fg and :bg take a color as an argument, which can either be a numeric
 ;;      index into the color map or a hexadecimal color in the form of "#fff"
 ;;      or "#ffffff".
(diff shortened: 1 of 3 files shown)


color.lisp
```diff
@@ -42,7 +42,7 @@
 ;;    - :fg and :bg take a color as an argument, which can either be a numeric
 ;;      index into the color map or a hexadecimal color in the form of "#fff"
 ;;      or "#ffffff".
-;;    - :reverse and :bright take either t or nil as an argument. T enables
```
impaktor: I really dislike starting sentences with variable names or numbers, e.g. years or dates. It's like starting a dance on the wrong foot or beat. Not that I dance, but as an analogy.

PR #377, opened 2017-07-17
Added functions IN-MAIN-THREAD-P and PUSH-EVENT

lokedhs:
The function `CALL-IN-MAIN-THREAD` checks whether the current thread is the main thread, in which case the function is called immediately. If not, an event is scheduled which will be called at a later time.

There are cases when one wants to schedule a function to be called from the event loop even if the executing thread is the event thread. A specific case for this is when calling the function `FLOAT-WINDOW` from a window create hook. `FLOAT-WINDOW` assumes that the window has already been placed, and if it's called directly from the hook, StumpWM will crash. Scheduling the call as an event using the new function `PUSH-EVENT` fixes this.

### DIFF
--- a/package.lisp
+++ b/package.lisp
@@ -20,9 +20,7 @@
 (defpackage :stumpwm
   (:use :cl
         #:alexandria)
-  (:shadow #:yes-or-no-p #:y-or-n-p)
-  (:export
-   #:call-in-main-thread))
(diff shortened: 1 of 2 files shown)


PuercoPop:
@lokedhs Thanks for the the making the code more readable

PR #380, opened 2017-07-22
Fix for incorrect selection targets property value

kriyative:
This commit ensures that any selection owned by stumpwm may be correctly requested by other X clients that use the `:targets` property request.

### DIFF
--- a/selection.lisp
+++ b/selection.lisp
@@ -50,7 +50,14 @@
   (case target
     ;; they're requesting what targets are available
     (:targets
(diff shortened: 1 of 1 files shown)


kriyative:
@dbjergaard - thanks for accepting this PR. I have another couple of PRs which I'll add in a little bit - that build upon this and adds more functionality for managing selections in StumpWM - do let me know if you have any feedback or suggestions. Thanks.

PR #381, opened 2017-07-24
Feature selection notify hook

kriyative:
Adds a `:selection-notify` event handler and `*selection-notify-hook*` to allow StumpWM to get clipboard (and other) selection from other X clients. Selection is saved in `*x-selection*`

The hook can be used to build useful utilities such as a `clipboard-history` module - an early draft (which depends on this hook) can be previewed [here](https://github.com/kriyative/stumpwm-contrib/commit/c25f6dfa887dd7f7ad1c18148592d76a6102d086).

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -383,6 +383,21 @@ converted to an atom is removed."
 (define-stump-event-handler :selection-clear (selection)
   (setf (getf *x-selection* selection) nil))
 
+(define-stump-event-handler :selection-notify (window property selection)
+  (dformat 2 "selection-notify: ~s ~s ~s~%" window property selection)
+  (when property
+    (let* ((selection (or selection :primary))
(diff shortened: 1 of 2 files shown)


PuercoPop:
Hey pretty cool PR. Just left a couple nitpicks. Regarding the clipboard history, @dbjergaard how do you feel about including it in StumpWM instead?

dbjergaard:
this is another example of a really great module that has been recently written.  It falls clearly under the scope of a module, but it provides functionality that is very useful to the window manager. I'm torn.  My long term vision for these situations is to somehow track how popular they are and when they reach a certain point move them from contrib to stumpwm proper.  For now, I think having the module added to stumpwm-contrib is the best approach.

kriyative:
@dbjergaard - thanks.

PR #394, opened 2017-09-13
Add a selection hook in select-from-menu.

ralt:
This lets the caller do something every time an item is selected.

A use case for this is is a "visual-pull-from-windowlist", which pulls
the selected window every time you select it. Here is an example
implementation that is currently used:

```common-lisp
(defcommand (visual-pull-from-windowlist tile-group) () ()
  (flet ((pull-selected-window (menu)
           (let ((selected-window (nth (menu-state-selected menu)
                                       (menu-state-table menu))))
             (pull-window (second selected-window)))))
    (let ((original-window (current-window))
          (pulled-window (select-window-from-menu
                          (group-windows (current-group))
                          *window-format*
                          nil
                          *window-menu-filter*
                          #'pull-selected-window)))
      (pull-window (if pulled-window pulled-window original-window)))))
```

### DIFF
--- a/menu.lisp
+++ b/menu.lisp
@@ -243,6 +243,7 @@ Returns the selected element in TABLE or nil if aborted. "
                     (when prompt-line
                       (push prompt-line strings)
                       (incf highlight))
+                    (run-hook-with-args *menu-selection-hook* menu)
                     (echo-string-list screen strings highlight))
                   (multiple-value-bind (action key-seq) (read-from-keymap keymap)
                     (if action
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -59,6 +59,7 @@
           *pre-command-hook*
           *post-command-hook*
           *selection-notify-hook*
+          *menu-selection-hook*
           *display*
           *shell-program*
           *maxsize-border-width*
@@ -327,6 +328,10 @@ the command as a symbol.")
 (defvar *selection-notify-hook* '()
   "Called after a :selection-notify event is processed. It is called
 with 1 argument: the selection as a string.")
+
+(defvar *menu-selection-hook* '()
+  "Called after an item is selected in the windows menu. It is called
+with 1 argument: the menu.")
 
 ;; Data types and globals used by stumpwm
 
--- a/stumpwm.texi.in
+++ b/stumpwm.texi.in
@@ -2152,6 +2152,7 @@ $$$ *urgent-window-hook*
 $$$ *event-processing-hook*
 $$$ *pre-command-hook*
 $$$ *post-command-hook*
+$$$ *menu-selection-hook*
 
 @node Modules, Hacking, Hooks, Top
 @chapter Modules


ralt:
I'm currently using it and it's super nice, you get a preview of the window before hitting "Enter" to switch to it, you can easily cycle through all your windows to see how things are going, etc. Highly recommend it.

ralt:
Here is the function I'm now using which fixes the 2nd issue mentioned above: 

```common-lisp
(defcommand (visual-pull-from-windowlist tile-group) () ()
  (flet ((pull-selected-window (menu)
           (let ((selected-window (second
                                   (nth (menu-state-selected menu)
                                        (menu-state-table menu)))))
             (unless (some (lambda (frame)
                             (eq (frame-window (if (listp frame) (first frame) frame))
                                 selected-window))
                           (tile-group-frame-tree (current-group)))
               (pull-window selected-window)))))
    (let ((original-window (current-window))
          (pulled-window (select-window-from-menu
                          (group-windows (current-group))
                          *window-format*
                          nil
                          *window-menu-filter*
                          #'pull-selected-window)))
      (pull-window original-window)
      (pull-window (if pulled-window pulled-window original-window)))))
```

(Basically, ignore windows that appear on other frames.)

As for the 1st issue, I still have this screen flickering, but it's acceptable and barely noticeable. If it wasn't for this issue, I would totally open a PR with this though :)

dmh43:
cool! reminds me of some of the functionality in helm

PuercoPop:
Nice work @ralt 🎉 !

One thing to consider is, now that `select-from-menu` has 5 optional arguments, would it make sense to turn them into keyword arguments? (I'm leaning towards yes, don't think performance would be a concern here).

Also if you could add an entry describing the `selection-hook` to the manual? https://stumpwm.github.io/git/stumpwm-git_14.html#Hooks

ralt:
Hey,

I don't really care about optional vs keywords, it's a stylistic choice, and stumpwm seems to be mostly about optional, so I kept it that way.

As for the documentation, it makes me realize that I should probably make it a global variable and not an argument, like the other hooks? There can be only one menu at a time, right?

ralt:
Meanwhile, here is the function I'm finally using that fixes all the issues I previously had, plus other.

1. No double or triple pull-window now, which means no screen flickering.
2. The order of the window's list is properly maintained.
3. Split frames work fine.

```common-lisp
(defun pull-selected-window (menu)
  (let ((selected-window (second
                          (nth (menu-state-selected menu)
                               (menu-state-table menu)))))
    (unless (some (lambda (frame)
                    (eq (frame-window (if (listp frame) (first frame) frame))
                        selected-window))
                  (tile-group-frame-tree (current-group)))
      (pull-window selected-window))))

(defun restored-windows (pulled-window original-group-windows)
  (let ((new-group-windows (group-windows (current-group))))
    (if pulled-window
        (append (list pulled-window)
                (remove pulled-window
                        (mapcar (lambda (window)
                                  (find-if
                                   (lambda (w)
                                     (eq (window-id w) (window-id window)))
                                   new-group-windows))
                                original-group-windows)))
        original-group-windows)))

(defcommand (visual-pull-from-windowlist tile-group) () ()
  (let* ((original-window (current-window))
         (original-group-windows (group-windows (current-group)))
         (pulled-window (select-window-from-menu
                         (group-windows (current-group))
                         *window-format*
                         nil
                         *window-menu-filter*
                         #'pull-selected-window)))
    (setf (group-windows (current-group))
          (restored-windows pulled-window original-group-windows))
    (unless pulled-window
      (pull-window original-window))))
```

PuercoPop:
> I don't really care about optional vs keywords, it's a stylistic choice, and stumpwm seems to be mostly about optional, so I kept it that way.

It isn't a matter of style. You only have to specify the keywords you care about, if you want to specify an optional argument that happens to be at the last position, you have to specify all the arguments, even those you don't care about and be aware of their default values. When its one or two arguments the difference doesn't matter. When its 4+ arguments it starts to matter.

However, this doesn't matter because as you point out the hook should be a global variable so the lambda list remains unchanged.

ralt:
Hey,

I have updated the PR code to reflect the latest discussions.

Here is the new visual-pull-from-windowlist (a bit ugly given that we have to use a global variable to track if we want to trigger the hook or not):

```common-lisp
(defun pull-selected-window (menu)
  (let ((selected-window (second
                          (nth (menu-state-selected menu)
                               (menu-state-table menu)))))
    (unless (some (lambda (frame)
                    (eq (frame-window (if (listp frame) (first frame) frame))
                        selected-window))
                  (tile-group-frame-tree (current-group)))
      (pull-window selected-window))))

(defun restored-windows (pulled-window original-group-windows)
  (let ((new-group-windows (group-windows (current-group))))
    (if pulled-window
        (append (list pulled-window)
                (remove pulled-window
                        (mapcar (lambda (window)
                                  (find-if
                                   (lambda (w)
                                     (eq (window-id w) (window-id window)))
                                   new-group-windows))
                                original-group-windows)))
        original-group-windows)))

(defvar *is-in-visual-pull* nil)

(add-hook *menu-selection-hook*
          (lambda (menu)
            (when *is-in-visual-pull*
              (pull-selected-window menu))))

(defcommand (visual-pull-from-windowlist tile-group) () ()
  (setf *is-in-visual-pull* t)
  (let ((original-window (current-window))
        (original-group-windows (group-windows (current-group)))
        (pulled-window (select-window-from-menu (group-windows (current-group)) *window-format*)))
    (setf *is-in-visual-pull* nil)
    (setf (group-windows (current-group))
          (restored-windows pulled-window original-group-windows))
    (when (and (not pulled-window)
               (not (= (window-id original-window) (window-id (current-window)))))
      (pull-window original-window))))
```

PuercoPop:
Thanks for your contribution

PR #396, opened 2017-09-15
Add the visual-pull-from-windowlist command.

ralt:
Following on up #394, this new command implements a new way to cycle
through the list of open windows and select the appropriate one.

This version has no drawback that I'm aware of:

- If a window is on another frame, it won't be pulled
- Windows are only pulled when necessary
- Stack history of windows is properly kept in the group's list.

The real drawback of this solution is that it's using a hook that is
externalised, and it's the only one that is non-nil that I can
see. Should it use a separate, internal-only, hook, that select-menu
will also call? Or instead, add a new argument to `select-window-from-menu`?

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -8,6 +8,9 @@
   "Set this to T if you never want windows to resize based on incremental WM_HINTs,
 like xterm and emacs.")
 
+(defvar *is-in-visual-pull* nil
+  "Whether we are visually pulling a window or not. Used to keep track of state.")
+
 (defclass tile-window (window)
   ((frame   :initarg :frame   :accessor window-frame)
    (normal-size :initform nil :accessor window-normal-size)))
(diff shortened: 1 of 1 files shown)


tile-window.lisp
```diff
@@ -336,6 +339,47 @@ when selecting another window."
...
+                                original-group-windows)))
+        original-group-windows)))
+
+(add-hook *menu-selection-hook*
+          (lambda (menu)
+            (when *is-in-visual-pull*
+              (pull-selected-window menu))))
+
+(defcommand (visual-pull-from-windowlist tile-group) () ()
+  (setf *is-in-visual-pull* t)
```
PuercoPop: Wouldn't it be better to do
(let ((*is-in-visual-pull* t))
  (declare (special **is-in-visual-pull*))
  ...)

ralt: This would be a better idea, but we need to decide on what to do w.r.t. the external hook being non-NIL first

dbjergaard:
@ralt these two PRs are awesome. Once you fix the -p renaming, I'm happy to have this merged.

PuercoPop:
Merged in https://github.com/stumpwm/stumpwm/commit/8441235056fc29e6cbfadf1d9c2865197bf70496

ralt:
Hey,

Sorry for not getting back before. I was waiting on the decision w.r.t. the hook not being empty, which I'm less and less a fan of. Perhaps that should be refactored before the next release so that the external hook not being empty isn't part of the public interface.

PuercoPop:
@ralt I'm not fan of it either, but its been 2 months and I didn't came up with a better way. Perfect is the enemy of good.

PR #400, opened 2017-10-13
Add new head hook

caioaao:
So we can run stuff when a new monitor is plugged, for instance.


caioaao:
@PuercoPop It was an experiment for making the mode-line appear when I plug my monitor, but it seems it's actually a bug on the frame area calculation when a new head is created instead (I still have to investigate further to be sure), but I decided to open this PR because it looked useful and was one hook I was missing (even though it didn't serve my purposes atm)

PR #402, opened 2017-11-15
Fix incorrect behavior when XDG_CONFIG_HOME doesn't end with /

steinuil:
I have $XDG_CONFIG_HOME set to $HOME/.config without the slash at the end (which is how it's defined in [the spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#variables)) so merge-pathnames would think .config is a file and elide it.

Adding a slash to the end would fix this behavior and still work even if the path already ends with a slash.


dbjergaard:
Thank you!

PR #404, opened 2017-11-21
Allow user to change the vertical padding of message windows

juki-pub:
Currently only the horizontal padding can be changed throught the variable `STUMPWM:*MESSAGE-WINDOW-PADDING*`. This adds another variable, `STUMPWM:*MESSAGE-WINDOW-Y-PADDING*` to control the vertical padding.

I just wanted to do something very simple to learn how to contribute things. Let me know if I'm doing something wrong here.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -114,7 +114,8 @@
 
 (defun setup-input-window (screen prompt input)
   "Set the input window up to read input"
-  (let* ((height (font-height (screen-font screen)))
+  (let* ((height (+ (font-height (screen-font screen))
+                    (* *message-window-y-padding* 2)))
          (win (screen-input-window screen)))
     ;; Window dimensions
     (xlib:with-state (win)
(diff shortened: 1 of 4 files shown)


input.lisp
```diff
@@ -262,20 +263,22 @@ match with an element of the completions."
          (full-string-width (+ string-width space-width))
          (pos (input-line-position input))
          (width (+ prompt-width
-                   (max 100 (+ full-string-width space-width tail-width)))))
+                   (max 100 (+ full-string-width space-width tail-width))))
+         (text-y (+ (font-ascent (screen-font screen))
```
PuercoPop: Maybe `height` would be a better name?

juki-pub: I'm not sure if `height` is better, since the variable shows the position of the text, not the height of the window. The window height is set earlier in `SETUP-INPUT-WINDOW` to be fontheight+padding*2 (it doesn't need to be changed when redrawing since the input window is always just one line high).

PuercoPop: Thanks for clearing that up

dbjergaard: Looks good to me!

PR #406, opened 2017-11-26
Allow double quotes as string delimiters for command arguments

juki-pub:
When using the `colon` command or stumpish to execute commands, there currently seems to be no way to enter strings with spaces as a single argument. In interactive use this isn't that big of a problem since you can enter the string in the separate prompt, but when controlling Stumpwm via stumpish it would be nice to be able to, for example, create a group with spaces in the name.

This changes the `ARGUMENT-POP` function so that it allows double quotes to be used as string delimiters. For example, with the `colon` command `gnew "foo bar"`will create a group named `foo bar`. Non-quoted arguments still work the same as before, so this shouldn't break anything.

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -581,7 +581,12 @@ know lisp very well. One might put the following in one's rc file:
 
 (defcommand colon (&optional initial-input) (:rest)
   "Read a command from the user. @var{initial-text} is optional. When
-supplied, the text will appear in the prompt."
+supplied, the text will appear in the prompt.
+
+String arguments with spaces may be passed to the command by
+delimiting them with double quotes. A backslash can be used to escape
+double quotes or backslashes inside the string. This does not apply to
+commands taking :REST or :SHELL type arguments."
   (let ((cmd (completing-read (current-screen) ": " (all-commands) :initial-input (or initial-input ""))))
     (unless cmd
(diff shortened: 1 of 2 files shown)


command.lisp
```diff
@@ -226,17 +226,47 @@ only return active commands."
...
+             ;; string.
+             (let ((start
+                     (loop for i from start to (length input)
+                           for char = (char input i)
+                           do (case char
+                                (#\space)             ;Skip spaces
+                                (#\" (return (1+ i))) ;Start position found
+                                (otherwise (return-from pop-string nil))))))
+               (let ((str (make-string-output-stream)))
+                 (loop for i from start to (length input)
```
PuercoPop: Wouldn't it be better iterate over the string itself?

```lisp
(loop
  :for char :across input
  :for idx :from 0
  :do ...)
```

juki-pub: Is there an easy way to start the iteration from the start position that way? Escaping characters would also require setting a flag for the next iteration.

PuercoPop: No, good point

dbjergaard:
Please update the documentation for colon/input to explain this since it changes the behavior across all input boxes. I'm not sure where in the manual this occurs.

PR #407, opened 2017-11-26
Add support for multi-lines prompt.

ralt:
This is useful for e.g. pinentry that wants to ask for a passphrase
after explaining which key needs to be unlocked.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -244,12 +244,28 @@ match with an element of the completions."
     (let ((k (read-key-no-modifiers)))
       (keycode->character (car k) (xlib:make-state-keys (cdr k))))))
 
+(defun prompt-text-y (index font y-padding)
+  "Calculate the y position of text in a prompt."
+  (+ y-padding
+     (* (font-height font) index)
+     (font-ascent font)))
 
 (defun draw-input-bucket (screen prompt input &optional (tail "") errorp)
   "Draw to the screen's input window the contents of input."
   (let* ((gcontext (screen-message-gc screen))
          (win (screen-input-window screen))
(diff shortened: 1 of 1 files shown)


input.lisp
```diff
@@ -288,14 +302,18 @@ match with an element of the completions."
...
                                       (string char)
                                       :translate #'translate-id
                                       :size 16))
             else
               do (draw-image-glyphs win gcontext (screen-font screen)
                                     x
-                                    (font-ascent (screen-font screen))
+                                    (* (+ *message-window-padding*
+                                          (font-ascent (screen-font screen)))
+                                       prompt-lines-length)
```
Harleqin: That is far enough repetition to warrant a separate function `text-height`.

dbjergaard: abstract this into a function and update the calls to it and I'll merge this PR.

ralt: I guess I need to rebase on top of #404 too, which should solve the repetition issue.

ralt:
@Harleqin @dbjergaard I mostly rewrote the PR after wanting to rebase on top of #404 (conflicts, woo!), which led me to a better understanding of the code, and the new result is much better than the original version, imho.

Please review :)

ralt:
Some quick testing can be done with the following in the REPL:

```
(read-one-line (current-screen) "foo: ")
(read-one-line (current-screen) "foo
bar: "))
(read-one-line (current-screen) "foo
bar

baz: "))
(let ((*message-window-y-padding* 50))
  (read-one-line (current-screen) "foo
bar: ")))
```

PuercoPop:
@ralt Awesome!

Issue #393, opened 2017-08-28
System date doesnt change running SBCL date


dmh43:
Since setting the date with `sudo date +%T -s "9:07:00"` does not update the current time of the running SBCL process, the `format` macro will keep returning the old system date. Accordingly, the `stumpwm:time` will continue to return the old time and date.

I've worked around this by swapping `C-a` for another command which queries `/usr/bin/timedatectl` instead of the running SBCL process.

PuercoPop:
We could use `sb-posix` get the updated system time?

dmh43:
Ah good call, looks like it does change. I'll see if I can find some free time this week to implement a fix

dbjergaard:
can we write a wrapper for get-universal-time in wrappers.lisp that uses sb-posix:time?

dmh43:
Looks like I didnt find the time that week :sweat_smile: Thanks for bringing this back to my attention. Hopefully I'll find some time in the next few weeks :)

PR #409, opened 2017-12-01
create wrapper for getting decoded system time

dmh43:
https://github.com/stumpwm/stumpwm/issues/393

Unfortunately, I have not had time to test this yet, but it seems to compile fine and the changes are quite simple. If anyone can test that it runs as expected and there are no issues with the related time commands that would be awesome. 

I can confirm that using the `sb-posix:time` function fixes the system time change issue described in the bug report.


PuercoPop:
Thanks :pray:

PR #413, opened 2017-12-20
Replaced 2 tab characters with spaces. Removed random form feed character from file.

graysoncroom:
As I was going through screen.lisp I found odd characters in the file. The form feed character I was referring to looked like ^L when viewed from my normal text editor.

Also in the first few lines I noticed there were 2 tab characters among all of the space characters being used to indent. I assume this was a mistake as no one used tabs to indent in lisp.

Not a big deal, but it is good to fix these small details as we find them. :)

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -244,12 +244,28 @@ match with an element of the completions."
     (let ((k (read-key-no-modifiers)))
       (keycode->character (car k) (xlib:make-state-keys (cdr k))))))
 
+(defun prompt-text-y (index font y-padding)
+  "Calculate the y position of text in a prompt."
+  (+ y-padding
+     (* (font-height font) index)
+     (font-ascent font)))
 
 (defun draw-input-bucket (screen prompt input &optional (tail "") errorp)
   "Draw to the screen's input window the contents of input."
   (let* ((gcontext (screen-message-gc screen))
          (win (screen-input-window screen))
-         (prompt-width (text-line-width (screen-font screen) prompt :translate #'translate-id))
+         (font (screen-font screen))
+         (prompt-lines (ppcre:split #\Newline prompt))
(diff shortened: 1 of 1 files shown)


PuercoPop:
@graysoncroom Thanks for taking the time to submit this PR :pray: 

 Lets not remove the formfeed character. The formfeed character is used by (some) Emacs users to divide a file into multiple sections. You can use with `M-x {for,back}ward-page` and `M-x narrow-to-page`. There are even packages to 'pretty print them' https://github.com/wasamasa/form-feed. Although it is common for the `^L` character to be followed by a 4 semi-colon comment with a Title for the section. You can read more about this practice on https://www.emacswiki.org/emacs/PageBreaks and https://www.gnu.org/software/emacs/manual/html_node/emacs/Pages.html

If we are going to remove the pesky tabs, lets remove all the tabs from the codebase in one go. Would you mind fixing the other cases? You can list them with `grep -nP "\t" *.lisp`.

dbjergaard:
@graysoncroom Thanks for your contribution! @PuercoPop hit all the points I was going to make.  I am receptive to this kind of tidying up, and it would be most useful if:
- The linefeed characters remained but you added a comment defining what that section covered (bonus points if you rearrange functions to be more organized)
- Any tab -> space conversions are handled in a way that doesn't break the indentation of the code.  It might be as simple as calling indent-region with the appropriate "only-use-tabs" variable set in Emacs.

I look forward to more contributions!

graysoncroom:
@dbjergaard @PuercoPop Just opened a new pull request -> #414. All I did was replace the spaces as I am not familiar with how Emacs handles form feed characters to change anything related to that. I did my best to replace them without making the code reflect my personal style preferences. 

Looking forward to contributing more to the project in the future. Cheers.

PR #419, opened 2018-01-09
make sure current-window exists when refreshing window

selesdepselesnul:
if current-window exists refresh the window without changing its size, else display message No Current Window.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -65,3 +65,4 @@ Caio Oliveira       caioaao at gmail com
 Thomas Atkinson     thomas at pinegrove io
 Grayson Croom       grayson at gmail com
 Sam Blumenthal      sam.sam.42 at gmail com
+Moch Deden R        moch.deden.r at gmail com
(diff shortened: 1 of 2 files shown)


PuercoPop:
`alexandria:if-let` would make code more concise imho

selesdepselesnul:
thanks for your comment, i commit the new one that use alexandria:if-let :smile:

PuercoPop:
@selesdepselesnul Hey sorry taking a while to get back to you. I was on holiday (so mobile only). [Merged your code with minor modifications](https://github.com/stumpwm/stumpwm/commit/76f20e2bbd2bc4b0a57e7b9dceee6a3ff447251c). Because there is no if-let* and I prefer to keep code short and avoid nesting is possible I went with when-let* and drop the message to the user instead.

 Some minor comments
- Because the STUMPWM package already uses alexandria there was no need to use the package qualifier when reference to the if-let symbol.
- Because if-let is an external symbol to alexandria there is no need to use double colon as a package separator

Thanks for contributing to StumpWM <3

PR #421, opened 2018-01-14
Added the command "gkill-other".

selesdepselesnul:
Kill other groups. All windows in other groups are migrated to the current group.

### DIFF
--- a/group.lisp
+++ b/group.lisp
@@ -542,6 +542,19 @@ The windows will be moved to group \"^B^2*~a^n\"
             (message "Canceled"))
         (message "There's only one group left"))))
 
+(defcommand gkill-other () ()
+  "Kill other groups. All windows in other groups are migrated
+to the current group."
+  (let* ((current-group (current-group))
+         (groups (remove-if (lambda (g) (eq g current-group))
+                            (screen-groups (current-screen)))))
+    (if (> (length groups) 0)
+        (progn
+          (dolist (dead-group groups)
+            (kill-group dead-group current-group))
+          (message "Deleted")) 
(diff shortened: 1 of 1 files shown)


group.lisp
```diff
@@ -542,6 +542,19 @@ The windows will be moved to group \"^B^2*~a^n\"
             (message "Canceled"))
         (message "There's only one group left"))))
 
+(defcommand gkill-other () ()
+  "Kill other groups. All windows in other groups are migrated
+to the current group."
+  (let* ((current-group (current-group))
+         (groups (remove-if (lambda (g) (eq g current-group))
```
PuercoPop: `(remove current-group (screen-groups (current-screen))` would be better

group.lisp
```diff
@@ -542,6 +542,19 @@ The windows will be moved to group \"^B^2*~a^n\"
...
+                            (screen-groups (current-screen)))))
+    (if (> (length groups) 0)
```
PuercoPop: What do you think about removing the messages so that there is only one execution path?

```lisp
(let* ((current-group (current-group))
         (groups (remove current-group
                         (screen-groups (current-screen)))))
    (dolist (dead-group groups)
      (kill-group dead-group current-group)))
```

group.lisp
selesdepselesnul: :+1:

group.lisp
selesdepselesnul: yup, i think removing the message make it more concise and straightforward :smile:

selesdepselesnul:
done :smile:

PR #423, opened 2018-01-16
Simplify describe-key command

selesdepselesnul:
* Using if-let instead of nested let and if
* Bind result of ```(mapcar 'print-key keys)``` in variable


PuercoPop:
@selesdepselesnul could you also uppercase the format directives (~A)?

selesdepselesnul:
done :smile

PR #424, opened 2018-01-19
Add KILL-WINDOWS command

selesdepselesnul:
Kill all windows in the current group

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -995,14 +995,18 @@ window. Default to the current window. if
   (when window
     (xwin-kill (window-xwin window))))
 
-(defun kill-windows (group)
-  "Kill all windows in the group @var{group}"
-  (dolist (window (group-windows group))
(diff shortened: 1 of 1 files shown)


dbjergaard:
Hi,

Please refactor into a defun by removing `current-group` and replace it with a group argument.  This keeps things a bit more general and lets you kill windows in another group.  

If you want to keep the current behavior, add another defcommand calling the refactored function.  Can we name it something a little more clear? `kill-windows-current-group` is a bit verbose, but does explain exactly what it does.

selesdepselesnul:
done :smile:

dbjergaard:
I like your proposal better.  I'm happy to merge as is... Let me know.

selesdepselesnul:
done :smile:

dbjergaard:
Beautiful!

PR #429, opened 2018-01-22
Add GKILL-WITH-WINDOWS and GKILL-OTHER-WITH-WINDOWS command

selesdepselesnul:
* ```gkill-with-windows```
Kill the current group and all of its windows.
* ```gkill-other-with-windows```
Kill other groups and all of its windows.

### DIFF
--- a/group.lisp
+++ b/group.lisp
@@ -551,6 +551,24 @@ to the current group."
     (dolist (dead-group groups)
       (kill-group dead-group current-group))))
 
+(defun kill-group-with-windows (group)
+"Kill the specified group @var{group} and all of its windows"
+  (let ((screen (group-screen group)))
+    (setf (screen-groups screen) (remove group (screen-groups screen)))
+    (netwm-update-groups screen)
+    (netwm-set-group-properties screen)))
+
+(defcommand gkill-with-windows () ()
+"Kill the current group and all of its windows."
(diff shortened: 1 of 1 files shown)


dbjergaard:
Thanks for these and related PRs.  I am extremely grateful for your contributions and I hope they continue.  However, I feel like these last two are over-implementing fairly common tasks that are exposed by the two kill-windows functions that you implemented in the original PR.  Part of the beauty of StumpWM is that it provides an API for you defining your own utilities when you need them. I struggle to see the use case for having so many kill-windows functions. 

That said, I am open to merging these if you can justify a use case!

selesdepselesnul:
When i call ```gkill``` command sometimes what i want exactly is kill the group and all of its window, not moving the windows to the next group and only kill the group, that was the main reason for  ```gkill-with-windows```, and sometimes i want to delete all groups except the one i work on, with the current condition if i want that behaviour i need to switch to the groups one by one and kill all of its windows with command ```kill-windows-current-group``` and then i kill the group one by one, with new command ```gkill-other-with-windows``` i can do it from my current group with only one command directly from the group i currently work on, all these new command just for adding convenient and productivity and i use it in my stumpwmrc file :smile:, but feel free to close this pr if you think these commands are not necessary :smile:

PR #431, opened 2018-01-24
Remove empty documentation in defvar

selesdepselesnul:


fstamour:
Wouln't it be better to actually put something in the string?

PuercoPop:
I agree with @fstamour, lets add a docstring instead.

selesdepselesnul:
I think the variable itself is self explanatory and beside that i have no idea what the comment i need to write :sweat_smile: , any suggestion :smile: for the content of documentation ?

Harleqin:
I guess that you'd need to explain what “depth” means in this context.  I am not familiar with it.  Maybe also describe the effect of setting it too low or too high.

PuercoPop:
I commited a docstring in https://github.com/stumpwm/stumpwm/commit/d614c45decf58093b83fcfeac1bce28b5efcd1dc. AFAIU the reason for having a limit is avoid infinite recursion during alias expansion to hang StumpWM.

@selesdepselesnul @fstamour @Harleqin Feel free suggest any improvements.

Issue #434, opened 2018-01-31
Inconsistency between documentation and behaviour of window placement rules


mtlll:
I've noticed several inconsistencies between the description in the manual of window placement rules, and the actual behaviour defined by the code.

First of all, the raise parameter seems to be entirely inconsequential.

From the documentation:  When non-nil, raise and focus the window in its frame
in SYNC-WINDOW-PLACEMENT:

```lisp
(defun sync-window-placement ()
  "Re-arrange existing windows according to placement rules"
  (dolist (screen *screen-list*)
    (dolist (window (screen-windows screen))
      (multiple-value-bind (to-group frame raise)
          (with-current-screen screen
            (get-window-placement screen window))
        (declare (ignore raise))
        (when to-group
          (unless (eq (window-group window) to-group)
            (move-window-to-group window to-group)))
        (when frame
          (unless (eq (window-frame window) frame)
            (pull-window window frame)))))))
```
raise is explicitly ignored, and indeed, the window is implicitly raised by PULL-WINDOW:

```lisp
(defun pull-window (win &optional (to-frame (tile-group-current-frame (window-group win))))
  (let ((f (window-frame win))
        (group (window-group win)))
    (unless (eq (frame-window to-frame) win)
      (xwin-hide win)
      (setf (window-frame win) to-frame)
      (maximize-window win)
      (when (eq (window-group win) (current-group))
        (xwin-unhide (window-xwin win) (window-parent win)))
      ;; We have to restore the focus after hiding.
      (when (eq win (screen-focus (window-screen win)))
        (screen-set-focus (window-screen win) win))
      (frame-raise-window group to-frame win)
      ;; if win was focused in its old frame then give the old
      ;; frame the frame's last focused window.
      (when (eq (frame-window f) win)
        ;; the current value is no longer valid.
        (setf (frame-window f) nil)
        (frame-raise-window group f (first (frame-windows group f)) nil)))))
```

suggestion: Add an optional parameter (raise-p t) to PULL-WINDOW and pass in raise from SYNC-WINDOW-PLACEMENT
---------
Secondly, the lock parameter also doesn't behave as indicated by the manual:

 lock
    When this is nil, this rule will only match when the current group matches target-group. When non-nil, this rule matches regardless of the group and the window is sent to target-group. If lock and raise are both non-nil, then stumpwm will jump to the specified group and focus the matched window.

in WINDOW-MATCHES-RULE-P:
```lisp
(defun window-matches-rule-p (w rule)
[...]

PR #438, opened 2018-01-31
Window placement improvements

mtlll:
#434


mtlll:
Could use some feedback on the last commit. Would it perhaps be better to do something along the lines of:
```lisp
(frame-raise-window group to-frame win raise-p)
```

instead of omitting the entire call?

dbjergaard:
I will try to look at this tonight. 

    David

dbjergaard:
I think that we should always raise the window when pulling. That’s why the original code ignore the raise request and unconditionally raised the window.

mtlll:
So then it would perhaps be better to conditionally focus the window instead? I think that would be more reasonable. I'll update the PR soon.

PR #442, opened 2018-02-07
Solve error when calling `resize-direction` directly as command and add docstring to it

selesdepselesnul:
There is no problem when i call it within iresize, but error occur when i call `resize-direction` directly as command
error

### DIFF
--- a/iresize.lisp
+++ b/iresize.lisp
@@ -56,8 +56,8 @@
   (draw-frame-outlines (current-group) (current-head)))
 
 (defcommand resize-direction (d)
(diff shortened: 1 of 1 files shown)


PuercoPop:
@selesdepselesnul Can you be my specific as to how to reproduce? I'm doing `C-t ;` -> iresize and everything works (Although I agree that a prompt is an improvement

PuercoPop:
@selesdepselesnul you are right, sorry for not reading more carefuly. :pray:

PR #444, opened 2018-02-13
Add session commands

sdilts:
This pull request adds the ability to cleanly shutdown, restart, and "logout" without root or having to manually close all windows first. Requires the user to use `systemd `and have the `polkit `package installed. See [the Arch wiki](https://wiki.archlinux.org/index.php/allow_users_to_shutdown) for details on how this works. 

This uses the external tool `wmctl` to close all windows before issuing any of the commands, and is needed as a dependency. This functionality should  be implemented in lisp instead, but I am not sure how to get a list of all available windows so that they can be killed.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -66,3 +66,4 @@ Thomas Atkinson     thomas at pinegrove io
 Grayson Croom       grayson at gmail com
 Sam Blumenthal      sam.sam.42 at gmail com
 Moch Deden R        moch.deden.r at gmail com
+Stuart Dilts        stuart.dilts at gmail com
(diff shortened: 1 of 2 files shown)


PuercoPop:
Wouldn't this be best suited for a [contrib](https://github.com/stumpwm/stumpwm-contrib)? Not all users use Linux or SystemD

dbjergaard:
@PuercoPop I agree.  That being said I can't wait to use this! @sdilts are you ok with moving it to a contrib module? I'll merge it ASAP over there.

sdilts:
A pull request has been submitted as a contrib modue.

PR #447, opened 2018-02-26
remove splicing from SPLIT-FRAME

mtlll:
This fixes a bug with unsplitting.

To reproduce:

Make an empty group and split it in half. We'll call our two new frames f1 and f2.
Split f2 in half again, in the same direction as the original split.
Move to one of the smaller frames, and remove it. Instead of returning to our two identical frames f1 and f2, it will make f1 larger to fill the space taken up by the deleted frame.

The reason for this is that the frame tree is not a pure binary tree due to the splicing performed in SPLIT-FRAME.

I have done some testing using this change, and as far as I can see it hasn't introduced any problems.

After some digging, the splicing seems to originate from [this commit](https://github.com/stumpwm/stumpwm/commit/ab3548a02cc9354826169dfa63293b25d3418f8d). No explicit reason is given for the splicing.

Is there some feature that depends on this splicing that I've missed? I suppose there might be a small performance benefit, but since splitting is not performed particularly often, this shouldn't be of any huge importance. And indeed, a pure binary tree seems to me to be the most logical data structure for this.


PuercoPop:
@dbjergaard I agree with the reasoning supporting a binary tree. Can you think of a reason why the current behaviour be preferable?

@mtlll Thanks for digging for the commit! Love the detailed explanation of the PR :tada:

PR #451, opened 2018-04-19
find-asd-file now only loads files with .asd extensions

sdilts:
Previously,` find-asd-file `could return files that contained "asd" in its name. This caused backup files or other unwanted files to be returned, which in turn caused some modules to not be found. This pull request makes `find-asd-file` check the full file extension before returning.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -67,3 +67,4 @@ Grayson Croom       grayson at gmail com
 Sam Blumenthal      sam.sam.42 at gmail com
 Moch Deden R        moch.deden.r at gmail com
 Audun Hoem          audun.hoem at gmail com
+Stuart Dilts        stuart.dilts at gmail com
--- a/module.lisp
+++ b/module.lisp
@@ -79,14 +79,9 @@ called each time StumpWM starts with the argument `*module-dir'"
 
 (defun find-asd-file (path)
   "Returns the first file ending with asd in `PATH', nil else."
-  (flet ((is-asd-file-p (file)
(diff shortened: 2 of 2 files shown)


PuercoPop:
Thanks

PuercoPop:
Indeed, `pathname-type` should be simpler! Want to send the PR?

Harleqin:
I lack the confidence to quickly say which parts of stumpwm I'd need to test.  It _should_ be something like (assuming that case should be regarded as irrelevant):

    (flet ((asd-file-p (path)
             (string-equal (pathname-type path) "asd")))
      …)

Apropos test: are there any opinions on how to move into automatic testing of more than “it compiles”?

PuercoPop:
You have as much confidence as I do ^_^. Its ok if you prefer not to send the PR, I'll add attribution in the commit message.

What I normally do is checkout the PR and run it in under Xephyr, especially if it is not trivial.

> Apropos test: are there any opinions on how to move into automatic testing of more than “it compiles”?

Lack of testing is definitely a problem for me and I think for others as well as it is the first design goal of  [paulownia](https://github.com/stumpwm/paulownia#design-goals). However I don't know where to start to do so. F/e how would the setup/test harness would look like? In what format would we write the tests? Do we use a headless X11 server on a VPS for automated testing or we rely on developers running tests locally on Xephyr. From what I've seen i3 does a great job with integration testing using perl as a 'Selenium for X11'. Would appreciate your input if you have an idea. Feel free to open an issue to further discuss it.

PR #457, opened 2018-05-14
Add XF86XK_AudioMicMute to keysims

sdilts:
Add `XF86XK_AudioMicMute` keysim as defined in `XF86keysym.h`. This file can be found at [freedesktop.org](https://cgit.freedesktop.org/xorg/proto/x11proto/tree/XF86keysym.h) if you want to verify.


sdilts:
Documentation for this key is kind of weird; it is already in stumpwm, but under a different name. I'll go ahead and close this.

PR #460, opened 2018-05-25
Fix checking if a directory contains a .asd file when loading modules

tpine:
Previously the check only checked if asd was contained in the entire
filename. Change this to only check the file type by splitting out the
type from the file before performing the check.


PuercoPop:
@tpine thanks! I was just bitten by this last week. A file under .git was being picked up

tpine:
All good I got bitten by it to.

PR #465, opened 2018-06-07
Bind current key sequence in dyn var ahead of calling a command

kriyative:
This PR adds a dynamic var binding of the current key sequence before invoking any command bound to that key sequence. This feature supports the [REMAP-KEYS](https://github.com/kriyative/stumpwm-contrib/commit/6acf68de83a3e439106b592cd56fd39d23643de6) module.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -231,6 +231,8 @@ The Caller is responsible for setting up the input focus."
       when (typep group (first i))
       collect (second i))))
 
+(defvar *current-key-seq* nil)
+
 (define-stump-event-handler :key-press (code state #|window|#)
   (labels ((get-cmd (code state)
              (with-focus (screen-key-window (current-screen))
@@ -243,7 +245,9 @@ The Caller is responsible for setting up the input focus."
                ((eq cmd t))
                (cmd
                 (unmap-message-window (current-screen))
-                (eval-command cmd t) t)
+                (let ((*current-key-seq* key-seq))
+                  (eval-command cmd t))
+                t)
                (t (message "~{~a ~}not bound." (mapcar 'print-key (nreverse key-seq))))))))))
 
 (defun bytes-to-window (bytes)


events.lisp
```diff
@@ -243,7 +245,9 @@ The Caller is responsible for setting up the input focus."
                ((eq cmd t))
                (cmd
                 (unmap-message-window (current-screen))
-                (eval-command cmd t) t)
+                (let ((*current-key-seq* key-seq))
+                  (eval-command cmd t))
```
dbjergaard: Can we get away with a simple setq instead?

kriyative: Possibly - dynamic binding seemed safer given my limited understanding of reentrancy in the event handling code in StumpWM.

For instance, I see `stumpwm:push-event` which seems to serialize event handlers in the `main-thread` (if I understand this correctly), so the `setq` would probably work fine - but if there's ever a future change to run the events in different threads the dynamic binding would be safer. HTH.

kriyative: Also, there shouldn't be a huge (performance or memory) cost to this specific use of dynamic binding.

kriyative:
Thanks.

kriyative:
@Harleqin - the PR comment describes the usage and links to a contrib module which directly uses this functionality.

> This PR adds a dynamic var binding of the current key sequence before invoking any command bound to that key sequence. This feature supports the [REMAP-KEYS](https://github.com/kriyative/stumpwm-contrib/commit/6acf68de83a3e439106b592cd56fd39d23643de6) module.

Harleqin:
@kriyative Sorry, I had come from a link to the change itself.  I even looked first whether there was an issue that might be related.  Only after I commented on the change, I saw that this comment actually landed in this conversation.

My point is that a PR comment does not show up anywhere in the repository, so that a future developer asking himself what this variable was for will have to do external history research.  If he fails to find a reason for the existence, he might just remove it as dead code.

I strongly recommend documenting this variable, especially because it is intended as part of the external interface.

kriyative:
@Harleqin - fair point, I just didn't want to conflate the specific `remap-keys` use case with the purpose of this variable. How about something like:
```common-lisp
(defvar *current-key-seq* nil
  "The sequence of keys which were used to invoke a command, available
  within the command definition as a dynamic var binding. Commands may
  dispatch further based on the value in *current-key-seq*. See the
  REMAP-KEYS contrib module for a working use case.")
```
The last sentence is optional - I thought it might be helpful to someone who comes across the var to see an example of how it could be used.

@dbjergaard - any thoughts on whether `remap-keys` can remain a module or be integrated into the core of StumpWM?

kriyative:
@Harleqin @dbjergaard -

https://github.com/stumpwm/stumpwm/pull/466

PR #466, opened 2018-06-09
Adds doc-string to *current-key-seq*

kriyative:


kriyative:
@dbjergaard - thanks.

PR #467, opened 2018-06-10
Adds remapped keys feature, with minimal info documentation

kriyative:


sdilts:
Great idea! Having different keybindings across applications is quite annoying, and I can see myself using this. I would mention "keyboard macro" somewhere in the documentation though; it will make this feature easier to find.

I think there should be a built-in way to reverse this, or to turn it on and off. It could be a real pain if you accidentally give it the wrong keybinding to modify, or someone else has to briefly use your computer. In addition, simply placing a flag to control this probably isn't good enough; since stumpwm is self-documenting, the help functions will be incorrect once this is activated, and stay incorrect until `*top-map*` is restored to its original state.

I know you can add and remove keybindings with `define-key`, but I wonder if there is a way to "mask" keybindings similar to minor-modes in emacs; new bindings should be able to be activated and deactivated quite easily. An issue could be raised on the tracker to generate ideas; I don't think easily activating/deactivating keybindings has been implemented. A solution involving`*focus-window-hook*` might be worth looking at as well. Overall, I look forward to having this feature!

dbjergaard:
@kriyative how hard will it be to implement toggling this on and off? We can iron out the other issues after this is merged.

Also, I was hoping to keep this in its own package and just have stump depend on the package in asd.  This is the modern modular way to organize a project and its something we're quasi-working towards.

kriyative:
@sdilts - thanks for the feedback. Toggling the keybindings, the way you describe, would be very nice, but is an orthogonal feature ideally implemented as a separate PR. I could implement an `unbind-remapped-keys` which could release the top level bindings that `define-remapped-keys` added. At this point it would simply find all keybindings for the `send-remapped-key` command and unbind them.

@dbjergaard - I can create a separate `asd` - perhaps a `remap-keys/` subdir with its own `package.lisp` etc.? Or is there an example of this I should follow?

Issue #456, opened 2018-05-06
Add menu to mark multiple values at once


sdilts:
Provide a menu to mark a list of items, and then return a table showing how they were marked. This could be used to implement something similar to Emacs's [buffer menu](https://www.gnu.org/software/emacs/manual/html_node/emacs/Several-Buffers.html), which is accessed by `C-x C-b` in vanilla Emacs. I can see this implemented as a function that is used similarly to `select-from-menu`, and look something like:
``` lisp
(defun select-from-batch-menu (screen table &key prompt 
                                              allowed-markers
                                              initial-selection
                                              extra-keymap)
```
It would return an `alist` with the values in the table grouped under how they were selected:
``` lisp
(select-from-batch-menu (current-screen) '(item1 item2 item3 item4))
 
  ==> ((#\d (item1)) (#\s (item2 item2))) (:not-selected (item4)))
```
An additional function could be provided to help run commands based on what was selected;

sdilts:
I was in the process of implementing this, and realized that it should be discussed before submitting a pull request. You can find my very rough idea on how to do this in my [batch-menu branch](https://github.com/sdilts/stumpwm/commits/batch-menu). It generalizes the code that is already in `menu.lisp` using CLOS, and has the potential to break backwards-compatibility.

PuercoPop:
I like the idea and are eager to read what you've come up with. <3

As for how we should return results, I think that a list of things (which includes alists) would be ideal as it works for the general case (0, 1 or n selections). What do you think about reifying menu entries. Something along the lines of:

```lisp
(defclass menu-entry ()
  ((label :initarg :label :reader menu-entry-label)
   (icon :initarg :icon :reader menu-entry-icon) ; A future addition?
   (selectedp :initform nil :reader menu-entry-selectedp)
   (action :initarg :action :reader menu-entry-action)))

;; Another alternative would be for entries to store value and have the menu
;; be responsible for applying the [same] handler to different values.
```

<hr />

Couple of loose comments:

Prefer using accessors to direct slot access. There is [ WITH-ACCESSORS][ WITH-ACCESSORS], but I've always found it kinda clunky (at least compared to `with-slots`) to use so I just use the accessor or let + accessors when there is multiple calls to the same accessor.

An alist is a list of _pairs_, not a list of lists, so your example above read be:

```lisp
((#\d . (item1))
 (#\s . (item2 item2)))
```

Lisp provides `acons` as a helper to when creating alists.


[...]

PR #470, opened 2018-06-23
Add a menu to select multiple values at once

sdilts:
Adds a menu that allows a user to mark multiple values at once as discussed in issue #456. It adapts what is currently in `menu.lisp`, and rewrites it so that much of the code can be reused. It introduces two new customization variables: `*single-menu-map*` and `*batch-menu-map*`, and two new functions, `select-from-batch-menu`, and `command-menu`. `command-menu` takes care of passing selected menu items to functions, so you don't need to directly call `select-from-batch-menu` and deal with its output. I've tried to document everything that may be confusing, but I may have missed something.


menu-declarations.lisp
```diff
@@ -0,0 +1,190 @@
...
+
+(defgeneric menu-abort (menu)
+  (:documentation "What to do when exiting the menu without results.
+Must signal :menu-quit with the result."))
+
+;; here for single-menu
+(defgeneric menu-backspace (menu)
+  (:documentation "What occurs when backspace is pressed in a menu"))
+
+(defgeneric menu-promt-line (menu)
```
dmh43: maybe this should be `menu-prompt-line`?

PR #479, opened 2018-07-19
PARSE-MODS: Update kbd-parse condition to its new name

PuercoPop:
In 38d19ca7 the condition KBD-PARSE was renamed to KBD-PARSE-ERROR but
one symbol was left unchanged. This made calls like

(stumpwm:kbd "Shift_L-Up")

attempt to signal a non-existent condition. Now it well translate it to
the corresponding KEY structure once again.

Also update a one armed if clause in the loop to when and updated the
indentation to improve clarity.

### DIFF
--- a/kmap.lisp
+++ b/kmap.lisp
@@ -145,7 +145,7 @@ the key failed to parse."
 
 (defun parse-key-seq (keys)
   "KEYS is a key sequence. Parse it and return the list of keys."
-  (mapcar 'parse-key (split-string keys)))
+  (mapcar 'parse-key (split-string keys " ")))
 
 (defun kbd (keys)
   "This compiles a key string into a key structure used by
--- a/kmap.lisp
+++ b/kmap.lisp
@@ -104,40 +104,44 @@ the time these just gets in the way."
     (apply 'xlib:make-state-mask mods)))
 
 (defun report-kbd-parse-error (c stream)
-  (format stream "Failed to parse key string: ~s" (slot-value c 'string)))
+  (format stream "Failed to parse key string: ~S." (slot-value c 'string)))
 
 (define-condition kbd-parse-error (stumpwm-error)
   ((string :initarg :string))
   (:report report-kbd-parse-error)
   (:documentation "Raised when a kbd string failed to parse."))
 
-(defun parse-mods (mods end)
-  "MODS is a sequence of <MOD CHAR> #\- pairs. Return a list suitable
-for passing as the last argument to (apply #'make-key ...)"
-  (unless (evenp end)
-    (signal 'kbd-parse-error :string mods))
-  (apply #'nconc (loop for i from 0 below end by 2
-                       if (char/= (char mods (1+ i)) #\-)
-                       do (signal 'kbd-parse)
-                       collect (case (char mods i)
-                                 (#\M (list :meta t))
-                                 (#\A (list :alt t))
-                                 (#\C (list :control t))
-                                 (#\H (list :hyper t))
-                                 (#\s (list :super t))
-                                 (#\S (list :shift t))
-                                 (t (signal 'kbd-parse-error :string mods))))))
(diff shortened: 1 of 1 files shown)


kmap.lisp
```diff
@@ -117,8 +117,8 @@ for passing as the last argument to (apply #'make-key ...)"
   (unless (evenp end)
     (signal 'kbd-parse-error :string mods))
   (apply #'nconc (loop for i from 0 below end by 2
-                       if (char/= (char mods (1+ i)) #\-)
-                       do (signal 'kbd-parse)
+                       when (char/= (char mods (1+ i)) #\-)
+                         do (signal 'kbd-parse-error)
```
jorams: Signalling `kbd-parse-error` without a `:string` argument leaves the slot in the error unbound, which will trigger an error when printing the condition. This should probably be the same as 3 lines up:

```lisp
(signal 'kbd-parse-error :string mods)
```

A example of when this goes wrong is evaluating `(kbd "Shift_L-Up")` through the `eval` command. The enclosing `handler-case` catches the error and tries to print it.

Thinking about this some more, I'm a bit confused why the code uses `signal` on an `error`. If anything in the callstack is set up to catch `error` or one of its superclasses, the error will be caught. With StumpWM's top-level condition handler, this basically means a call like `(kbd "Shift_L-Up")` will never really succeed inside the StumpWM event loop. When running in the REPL, however, it will appear to run just fine.

TeMPOraL:
Joram Schrijver <notifications@github.com> writes:

> Thinking about this some more, I'm a bit confused why the code uses `signal` on an `error`. If anything in the callstack is set up to catch `error` or one of its superclasses, the error will be caught. With StumpWM's top-level condition handler, this basically means a call like `(kbd "Shift_L-Up")` will never really succeed inside the StumpWM event loop. When running in the REPL, however, it will appear to run just fine.

The distinction between `error` and `signal` should matter only when the
condition reaches top level; if its caught in the middle, it should all
depend on the parent type of the condition.

--
Jacek Złydach
http://jacek.zlydach.pl | http://temporal.pl

jorams:
> The distinction between `error` and `signal` should matter only when the condition reaches top level; if its caught in the middle, it should all depend on the parent type of the condition.

That's true, but in this case the use of `signal` is incorrect. It obscures errors when just testing in the REPL, where the condition will reach top level. In fact, in the commit message @PuercoPop says the following:

> Now it well translate it to the corresponding KEY structure once again.

But that only happens when the condition manages to reach top level. The input string is very much invalid, and should not translate to a `KEY` structure, so `error` is much more appropriate.

This is what currently happens in the REPL:

```lisp
STUMPWM-USER> (kbd "Shift_L-Up")
#S(STUMPWM::KEY
   :KEYSYM 65362
   :SHIFT T
   :CONTROL NIL
   :META NIL
   :ALT NIL
   :HYPER NIL
   :SUPER NIL)
```

That looks correct, but only by coincidence:

```lisp
STUMPWM-USER> (kbd "SoMeChHaAracteers-Up")
#S(STUMPWM::KEY
   :KEYSYM 65362
   :SHIFT T
   :CONTROL T
   :META T
   :ALT T
   :HYPER T
   :SUPER T)
```

PuercoPop:
@jorams Yeah, I realized after your first comment that it worked by interpreting the S in Shift_L and discarding the rest. As you said, there is no good reason for signals to be involved in the regular control flow of a function. I'll rewrite the code to a straight forward style. But that means only supporting the syntax from Emacs' `kdb`.

PuercoPop:
@jorams @TeMPOraL Let me know what you think about the new implementation. I don't like nesting conditionals but I prefer it to having a cond with 6 clauses with small textual differences.

This seems a good opportunity to add some tests so I'll see if I get to it later

PuercoPop:
I was unable to come up with a better solution so I'm closing the PR and opening an issue instead

PR #481, opened 2018-07-21
Store frames as binary trees and make BALANCE-FRAMES work as in Emacs

sjl:
This patch does two related things:

1. Changes frame trees to be stored as binary trees instead of plain old vanilla n-ary trees.
2. Changes balance-frames to match Emacs' balance-frames and fully balance all frames.

Part 1 is what was done in https://github.com/stumpwm/stumpwm/pull/447 which was reverted later because it broke frame balancing.  The upside to storing frames in a binary tree is described in that PR, but in short: it means that if you've got your frames in some layout, then split a frame, and then immediately delete it, you'll end up back in your original layout.

Part 2 fixes the frame balancing to work with the new binary tree scheme.  It makes the `balance-frames` command match the one in Emacs, which balances *all* frames, not just one particular level.  If this isn't desireable and the old level of only balancing some frames is wanted, it should be possible to modify this patch to account for that.  This is mostly just a starting point.

The actual balancing algorithm is a bit wordy because it has to handle the two split cases separately.  Counting the number of side-by-side splits of a given type is also slightly trickier because it's no longer a single list within the tree, but it's not too bad to do it recursively.

I'm not 100% sure this works properly with multiple montior setups.  I'll have to test it on my other setup (which has two separate monitors) when I get a chance.

This is kind of a work-in-progress branch, but I figured I'd open a PR just to see if folks are interested.

### DIFF
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -784,28 +784,68 @@ either :width or :height"
                             (sync-frame-windows group leaf)))))))))
 
 (defun balance-frames-internal (group tree)
-  "Resize all the children of tree to be of equal width or height
-depending on the tree's split direction."
-  (let* ((split-type (tree-split-type tree))
-         (fn (if (eq split-type :column)
-                 'tree-width
-                 'tree-height))
-         (side (if (eq split-type :column)
-                   :right
-                   :bottom))
-         (total (funcall fn tree)))
-    (multiple-value-bind (size rem)
-      (truncate total (length tree))
-      (loop
-        for i in tree
-        for j = rem then (1- j)
-        for totalofs = 0 then (+ totalofs ofs)
-        for ofs = (+ (- size (funcall fn i)) (if (plusp j) 1 0))
-        do
-           (expand-tree i ofs side)
-           (offset-tree-dir i totalofs side)
-           (tree-iterate i (lambda (leaf)
-                             (sync-frame-windows group leaf)))))))
+  "Fully balance all the frames contained in tree."
+  (labels
+      ((balance (tree x y width height)
+         (etypecase tree
+           (frame (balance-frame tree x y width height))
+           (list (balance-tree tree x y width height))))
+       (balance-frame (frame x y width height)
+         (setf (frame-x frame) x
+               (frame-y frame) y
+               (frame-width frame) width
+               (frame-height frame) height)
+         (sync-frame-windows group frame))
+       (count-splits (tree split-type)
+         "Count the number of top-level splits of split-type in tree."
+         (cond ((frame-p tree) 1)
+               ((eql split-type (tree-split-type tree))
+                (+ (count-splits (first tree) split-type)
+                   (count-splits (second tree) split-type)))
+               (t 1)))
+       (divide-dimension (value first-splits second-splits)
+         "Divide a width or height between two sides of a binary tree.
+
+         Returns two values: the number of pixels to give to the first and
+         second child, respectively.
+
+         For example: (divide-dimension 500 3 2) will divide 500 pixels into 300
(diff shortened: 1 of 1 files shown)


scottjad:
> If this isn't desireable and the old level of only balancing some frames is wanted, it should be possible to modify this patch to account for that.

@sjl Any idea how easy it would be to also support the old behavior?

I think both behaviors are useful so I think a second command makes more sense than having a variable to toggle.

Some possible names I thought of: balance-frames-one-direction or balance-frames-single-level.

sjl:
@scottjad The balancing code itself can take any node of the binary tree and will balance its contents, so that doesn't need to change at all.  If you can write logic to select the node you'd like to balance, that should be all that's needed.

The old behavior had a docstring of:

> "Make frames the same height or width in the current frame's subtree."

After reading that and thinking about it, I realized I don't fully understand what it's trying to tell me.  What is "the current frame's subtree"?  My first though would be "the smallest part of the tree that contains the frame" but thinking back on how this used to work, that can't be it.

The code to select the node previously looked like this:

    (tree (tree-parent (tile-group-frame-head group (current-head))
                                  (tile-group-current-frame group))

Which, I *think* did something like:

> Look at the first/top level of splits in the current head, find the split that contains the current frame somewhere within its descendants, and balance that split.

So, for example, if `C` marks the current frame:

    +------------------------------------------------+
    |           |     |     |                   |    |
    +-----------+     |     |                   |    |
    |           |     |     |                   |    |
    |           |     |     |                   |    |
    |           |     |     |                   |    |
    |           |     |     +-------------------+    |
    |           |     |     |CCCCCCCCCCCCCCCCCCC|    |
    |           |     |     +-------------------+    |
    |           |     |     |    |           |  |    |
    |           |     |     |    |           |  |    |
    |           |     |     |    |           |  |    |
    +------------------------------------------------+

Then it would select this for balancing:

    +-----------------------SSSSSSSSSSSSSSSSSSSSS----+
    |           |     |     S                   S    |
    +-----------+     |     S                   S    |
    |           |     |     S                   S    |
    |           |     |     S                   S    |
    |           |     |     S                   S    |
    |           |     |     S-------------------S    |
    |           |     |     SCCCCCCCCCCCCCCCCCCCS    |
    |           |     |     S-------------------S    |
    |           |     |     S    |           |  S    |
    |           |     |     S    |           |  S    |
    |           |     |     S    |           |  S    |
    +-----------------------SSSSSSSSSSSSSSSSSSSSS----+

And the result would end up like this:

    +------------------------------------------------+
    |           |     |     |                   |    |
    +-----------+     |     |                   |    |
    |           |     |     |                   |    |
    |           |     |     +-------------------+    |
    |           |     |     |CCCCCCCCCCCCCCCCCCC|    |
    |           |     |     |CCCCCCCCCCCCCCCCCCC|    |
    |           |     |     |CCCCCCCCCCCCCCCCCCC|    |
    |           |     |     +-------------------+    |
    |           |     |     |      |      |     |    |
    |           |     |     |      |      |     |    |
    |           |     |     |      |      |     |    |
    +------------------------------------------------+

So implementing this would be a matter of writing a function find the current top-level split containing the current frame.  This is basically just `find-if` in the previous code, because all splits of the same orientation were stored in a single list.  But now that it's a binary tree it will be (slightly) more work.

On a side note, I think it might be worth extracting the frame tree stuff into its own file/package.  Right now a lot of code traverses/mutates the frame tree by assuming it's made out of cons cells, which is a little tricky.  Having a separate interface for the tree would make all the code that uses them more resilient to changes in the implementation of the frame tree, and would be a good place to implement functions to find particular nodes like the one I mentioned.  I'm willing to do that refactoring if people think it's worth it.

Issue #453, opened 2018-04-24
Testing


Harleqin:
I propose introducing at least some automated testing.

I'd gradually start with unit and simple integration tests that just run in-process, using something like hu.dwim.stefil or fiveam.  This might get integration up to command level.

Having UI tests would of course add a lot of confidence.

dbjergaard:
Please do! We need tests.  I looked into them in when I was flirting with re-writing stumpwm in paulownia.  Ultimately, the stability of the stumpwm codebase made me realize that this was not the right approach.

I did learn about prove, and you can cannibalize some of the work there to set up a framework.

We have had a few efforts to integrate testing, but they run out of steam.  We have travis-ci to ensure it builds.  If you can give us a PR as POC, I'm happy to add you as a member so you can commit tests directly without going through PRs.

dbjergaard:
@PuercoPop put up some code for setting up testing.  Lets continue discussion in that PR #483

PR #483, opened 2018-07-22
Add tests for PARSE-KEY

PuercoPop:
Also setup the necessary machinery to run the tests through ASDF's
test-op:

  (asdf:test-system "stumpwm")

While trying to rewrite parse-key I kept running into edge-cases. I've added some happy path examples as well as some edge cases to test parse-key to get a test suite going.

Something I'm unsure of, although it makes sense conceptually for the test-op to be on StumpWM, as it currently stands every time we change a file on StumpWM the test-op will recompile the system before running the tests, which are not immediate.

Let me know what you think

Ref. #453

### DIFF
--- a/stumpwm-tests.asd
+++ b/stumpwm-tests.asd
@@ -1,0 +1,10 @@
+(defsystem "stumpwm-tests"
+  :name "StumpWM tests"
+  :serial t
+  :depends-on ("stumpwm"
+               "fiasco")
+  :pathname "tests/"
+  :components ((:file "package")
+               (:file "kmap"))
+  :perform (test-op (o c)
+             (uiop/package:symbol-call "FIASCO" "RUN-TESTS" 'stumpwm-tests)))
--- a/stumpwm.asd
+++ b/stumpwm.asd
@@ -59,4 +59,5 @@
                (:file "remap-keys")
                ;; keep this last so it always gets recompiled if
                ;; anything changes
-               (:file "version")))
+               (:file "version"))
+  :in-order-to ((test-op (test-op "stumpwm-tests"))))
--- a/tests/kmap.lisp
+++ b/tests/kmap.lisp
@@ -1,0 +1,22 @@
+(in-package #:stumpwm-tests)
+
+(defun expand-key-description (&rest desc)
+  (let ((args (list (car desc) :keysym)))
+    (dolist (mod (cdr desc))
+      (push mod args)
+      (push t args))
+    (apply 'stumpwm::make-key (nreverse args))))
+
+(defmacro expect-key (kbd &key to-be)
+  `(is (equalp (stumpwm::parse-key ,kbd) (expand-key-description ,@to-be))))
+
+(fiasco:deftest test-parse-key ()
+
+  (expect-key "C-l" :to-be (108 :control))
+  (expect-key "C-S-l" :to-be (108 :control :shift))
+  (expect-key "C-s-l" :to-be (108 :control :super))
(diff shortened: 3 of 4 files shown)


stumpwm-tests.asd
```diff
@@ -0,0 +1,8 @@
+(defsystem "stumpwm-tests"
```
sjl: If you don't want to add a separate `.asd` file you can name the system with a slash `stumpwm/tests` and put this in the main `.asd` and ASDF will be smart enough to look there when you try to load it, e.g. https://github.com/sjl/bobbin/blob/master/bobbin.asd

More info in the last paragraph of `find-system` in the ASDF manual: https://common-lisp.net/project/asdf/asdf.html#index-find_002dsystem

joaotavora:
> I'm fairly certain that it is used by SISCOG, where Luismbo and Joao works (or used to idk?).

Yes, I can confirm that.

> Fiasco does a good job hiding them if I don't want to use them, but I'd rather remove them from the code.

Open an issue and propose an alternative (though @luismbo would probably want to weigh in on that too).

> And although I originally like the concept of a test being a function, If I can't reuse the same deftest with different parameters there is little to gain from them being functions.

`trace` support and find-definition are two useful things (I don't know if you get that with any other test suites).  When an error or test failure happens, you also get a readable backtrace where your test appears as its own frame.   And you get autodoc too (but if you don't use arguments then I guess It's not super useful).

stumpwm-tests.asd
PuercoPop: I'm aware, but I prefer the one file per project. I remember there was a bug with the logic handling the / in ASDF recently (that has already been fixed) where it treated the slash as a sub-directory. I rather make it 'easier' for ASDF.

sdilts:
Would it be a good idea to run the tests with travis? If we have tests, we might as well run them when we are checking if the project builds or not.

stumpwm.asd
```diff
@@ -59,4 +59,7 @@
                (:file "remap-keys")
                ;; keep this last so it always gets recompiled if
                ;; anything changes
-               (:file "version")))
+               (:file "version"))
+  :in-order-to ((test-op (load-op "stumpwm-tests")))
+  :perform (test-op (o c)
+                    (uiop/package:symbol-call "FIASCO" "RUN-TESTS" 'stumpwm-tests)))
```
jorams: Instead of adding this `:perform` to the `stumpwm` system, I think it should be in the `stumpwm-tests` system. Then the option above it would be:

```lisp
:in-order-to ((test-op (test-op "stumpwm-tests")))
```

PuercoPop: Thanks, updated the PR

PuercoPop:
I had forgotten about Parachute. I've never used it but my experience Shinmera's software has always been well documented and thought out. I wound't mind using it either.

@dbjergaard Let me know what's your final call on the test framework.

<hr />

@jorams afaiu, the reason for Prove's (plan ...) is that it is required by the TAP protocol. It is used as a heuristic to detect when your test suite has hanged. Useful when running on the CI.

dbjergaard:
Fiasco is fine. I will merge this PR soon. Can someone make sure we don’t need to update the build instructions to reflect the new dependency on fiasco?

    David

PuercoPop:
@dbjergaard We don't need to update the update the new build instructions because Fiasco is a dependency of a new system, "stumpwm-tests", not of Stumpwm itself. We should update the travis instructions as @sdilts mentioned though.

PR #492, opened 2018-08-15
 Add *remove-split-hook* 

herbertjones:
I would like the mouse to follow all folder splits and removals so the mouse stays inside the frame.  To do this I've added a hook so that the mouse can be repositioned when a frame is removed.

Example config code I use to get this functionality:
```common-lisp
;; :click, :ignore, :sloppy
(setf *mouse-focus-policy* :sloppy)
(defun temporarilly-disable-sloppy-pointer ()
  "Disable the sloppy pointer for a brief period of time."
  (when (eq *mouse-focus-policy* :sloppy)
    (setf *mouse-focus-policy* :ignore)
    (run-with-timer 0.2 nil #'reenable-sloppy-pointer)))
(defun reenable-sloppy-pointer ()
  (setf *mouse-focus-policy* :sloppy))

(defun mouse-inside-frame-p (frame)
  "Determine if mouse already inside frame."
  (multiple-value-bind (mouse-x mouse-y window)
      (xlib:global-pointer-position  *display*)
    (let* ((group (current-group))
           (min-x (frame-x frame))
           (min-y (frame-display-y group frame))
           (max-x (+ min-x (frame-width frame)))
           (max-y (+ min-y (frame-display-height group frame))))
      (and (<= min-x mouse-x max-x)
           (<= min-y mouse-y max-y)))))

(defun my-banish-frame (frame)
  "Banish mouse to corner of frame"
  (let* ((group (current-group))
         (x-offset 15)
         (y-offset 15)
         (win-x (frame-x frame))
         (win-y (frame-display-y group frame))
         (w (frame-width frame))
         (h (frame-display-height group frame))
         (x (- (+ win-x w)
               x-offset))
         (y (- (+ win-y h)
               y-offset)))
    (temporarilly-disable-sloppy-pointer)
    (ratwarp x y)))

(defun my-focus-frame-hook (cur-frame last-frame)
  (unless (eq cur-frame last-frame)
    (unless (mouse-inside-frame-p cur-frame)
      (my-banish-frame cur-frame))))

(defun my-split-frame-hook (original-frame a-frame b-frame)
  "Reposition the mouse when a frame is created."
  (my-banish-frame a-frame))

(defun my-remove-split-hook (cur-frame old-frame)
  "Reposition the mouse when a frame is removed."
  (my-banish-frame cur-frame))

;; Clear hooks I use on restart
(remove-all-hooks *focus-frame-hook*)
(remove-all-hooks *split-frame-hook*)
[...]

PR #495, opened 2018-08-27
Add policy to keep pointer on current window or frame

herbertjones:
As recommended in #492.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -577,7 +577,9 @@ the window in it's frame."
       (group-focus-window (window-group win) win))))
 
 (define-stump-event-handler :enter-notify (window mode)
-  (when (and window (eq mode :normal) (eq *mouse-focus-policy* :sloppy))
+  (when (and window (eq mode :normal)
+             (and (eq *mouse-focus-policy* :sloppy)
+                  (not *sloppy-mouse-focus-ignored*)))
     (let ((win (find-window window)))
       (when (and win (find win (top-windows)))
         (focus-all win)
--- a/mouse.lisp
+++ b/mouse.lisp
@@ -1,0 +1,164 @@
+;; Copyright (C) 2003-2008 Shawn Betts
+;; Copyright (C) 2017 David Bjergaard
+;;
+;;  This file is part of stumpwm.
+;;
+;; stumpwm is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation; either version 2, or (at your option)
+;; any later version.
+
+;; stumpwm is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this software; see the file COPYING.  If not, see
+;; <http://www.gnu.org/licenses/>.
+
+;; Commentary:
+;;
+;; Functions relating to mouse policies.
+;;
+;; Code:
+(in-package :stumpwm)
+
+(defun temporarilly-disable-sloppy-focus ()
+  "Disable the sloppy pointer for a brief period of time."
+  (setf *sloppy-mouse-focus-ignored* t)
+  (run-with-timer 0.1 nil #'reenable-sloppy-focus))
+
+(defun reenable-sloppy-focus ()
+  "Remove disable flag on sloppy focus policy."
+  (setf *sloppy-mouse-focus-ignored* nil))
+
+;; A few helpers for dealing with window and frame calculations
+(defgeneric mouse-get-x (window)
+  (:documentation "Generic get x position"))
+(defgeneric mouse-get-y (window)
+  (:documentation "Generic get y position"))
+(defgeneric mouse-get-width (window)
+  (:documentation "Generic get width"))
+(defgeneric mouse-get-height (window)
+  (:documentation "Generic get height position"))
+
+(defmethod mouse-get-x ((window xlib:window))
+  (xlib:drawable-x window))
+(defmethod mouse-get-y ((window xlib:window))
+  (xlib:drawable-y window))
+(defmethod mouse-get-width ((window xlib:window))
+  (xlib:drawable-width window))
+(defmethod mouse-get-height ((window xlib:window))
+  (xlib:drawable-height window))
+
+(defmethod mouse-get-x ((frame stumpwm::frame))
+  (frame-x frame))
+(defmethod mouse-get-y ((frame stumpwm::frame))
+  (frame-y frame))
+(defmethod mouse-get-width ((frame stumpwm::frame))
+  (frame-width frame))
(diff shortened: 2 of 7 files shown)


herbertjones:
I think it may be necessary to create generic functions for the window hooks, and on tiled windows use its frame's geometry for detecting if the pointer needs to be moved or not, as windows can be smaller than the frame they occupy.  When coming from the mode-liine, certain windows trigger the banish, I believe because they are smaller than the frame.

mouse.lisp
```diff
@@ -98,18 +98,37 @@
...
+(defmethod mouse-inside-window-p ((window stumpwm::tile-window))
+  (let ((frame (window-frame window)))
+    (mouse-inside-frame-p frame)))
+
+(defgeneric mouse-banish-window (window)
+  (:documentation "Generic banish window"))
+
+(defmethod mouse-banish-window ((window stumpwm::tile-window))
+  (let ((frame (window-frame window)))
+    (unless (mouse-inside-frame-p frame)
```
herbertjones: Should remove this.

events.lisp
```diff
@@ -577,7 +577,9 @@ the window in it's frame."
       (group-focus-window (window-group win) win))))
 
 (define-stump-event-handler :enter-notify (window mode)
-  (when (and window (eq mode :normal) (eq *mouse-focus-policy* :sloppy))
+  (when (and window (eq mode :normal)
+             (and (eq *mouse-focus-policy* :sloppy)
```
PuercoPop: There is not need for the nested and

```
(and window
             (eq mode :normal)
             (eq *mouse-focus-policy* :sloppy)
             (not *sloppy-mouse-focus-ignored*))
```

Works just as well.

events.lisp
```diff
@@ -577,7 +577,9 @@ the window in it's frame."
...
+                  (not *sloppy-mouse-focus-ignored*)))
```
PuercoPop: Instead of this global why not change the `*mouse-policy-focus*` to ignore and then restore the previous policy after the mouse has been re-positioned?

mouse.lisp
```diff
@@ -0,0 +1,137 @@
...
+;;
+;; Code:
+(in-package :stumpwm)
+
+(defun temporarilly-disable-sloppy-focus ()
+  "Disable the sloppy pointer for a brief period of time."
+  (unless (or *sloppy-mouse-focus-ignored*
+              (not (eq *mouse-follow-policy* :follow)))
+    (setf *sloppy-mouse-focus-ignored* t)
+    (run-with-timer 0.1 nil #'reenable-sloppy-focus)))
```
PuercoPop: I'm not keen on having this functionality tied to a timer. I'd prefer if we figure out how handle this with some sort of 'entry'/'exit' hook or `:around` method.

herbertjones:
I'm not currently able to duplicate the erratic behavior of the mouse cursor jumping back and forth between windows.  One way it could happen was to set up a few windows side by side and move them back and forth in-between their frames until it triggered.  It may have been fixed when I moved to using the `xlib:drawable-` functions.  Therefore I've completely removed the timer and global.

mouse.lisp
```diff
@@ -0,0 +1,124 @@
...
+    (unless (or (eq current-window last-window)
+                (mouse-inside-window-p current-window))
+      (mouse-banish-window current-window))))
+
+(defun mouse-handle-focus-group (group)
+  "Disable sloppy pointer when switching groups to prevent floating windows from
+getting stuck and banish to last window or frame."
+  (when (eq *mouse-follow-policy* :follow)
+    (let ((window (current-window)))
+      (if window
```
PuercoPop: :thought_balloon: You can combine the (let () (if ....)) with alexandria's if-let, which is already exported in the StumpWM package.

```lisp
(if-let ((window (current-window)))
  (mouse-banish-window window)
  (mouse-banish-frame (tile-group-current-frame group)))
```

mouse.lisp
```diff
@@ -0,0 +1,124 @@
...
+                    (+ min-y *mouse-follow-banish-y-offset*))))
+    (ratwarp (clamp new-x min-x max-x)
+             (clamp new-y min-y max-y))))
+
+(defun mouse-handle-focus-frame (current-frame last-frame)
+  "Move mouse when moving frames."
+  (when (eq *mouse-follow-policy* :follow)
+    (unless (or (eq current-frame last-frame)
+                (mouse-inside-frame-p current-frame))
+      (let ((current-window (frame-window current-frame)))
```
PuercoPop: Same comment about `if-let` here

mouse.lisp
```diff
@@ -0,0 +1,124 @@
...
+         (new-y (if (minusp *mouse-follow-banish-y-offset*)
+                    (+ max-y *mouse-follow-banish-y-offset*)
+                    (+ min-y *mouse-follow-banish-y-offset*))))
+    (ratwarp (clamp new-x min-x max-x)
+             (clamp new-y min-y max-y))))
+
+(defun mouse-handle-focus-frame (current-frame last-frame)
+  "Move mouse when moving frames."
+  (when (eq *mouse-follow-policy* :follow)
+    (unless (or (eq current-frame last-frame)
```
PuercoPop: We should avoid nesting if when and unless and move the conditions to the outer when.

```
(when (and (eq *mouse-follow-policy* :follow)
           ;; Although frames should be compared by EQUALP afaiu
           (not (eq current-frame last-frame))
           (mouse-outside-frame-p current-frame))) ; or (not (mouse-inside ..))
```

Ideally move the predicate to a function that conveys their intent. Maybe something like

```lisp
(defun mouse-outside-of-old-frame (current-frame last-frame)
  (and (not (eq current-frame last-frame))
       (not (mouse-inside-frame-p current-frame))))
```

(The name is just the first thing I came up with, if you can please come up with a better name for the predicate that communicate what we want to find out)

PuercoPop:
@herbertjones Thanks for the swift response and sorry for taking so long to review the PR. The only I don't love about the current solution is the calls to `mouse-handle-FOO` sprinkled through out the code base, but given that I don't have a better idea the PR is good enough to me as is. But I still want to wait to see what @dbjergaard thinks.

herbertjones:
The erratic mouse behavior seems to have returned.  I'll have to figure out some way of better handling those events without resorting to timers.

dbjergaard:
Unless you have a strong need for this to work on floating windows, I think its most appropriate for the mouse to follow tiled frames.  Conceptually, moving the mouse into floats makes less sense to me.  

I'm not sure why we need a timer to toggle the focus policy.  If I understand your intended behavior it works like this:
- If the frame is split in multiple pieces and user moves the mouse across a frame boundary, the focus follows the mouse by making the frame the mouse is in the current focus (this is the behavior of mouse-focus-policy == :sloppy
- If the frames are split, and the user switches the frame focus then the mouse is warped to the currently focused frame (this is how the mouse follows focus)

I see two ways to expose this behavior to users:
- mouse-focus-policy stays the same, and can be set to its current values with current behavior.  The mouse-follows-focus policy is set separately so the user can compose the behaviors as they wish (I prefer this solution)
- mouse-follows-focus  policy is added as a fourth option to the mouse-focus-policy variables with the combined behavior of :sloppy and mouse-follows-focus described above.

Issue #499, opened 2018-09-01
retaining echoed messages 


erjoalgo:
In my own setup, I made a small enhancement to `echo-string-list` to support providing a macro that retains all messages echoed within a form body:

```lisp
(with-retained-messages t
           (echo "this")
           (echo "message")
           (echo "is")
           (echo "retained"))
```

This allows me to compose commands that echo something to the screen without losing previous messages, and I also found it useful for debugging hooks and events that may fire from various places.

![demo](https://user-images.githubusercontent.com/5349288/44942304-e7abdb00-ad62-11e8-8c72-254a37f23c16.png)

Would it be worth adding this to the main branch? Below is a preview of the patch as I have it now. 

```diff
diff --git a/message-window.lisp b/message-window.lisp
index 1ea7311..60ce3a6 100644
--- a/message-window.lisp
+++ b/message-window.lisp
@@ -27,7 +27,8 @@
 (export '(echo-string
           err
           message
-          gravity-coords))
+          gravity-coords
+          with-retained-messages))
 
 (defun max-width (font l)
   "Return the width of the longest string in L using FONT."
@@ -212,10 +213,39 @@ function expects to be wrapped in a with-state for win."
   (let ((*record-last-msg-override* t))
     (apply 'echo-string-list screen (nth n (screen-last-msg screen)) (nth n (screen-last-msg-highlights screen)))))
 
+(defvar *retain-messages-p* nil
+  "When non-nil, retain old messages in addition to new ones.
+When the value is :log-order, new messages are added to the bottom as in a log file.")
+
+(defmacro with-retained-messages (log-order-p &body body)
+  "Retain all messages sent by (message ...), (echo ...) forms within body
+ without clearing the screen."
+  `(progn
+     ;; clear current messages if not retaining
+     (unless *retain-messages-p*
+       (setf (screen-current-msg (current-screen)) nil
+             (screen-current-msg-highlights (current-screen)) nil))
+     (let ((*retain-messages-p* ,(if log-order-p :log-order t)))
+       ,@body)))
+
 (defun echo-string-list (screen strings &rest highlights)
   "Draw each string in l in the screen's message window. HIGHLIGHT is
   the nth entry to highlight."
   (when strings
+    (when *retain-messages-p*
+      (let (top top-high bot bot-high)
+        (if (eq *retain-messages-p* :log-order)
+            ;; new messages added to the bottom, like a log file
[...]

PR #500, opened 2018-09-01
add with-message-queuing macro

erjoalgo:
add `with-message-queueing` macro:

Queue all messages sent by (MESSAGE ...), (ECHO-STRING ...), (ECHO-STRING-LIST ...) forms within BODY without clobbering earlier messages. When NEW-ON-BOTTOM-P is non-nil, new messages are queued at the bottom.

See #499

### DIFF
--- a/message-window.lisp
+++ b/message-window.lisp
@@ -28,7 +28,8 @@
           err
           message
           gravity-coords
-          with-message-queuing))
+          with-message-queuing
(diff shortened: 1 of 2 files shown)


dbjergaard:
should we also export the variable?

erjoalgo:
thanks for taking this.

Issue #502, opened 2018-09-05
xlib::make-window is undefined in newer xlib versions


sdilts:
I just updated all of my quickllsp distributions, and recompiled stumpwm. I received the following warning message:
```
; file: /home/stuart/Applications/stumpwm/events.lisp
; in: DEFUN STUMPWM::MAKE-XLIB-WINDOW
;     (XLIB::MAKE-WINDOW :ID (XLIB:DRAWABLE-ID STUMPWM::DRAWABLE) :DISPLAY
;      STUMPWM:*DISPLAY*)
; 
; caught STYLE-WARNING:
;   undefined function: XLIB::MAKE-WINDOW
; 
; compilation unit finished
;   Undefined function:
;     XLIB::MAKE-WINDOW
```
It seems as though `xlib:window` is now a class:
```
* (describe 'xlib:window)     

XLIB:WINDOW
  [symbol]

WINDOW names the standard-class #<STANDARD-CLASS XLIB:WINDOW>:
  Direct superclasses: XLIB:DRAWABLE
  No subclasses.
  Not yet finalized.
  No direct slots.

Symbol-plist:
  XLIB::BYTE-WIDTH -> 4
* (describe 'xlib:drawable)

XLIB:DRAWABLE
  [symbol]

DRAWABLE names the standard-class #<STANDARD-CLASS XLIB:DRAWABLE>:
  Direct superclasses: STANDARD-OBJECT
  Direct subclasses: XLIB::BACK-BUFFER, XLIB:PIXMAP, XLIB:WINDOW
  Not yet finalized.
  Direct slots:
    XLIB::ID
      Type: XLIB:RESOURCE-ID
      Initargs: :ID
      Initform: 0
      Readers: XLIB:DRAWABLE-ID
      Writers: (SETF XLIB:DRAWABLE-ID)
    XLIB:DISPLAY
      Type: (OR NULL XLIB:DISPLAY)
      Initargs: :DISPLAY
      Readers: XLIB:DRAWABLE-DISPLAY
    XLIB::PLIST
      Type: LIST
      Readers: XLIB:DRAWABLE-PLIST
      Writers: (SETF XLIB:DRAWABLE-PLIST)
      Documentation:
       Extension hook

Symbol-plist:
  XLIB::BYTE-WIDTH -> 4
```
[...]

PR #503, opened 2018-09-06
Change xlib::make-window to equivalent `make-instance` call

sdilts:
+ may be worth checking if the issues mentioned in #502 are still present in `clx` that were fixed in 
   commit 9061ec2ba3bf72b38848ebf8821da8ca5c51d0ee, which is where this code originated.


PuercoPop:
Thanks for the swift PR

PR #508, opened 2018-09-26
Fix style-warning.

ralt:
The argument is defined so that user-provided hooks can play with it,
but the default function does not use it, so we can declare it as
ignored to remove the style-warning during compilation.


PuercoPop:
Thanks for catching this warning!

Issue #509, opened 2018-09-27
fselect + mouse-click?


ralt:
I'm not sure how it would be feasible to have an `fselect` that lets you click on the frame to select it? I'd be happy to write it, but I need some help to figure out where to start handling the "mouse click on frame" bit.

PR #510, opened 2018-09-27
Make fselect frames clickable.

ralt:
Follow-up of https://github.com/stumpwm/stumpwm/issues/509

This essentially requires the creation of `read-one-char-or-click`, which listens for both "one char" or "click" events together, and returns the appropriate information to `choose-frame-by-number`.

Note: it makes `expose` much more pleasant.


ralt:
The 2nd commit is just a cleanup of existing functions.

dbjergaard:
Are there any parts of this not documented in the manual?

ralt:
Yes, there was! I've pushed a fix.

ralt:
Thanks for the merge!

PR #511, opened 2018-09-28
Declare mode-line and timer before using them.

ralt:
Various mode-line functions and timer functions are used before the
struct is declared, which means the functions were already compiled,
and they thus can't be inlined.

SBCL gives this warning:

; caught STYLE-WARNING:
;   Previously compiled calls to STUMPWM::MODE-LINE-SCREEN,
;   STUMPWM::MODE-LINE-POSITION, STUMPWM::MODE-LINE-HEIGHT,
;   STUMPWM::MODE-LINE-FACTOR and STUMPWM::MODE-LINE-MODE could not be inlined
;   because the structure definition for STUMPWM::MODE-LINE was not yet seen. To
;   avoid this warning, DEFSTRUCT should precede references to the affected
;   functions, or they must be declared locally notinline at each call site.

Which this commit fixes by moving the struct declarations earlier.

It also probably gives a very tiny bit of better performance.

### DIFF
--- a/mode-line.lisp
+++ b/mode-line.lisp
@@ -131,18 +131,6 @@ timer.")
 (defvar *mode-lines* ()
   "All current mode lines.")
 
-(defstruct (mode-line (:constructor %make-mode-line))
-  screen
-  head
-  window
-  format
-  position
-  contents
-  cc
-  height
-  factor
-  (mode :stump))
-
 ;;; Utilities
 
 (defun screen-mode-lines (screen)
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -1264,3 +1264,18 @@ of :error."
 
 (defun command-mode-end-message ()
   (message "Exited command-mode."))
+
+(defstruct (mode-line (:constructor %make-mode-line))
+  screen
+  head
+  window
+  format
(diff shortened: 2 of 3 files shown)


ralt:
Side note: after this, we're down to 1 style-warning left in the full compilation of stumpwm.

```
; compiling (DEFUN ARGUMENT-POP ...)
; file: /home/ralt/common-lisp/stumpwm/command.lisp
; in: DEFUN ARGUMENT-POP
;     (LOOP STUMPWM::FOR STUMPWM::I STUMPWM::FROM STUMPWM::START STUMPWM::BELOW (LENGTH
;                                                                                STUMPWM::INPUT)
;           STUMPWM::FOR CHAR = (CHAR STUMPWM::INPUT STUMPWM::I)
;           DO ...)
; --> BLOCK LET LET TAGBODY 
; ==>
;   NIL
; 
; caught STYLE-WARNING:
;   The binding of I is not a REAL:
;    NIL
;   See also:
;     The SBCL Manual, Node "Handling of Types"
```

I'll try to look into it and fix it in another PR, but it appears to be a bit weird.

PR #512, opened 2018-09-28
Fix style-warning about START being potentially NIL.

ralt:
SBCL doesn't like this LOOP with `return`. It can't correctly guess
the return type and thinks that the START symbol can be bound to NIL.

```
; caught STYLE-WARNING:
;   The binding of I is not a REAL:
;    NIL
;   See also:
;     The SBCL Manual, Node "Handling of Types"
```

By being slightly more defensive (i.e. return NIL from POP-STRING if
the loop returned NIL), we can get rid of the warning.

And with this, we have no STYLE-WARNING left in `make` :-)

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -241,12 +241,13 @@ only return active commands."
              ;; START (if there is one) and the end position of the
              ;; string.
              (let ((start
-                     (loop for i from start below (length input)
-                           for char = (char input i)
-                           do (case char
-                                (#\space)             ;Skip spaces
-                                (#\" (return (1+ i))) ;Start position found
-                                (otherwise (return-from pop-string nil))))))
+                     (or (loop for i from start below (length input)
+                            for char = (char input i)
+                            do (case char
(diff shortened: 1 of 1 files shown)


PuercoPop:
@ralt which version of SBCL are you running? I can't reproduce this warning.

ralt:
1.4.6-1.fc28 (fedora 28)

PuercoPop:
@ralt Could you try with a the most recent SBCL?(1.4.12). I'm a little past 1.4.9 (On 5259d0b to be precise). I've tried building StumpWM from the CLI and `(asdf:make "stumpwm" :force t)` but have been unable to reproduce the warning

dbjergaard:
I don't get why we have the otherwise clause in the loop body as well as the second condition of the or function.

PuercoPop:
@dbjergaard Indeed. Another thing I'm unclear of is why we can't process the whole argument-line at the start? Wouldn't it make more sense?

PR #516, opened 2018-10-11
Add interactive keymap w/o return bindings

szos:
In writing a better key translation module I'm finding I need to make interactive keymaps without RET, C-g, and ESC being bound. I initially implemented this in the module I'm writing, but since it access StumpWM internals I thought it would be better fit for here.

### DIFF
--- a/interactive-keymap.lisp
+++ b/interactive-keymap.lisp
@@ -36,7 +36,10 @@
   (message "~S finished" name)
   (pop-top-map))
 
-(defmacro define-interactive-keymap (name (&key on-enter on-exit abort-if) &body key-bindings)
+(defmacro define-interactive-keymap (name (&key on-enter on-exit abort-if
+                                                (exit-on '((kbd "RET") (kbd "ESC")
+                                                           (kbd "C-g"))))
+                                          &body key-bindings)
   "Declare an interactive keymap mode. This can be used for developing
 interactive modes or command trees, such as @command{iresize}.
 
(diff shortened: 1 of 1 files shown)


PuercoPop:
@szos Thanks for the submitting the PR :tada: May this be the first of many to come.

 I don't think we need to define another macro to achieve what you want. I can think of 2 different approaches. The easiest one would be to bind the exit commands before looping through the provided `key-bindings`. This way the user can overwrite those keybindings if they so desire. This approach would run into problems when the user want to bind the exit-command to a different key which brings me to the second approach, we can add a new keyword argument, `exit-on` that by default would be bound to a list of the current keys. If the user wants to omit the key exit bindings they can pass nil, if they want their own key-bindings they can pass the list of keys they want to bind the `exit-command` to.

I think we should go with the first approach as it just requires moving code. Would that work for you? Do you want to implement the second one instead?

P.D. Regarding the second commit to update your PR with the current master you can rebase your branch using `git rebase master` with your branch checked out. It will sync your branch without an extra merge commit.

szos:
@PuercoPop I think approach one would be best for now, as like you said its the least effort for the payoff. That being said I think I'll take a whack at approach two, as that would provide the most seamless user experience for what I'm working on. 

Thanks for the info, I'm still getting used to github.

interactive-keymap.lisp
```diff
@@ -36,7 +36,10 @@
   (message "~S finished" name)
   (pop-top-map))
 
-(defmacro define-interactive-keymap (name (&key on-enter on-exit abort-if) &body key-bindings)
+(defmacro define-interactive-keymap (name (&key on-enter on-exit abort-if
+                                                (exit-on '((kbd "RET") (kbd "ESC")
```
PuercoPop: Minor style nit `(exit-on (list (kbd "RET" ...))` is preferable to `'((kbd "RET") ...)` as that way the default is the list of values instead of a the code that would produce the list of values.

PR #517, opened 2018-10-11
Added support for a new completion style.

drdo:
The new style, instead of cycling through possible completions,
only completes when there is exactly one match, otherwise it displays
a message window with the possible completions.

Added special variable *input-cycle-completions* that enables
the new completion style when its value is nil.

Added special variable *message-window-input-gravity*
that is to be used as a message window's gravity
when the input window is being displayed so that they do not overlap.

The function message-window-real-gravity should be used to compute
the proper gravity to use when displaying a message window.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -23,7 +23,8 @@
 ;; Code:
 (in-package :stumpwm)
 
-(export '(*input-history-ignore-duplicates*
+(export '(*input-cycle-completions*
+          *input-history-ignore-duplicates*
           *input-map*
           *numpad-map*
           completing-read
@@ -36,6 +37,25 @@
           input-validate-region
           read-one-char
           read-one-line))
+
+;;; General Utilities
+
+(defun take (n list)
+  "Returns a list with the first n elements of the given list."
+  (loop repeat n
+        for x in list
+        collect x))
+
+;; This could use a much more efficient algorithm.
+;; But for our purposes with small sequence sizes it's likely ok.
+(defun longest-common-prefix (seqs &key (test #'eql))
+  "Returns the length of the longest common prefix of the sequences."
+  (flet ((longest-common-prefix-2 (seq1 seq2)
+           (if-let ((i (mismatch seq1 seq2 :test test)))
+             i
+             (length seq1))))
+    (apply #'min (map-product #'longest-common-prefix-2 seqs seqs))))
+
 
 (defstruct input-line
   string position history history-bk password)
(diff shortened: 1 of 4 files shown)


input.lisp
```diff
@@ -90,6 +110,12 @@
 (defvar *input-current-completions-idx* nil
   "The current index in the current completions list.")
 
+(defvar *input-cycle-completions* t
+  "Cycle through completions when there are several possible completions.")
+
+(defvar +input-completion-display-limit+ 64
```
PuercoPop: Please use earmuffs, `*` instead of `+` for this variable as well

input.lisp
```diff
@@ -255,7 +281,9 @@ match with an element of the completions."
                            ((process-input screen prompt input (car key) (cdr key))
                             (if (or (not require-match)
                                     (match-input))
-                                (return (input-line-string input))
+                                (progn
+                                  (format t "lol")
```
PuercoPop: :pray: :scissors:

dbjergaard:
I would like to look at this before it’s merged. 

    David

drdo:
> Thanks for taking the time to hack on StumpWM and submit your PR pray
> 
> I think this is a most needed feature, however I wonder if it could be implemented in a more extensible way, like storing the scoring function in a special variable and the user can chose which strategy to use by setting that variable. What to you think that approach?

What do you mean with "scoring function"?

I agree that this could be done in a more extensible way, we could for example have a special variable with a function that takes a list of completions and an input and whose job is to do the actual completion (or partial completion and option display in the case of this new style).

PuercoPop:
> I would like to look at this before it’s merged.

Sure thing

> What do you mean with "scoring function"?

A function that ranks the candidates according to how well they match the pattern (the current input in this case). 1 being a perfect match, 0 being no relation whatsoever.

> we could for example have a special variable with a function that takes a list of completions and an input and whose job is to do the actual completion

Sounds like a good idea

drdo:
> A function that ranks the candidates according to how well they match the pattern (the current input in this case). 1 being a perfect match, 0 being no relation whatsoever.

Oh, I agree that it would be a good idea to have this, although this is orthogonal to the feature in this pull request.

That said, I don't mind working on that and including it in the PR.

dbjergaard:
Hi,  two thoughts:
1. These are two conceptually different features.  is it hard to separate the window-gravity from the input completion and submit them as separate PRs? If so, we can merge this as is, I just like to keep the PR history clean for future archaeology.
2. I prefer having the default behavior of a variable set to nil, and then setting a variable to true to toggle the new behavior.  If I'm reading it right, it looks like this is the opposite.

Thanks so much for the PR! I like both new features!

PR #519, opened 2018-10-21
Added support for pluggable completion styles.

drdo:
A completion style is anything that implements
input-completion-reset and input-completion-complete.

The default completion style is now called
input-completion-style-cyclic and can be constructed
with make-input-completion-style-cyclic.

Added special variable *input-completion-style*.
Its default value is (make-input-completion-style-cyclic)
to maintain the current behaviour.

Added a new style that instead of cycling through possible completions,
completes as much as is possible unambiguously and otherwise displays
a message window with the possible completions.
This style is called input-completion-style-unambiguous.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -87,8 +87,7 @@ and complete the input by mutating it."))
           (progn
             (let ((elt (nth idx completions)))
               (input-delete-region input 0 (input-point input))
-              (input-insert-string input (if (listp elt) (first elt) elt))
-              (input-insert-char input #\Space))
+              (input-insert-string input (if (listp elt) (first elt) elt)))
             ;; Prepare the next completion
             (setf idx (mod (+ idx (if (eq direction :forward) 1 -1)) completion-count))
           :error)))))
@@ -831,19 +830,20 @@ input (pressing Return), nil otherwise."
 ;; (defun cook-keycode (code state)
 ;;   (values (xlib:keycode->keysym *display* code 0) (x11mod->stumpmod state)))
 
-;; (defun y-or-n-p (message)
-;;   "ask a \"y or n\" question on the current screen and return T if the
-;; user presses 'y'"
-;;   (message "~a(y or n) " message)
-;;   (char= (read-one-char (current-screen))
-;;         #\y))
+(defun y-or-n-p (message)
+  "ask a \"y or n\" question on the current screen and return T if the
+user presses 'y'"
+  (message "~a(y or n) " message)
+  (char= (read-one-char (current-screen))
+        #\y))
 
-;; (defun yes-or-no-p (message)
-;;   "ask a \"yes or no\" question on the current screen and return T if the
-;; user presses 'yes'"
-;;   (loop for line = (read-one-line (current-screen)
-;;                                   (format nil "~a(yes or no) " message))
-;;         until (find line '("yes" "no") :test 'string-equal)
-;;         do (message "Please answer yes or no")
-;;            (sleep 1)
-;;         finally (return (string-equal line "yes"))))
+(defun yes-or-no-p (message)
+  "ask a \"yes or no\" question on the current screen and return T if the
+user presses 'yes'"
+  (loop for line = (read-one-line (current-screen)
+                                  (format nil "~a(yes or no) " message)
+                                  '("yes" "no"))
+        until (find line '("yes" "no") :test 'string-equal)
+        do (message "Please answer yes or no")
+           (sleep 1)
+        finally (return (string-equal line "yes"))))


PuercoPop:
Sorry for taking so long to reply :bowing_man: . I'm not keen on the generic function protocol. Beyond having a `*input-completion-style*` contain a function that takes the prompt input and the completion candidates. Why is it necessary to reset the candidates?

input.lisp
```diff
@@ -768,19 +831,19 @@ input (pressing Return), nil otherwise."
...
-(defun yes-or-no-p (message)
-  "ask a \"yes or no\" question on the current screen and return T if the
-user presses 'yes'"
-  (loop for line = (read-one-line (current-screen)
-                                  (format nil "~a(yes or no) " message))
-        until (find line '("yes" "no") :test 'string-equal)
-        do (message "Please answer yes or no")
-           (sleep 1)
-        finally (return (string-equal line "yes"))))
+;; (defun y-or-n-p (message)
```
PuercoPop: Did you intend to comment this block of code?

input.lisp
```diff
@@ -37,9 +38,88 @@
...
+
+(defun take (n list)
+  "Returns a list with the first n elements of the given list."
+  (loop repeat n
+        for x in list
+        collect x))
+
+;; This could use a much more efficient algorithm.
+;; But for our purposes with small lists it's likely ok.
+(defun longest-common-prefix (seqs &key (test #'eql))
```
PuercoPop: https://github.com/joaotavora/sly/blob/c338f22cc2b88c05b80b417b8a7a13b3e3346ca9/slynk/slynk-completion.lisp#L70

We could re-use this implementation. In particular because it uses reduce instead of apply which can reach the maximum argument limit.

drdo:
> Sorry for taking so long to reply 🙇 . I'm not keen on the generic function protocol. Beyond having a `*input-completion-style*` contain a function that takes the prompt input and the completion candidates. Why is it necessary to reset the candidates?

The reset is necessary because some completion styles need to keep state. In particular the current default and only completion style needs to keep the index to know which of the possible completions it is currently at.

dbjergaard:
Are you proposing the at add slynk as a dependency or that we reuse their code within stump?

Does the other PR still need to be merged?

PuercoPop:
Although it is a bigger change such state should be kept in the input window imho.

drdo:
> Although it is a bigger change such state should be kept in the input window imho.

But the state is entirely depends on the particular completion style.
For example the new completion style introduced in this PR needs no state at all.

PuercoPop:
@drdo I can't see why, but I'm not familiar with that part of the codebase. After you un-uncomment the yes-or-no commands I have no objections with this PR. @dbjergaard its up to you

(And again sorry for taking so long to respond)

PuercoPop:
> Are you proposing the at add slynk as a dependency or that we reuse their code within stump?

Never as a dependency. More like looking at the API for configurable completion functions.

scottjad:
@drdo This is the bug I'm referring to:
```
(defun make-input-completion-style-unambiguous (&key (display-limit 64))
  (make-instance 'input-completion-style-
                 :display-limit display-limit))
```
It should be `'input-completion-style-unambiguous` not `'input-completion-style-`

I was thinking `set-completion-style` (perhaps `set-input-completion-style`) would be something like the following:

```
(defun set-completion-style (style)
  (setf *input-completion-style*
        (ecase style
          (:cyclic (make-input-completion-style-cyclic))
          (:unambiguous (make-input-completion-style-unambiguous)))))
```

Advanced users could still change options or use a custom completion style by setting *input-completion-style* directly. If additional styles are added to stumpwm, they could easily be added to the case expression.

Not sure what a design for a `set-completion-style` that is extensible would look like. There are probably better designs than these, but options I thought of:

Used like this: `(set-completion-style 'input-completion-style-unambiguous :display-limit 10)`

Could be defined like this 

```
(defun set-completion-style (style &rest args)
  (setf *input-completion-style*
        (apply #'make-instance style args)))
```

There's also this option:

```
(defun set-completion-style (style)
  (setf *input-completion-style*
        (case style
          (:cyclic (make-input-completion-style-cyclic))
          (:unambiguous (make-input-completion-style-unambiguous))
          (otherwise style))))
```

Which could be used like `(set-completion-style :cyclic)` or `(set-completion-style (make-input-completion-style-unambiguous :display-limit 10))`

drdo:
@scottjad It would also be possible to define the completion-style protocol functions on the keywords themselves, with some canonical global object for each. Then you could just `(setf *input-completion-style* :unambiguous)`.

jorams:
One annoying change this introduced is that a trailing space was no longer added after the completed input. I just fixed that in 4dd7b0ce1488f1ee0991c0cdceea9398f817641e.

drdo:
> One annoying change this introduced is that a trailing space was no longer added after the completed input. I just fixed that in [4dd7b0c](https://github.com/stumpwm/stumpwm/commit/4dd7b0ce1488f1ee0991c0cdceea9398f817641e).

Interesting of you to mention that.
I personally viewed the extra space as annoying and took the opportunity to fix it.

Different opinions I suppose.

jorams:
Ah, so the change was more intentional than I thought.

The extra space doesn't (normally) change any behaviour, but it does allow to instantly continue typing after completing a command. Typing `<prefix> ; ren<TAB>2` was always a quick way to renumber a window, and removing the trailing space completely killed my muscle memory ("Command 'renumber2' not found"). Meanwhile, `exit ` is equivalent to `exit`.

I never really noticed before, but I now see that consecutive spaces in a command line end up as spaces at the start of a `:rest` argument. I see how that makes it possible for the space to be annoying. Maybe we should make `:rest` skip any leading spaces.

Are there any other cases where the space causes issues?

PR #534, opened 2018-12-05
Add *float-window-modifier* for using modifiers other than super

scottjad:
Previously windows could be moved or resized using the super modifier
plus a mouse click on the window. This variable allows the user to
customize which modifier is used.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -12,6 +12,10 @@
 
 (defvar *float-window-border* 1)
 (defvar *float-window-title-height* 10)
+(defvar *float-window-modifier* :super
+  "The keyboard modifier used to resize and move floating windows
+  without clicking on the top border. Valid values include (but are not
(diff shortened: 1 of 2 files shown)


epanji:
It seem i have to specify optional package :stumpwm on find-symbol function. Otherwise i will get error message like this.

```
When attempting to read the slot's value (slot-value), the slot NIL is
   missing from the object   #S(STUMPWM:MODIFIERS      :META (:MOD-1)
      :ALT NIL      :HYPER NIL      :SUPER (:MOD-4)      :ALTGR (:MOD-5)
      :NUMLOCK (:MOD-2))
```
I don't know why it's behave like that. But it's became normal after i add the optional.
```
(find-symbol (symbol-name *float-window-modifier*) :stumpwm)
```
May be it's good to add the optional if someone else have same error message like me. Thanks.

PR #538, opened 2018-12-14
Fix broken READ-ONE-LINE

shamazmazum:
This patch fixes broken calls to `read-one-line`. Calls in commands.lisp pass 2 arguments, while `read-one-line` now requires three.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -307,10 +307,13 @@ passed the substring to complete on and is expected to return a list
 of matches. If require-match argument is non-nil then the input must
 match with an element of the completions."
   (check-type completions (or list function symbol))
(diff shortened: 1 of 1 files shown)


epanji:
I have tried this PR and it works when \*input-completion-style\* = cyclic.
When I ```(setf *input-completion-style* (make-input-completion-style-unambiguous))``` and type ```eval + TAB```, popup window appeared as expected. But when I enter "interactive command" the popup stay there. Isn't it supposed to disappear?

input

I'm sorry I only reported it and couldn't help because I didn't know how to fix it.

PuercoPop:
@shamazmazum Thanks for catching this. Merged with an updated commit message to make clear which functions were the broken ones.

PR #539, opened 2018-12-15
Rework the logic to *float-window-modifier*

PuercoPop:
Instead of depending that user-friendly name is a keyword with the same
name as the slot in the MODIFIERS Structure make the mapping between the
user-friendly name and the accessor of the MODIFIERS Structure explicit.

https://github.com/stumpwm/stumpwm/pull/534#issuecomment-447211186

@scottjad @epanji What do you think of this solution?


epanji:
Your solution is much cleaner in my humble opinion. But, did we really need to introduce new global variable? \*modifier-accessors\*.

Since in your solution you create new function. How about create function like this:
```
(defun x11-modifier (&optional (mod *float-window-modifier*) (mods *modifiers*))
  "Retrieve the X11 modifier."
  (funcall (symbolicate :modifiers- mod) mods))

;; Then we can call it like this:
;; (x11-modifier) => (:MOD-4)
;; (x11-modifier :meta) => (:MOD-1)
;; (x11-modifier :alt) => NIL
;; (x11-modifier :super) => (:MOD-4)
;; (x11-modifier :hyper) => NIL
;; (x11-modifier :altgr) => (:MOD-5)
```
Or we could add handler-case before funcall to be safe.

With this x11-modifier, we could use it outside float window if needed. But it's not heavily tested yet. Any other opinion is welcome. Thanks.

PuercoPop:
Thanks @epanji, you are right, the extra indirection is unnecessary. I've simplified the code incorporating your input

PR #540, opened 2018-12-16
Simplify the implementation of ALL-MODIFIER-CODES

PuercoPop:
This way `all-modifier-codes` is oblivious what to how many values or in which order `xlib:modifier-mapping` returns.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -755,17 +755,8 @@ input (pressing Return), nil otherwise."
        nil))))
 
 (defun all-modifier-codes ()
-  (multiple-value-bind
-        (shift-codes lock-codes control-codes mod1-codes mod2-codes mod3-codes mod4-codes mod5-codes)
-      (xlib:modifier-mapping *display*)
-    (append shift-codes
-            lock-codes
-            control-codes
-            mod1-codes
-            mod2-codes
-            mod3-codes
-            mod4-codes
-            mod5-codes)))
+  "Return all the keycodes that are associated with a modifier."
+  (flatten (multiple-value-list (xlib:modifier-mapping *display*))))
 
 (defun get-modifier-map ()
   (labels ((find-mod (mod codes)


input.lisp
```diff
@@ -184,8 +184,8 @@ Available completion styles include
 ;;; keysym functions
 
 (defun is-modifier (keycode)
-  "Return t if keycode is a modifier"
-  (or (find keycode *all-modifiers* :test 'eql)
+  "Return T if KEYCODE is a modifier."
+  (or (member keycode *all-modifiers* :test 'eql)
       ;; Treat No Symbol keys as modifiers (and therefore ignorable)
```
Harleqin: This comment implies a little smell: either the function should be named `ignorable-keycode-p`, or it should be named `modifier-p` and the calling site be checking for 0 separately.

input.lisp
```diff
@@ -184,8 +184,8 @@ Available completion styles include
 ;;; keysym functions
 
 (defun is-modifier (keycode)
-  "Return t if keycode is a modifier"
-  (or (find keycode *all-modifiers* :test 'eql)
+  "Return T if KEYCODE is a modifier."
+  (or (member keycode *all-modifiers* :test 'eql)
```
Harleqin: Is there a reason for this change? Is `member` maybe measurably faster than `find`?  Other than that, I'd find it more useful to get the keycode back than the tail of `*all-modifiers*, assuming that the “No Symbol” alternative is removed here.

PuercoPop: > Is there a reason for this change?

Not a strong one just that semantically I care about membership

> Is member maybe measurably faster than find?

I wouldn't expect member to be faster. They both have to traverse the list linearly and upon find it they return. If anything member would return 'more' items and it could put more pressure on the GC, but I doubt there be any difference

:pray: I'll revert this change

input.lisp
PuercoPop: Yeah, I think all the modifier related code could be written in a more straightforward manner. This PR is part of a larger effort I'm trying to simplify it. For example why keep the mapping of keycode to modifer in a struct if I'm 'storing' the struct in a global variable. Using an alist makes more sense to me. Why do we not keep track of the control key? And so forth. But I still need to understand the current code in input.lisp better.

input.lisp
Harleqin: I understand.  For me, the semantic difference between “find” and “member” is not that big (but I'm not a native English speaker).

Thanks for your consideration.  I'm just adding my thoughts, but I do not want to imply that I'd know better than you (and you do all the work here).

PuercoPop: > I'm just adding my thoughts, but I do not want to imply that I'd know better 

Misunderstandings are more likely to happen when communicating over text. I'm sorry if what I wrote gave you the impression that your comment wasn't received. I'm actually grateful when you, or anyone else for that matter, chimes in and reviews a pull request.

dbjergaard: In general, I prefer performance discussions to be framed around empirical evidence that the code in question is a hotspot.  Thanks for taking the time to review this PR!

PR #542, opened 2018-12-18
DRAW-IMAGE-GLYPHS: Restore font to its original value afterwards

PuercoPop:


jorams:
This change is causing StumpWM to crash for me due to some issue with the default 9x15 font (that I never use).

Since font rendering always sets the desired font before rendering, I don't really see a need to restore the value afterwards. What's the motivation for this change?

PuercoPop:
> This change is causing StumpWM to crash for me due to some issue with the default 9x15 font (that I never use).

Sorry about that, I'll revert asap

> What's the motivation for this change?

A weak one, just leaving things like we found it

jorams:
> just leaving things like we found it

In principle that's not a bad idea, but in this case it has the effect of always returning to the default font. Perhaps we could make `set-font` update the `gcontext-font`, then this change should be unable to cause issues.

For reference, the crash was caused by the fact that X triggered a font error, which StumpWM tried to display, which triggered a font error, etc.

PR #544, opened 2018-12-21
Border calculation on float-window

epanji:
___
After I change config to distinct border like this:
```
(setf *normal-border-width* 4
      *transient-border-width* 3
      *float-window-border* 5
      *float-window-title-height* 10)
```
I realize i made mistake with counting windows geometry. Here is what I got:

- It seem \*normal-border-width\* only showed on float-window and not selected tile-window
- frame-height != window-height, it's involve \*normal-border-width\*
- frame-height not always same with window-height  plus 2  \*normal-border-width\*
- geometry-hints always helpful to work with tile-windows

With that discovery I changed the float-window-align and set the geometry so that when adding borders, it won't change the size of the window. Although it change the width and height of slot from window because of addition borders.
___

Set mouse focus policy to :click
```
(setf *mouse-focus-policy* :click)
```
Everything works fine except when I only have one frame, two or more tile-window then execute command ```float-this```,  move or resize float-window and finally click tile-window behind it. The focus did not change to tile-window.

So I change float-window function to find tile-window on list and use it to set in funcall-on-node and set focus on the float-window. Additionally I add check type for float-window to not call ```change-class``` and ```float-window-align```.
___

Sorry about mistake I made concerning my previous PR.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -170,7 +170,7 @@
 
 (defun float-window-align (window)
   (with-accessors ((parent window-parent)
-                   (xwin window-xwin)
+                   (screen window-screen)
                    (width window-width)
                    (height window-height))
       window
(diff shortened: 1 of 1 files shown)


epanji:
```What do we have to take into consideration for the calculations?```

We should distinct content size, border size and overall size. In this case overall size is content plus border size.

```What did we fail to take into consideration previously?```

We did not consider overall size for float-window that make bottom content unable to see before manually resize on some application.

```What are the edge cases you've found out?```

If we open window in float-group or float-this tile-window in single frame, the content will move down because of border/title-height and we can't see bottom content until we resize manually. In some application like emacs, we will lose minibuffer from sight. Then we have to adjust size manually to be able to see minibuffer.

```Why are we multiplying the border-width times 2 in window-display-height and maximize-float```

Because when we calculate height there are two borders involved, top-border/title and bottom-border. And for calculate width, there are two borders too, left-border and right-border. Since we did not distinct left and right, so it is twice border.

```Why is it not enough to align the window in float-this. If I understand you correctly you want to focus the floating window?```

Because the border/align case also affect float-group when open new window/application. And for focus, may be i should make two PR. I only discover it when working with borders.

Note:
I don't mind if you change or add more to correct my code. I understand the hard thing in code is naming. :-)

PuercoPop:
@epanji wanted to let you know I haven't forgotten about this, but with this branch if I float Emacs when I have only one frame without modifying `*normal-border-width*` or *transient-border-width*` I see that window has shrunk a little too much and in the bottom I can see the window behind. This doesn't happen in master.

In the mean time, can you send the focus fix as a separate PR?

epanji:
@PuercoPop sure, will do it after trying a few ideas first.

dbjergaard:
Is this good to go?

PR #548, opened 2018-12-28
Fix focus click issue

epanji:
Apparently the tile-window behind the float-window is not included as a top-windows, so it fails when trying to get a window by parent. 

https://github.com/stumpwm/stumpwm/pull/544#issuecomment-450286794

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -71,3 +71,4 @@ Stuart Dilts        stuart.dilts at gmail com
 Ram Krishnan        kriyative at gmail.com
 Herbert Jones       jones dot herbert at gmail.com
 Daniel Oliveira     drdo at drdo.eu
+Panji Kusuma        epanji at gmail dot com
--- a/events.lisp
+++ b/events.lisp
@@ -592,8 +592,7 @@ the window in it's frame."
 (define-stump-event-handler :button-press (window code x y child time)
   (let ((screen (find-screen window))
         (mode-line (find-mode-line-by-window window))
-        (win (or (find-window-by-parent window (top-windows))
-                 (find-window-by-parent window (group-windows (current-group))))))
+        (win (find-window-by-parent window (top-windows))))
     (cond
       ((and screen (not child))
        (group-button-press (screen-current-group screen) x y :root)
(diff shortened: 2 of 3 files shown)


PuercoPop:
You've found the root cause, but the issue is that the `window-visible-p` method that specializes on tile-windows incorrectly says the the window behind, which is visible is not https://github.com/stumpwm/stumpwm/blob/84f093ebc4dcc571ef6d3ad3fa1afdda2a63fd7b/tile-window.lisp#L39

We need to fix that method instead

epanji:
@PuercoPop I have undo my change and add what you suggested:
```
(defmethod window-visible-p ((window tile-window))
  ;; In this case, visible means the window is the top window in the
  ;; frame. This is not entirely true when it doesn't take up the
  ;; entire frame and there's a window below it.
  (eq window
      ;; Assume window visible if frame-window NIL
      ;; #S(frame 0 NIL 0 0 1366 768)
      (or (frame-window (window-frame window))
          window)))
```
But it's only work after command float-this/float-window and failed for some case like this:

click-2

Note:
The MPlayer control and Firefox are both tile-window. It can't change focus to Firefox by clicking. I don't know if that is expected.

I prefer my changes before because the code only runs if the first attempt fails. But if you have more suggestions, let me know.

epanji:
@PuercoPop I guess, I got the logic for window-visible-p. Please look at the picture below. 

window-visible-p

PuercoPop:
Thanks for following up on this @epanji! :tada: :heart: I've tested it locally and it works! Merged with some minor style changes to the code.

Happy new year!

epanji:
Thanks @PuercoPop and happy new year.

epanji:
@PuercoPop: Unfortunately when you change the style for loop i proposed, you accidentally change the logic. That leads to wrong results.

https://github.com/stumpwm/stumpwm/blob/91b9961ccd1a1a01c4c82032a4ce415eaae0972d/tile-window.lisp#L50-L53

With same style you use, it should be like this.

``` common-lisp
        (loop :for current-window :in frame-windows
              :for next-window :in (rest frame-windows)
              :until (full-frame-p current-window)
              :when (eq window next-window)
                :do (return t))
```
You could refer to pictures in the comments above. Thanks.

PuercoPop:
@epanji Thanks for catching this. Sorry for not being more careful. Your fix is now in master

PR #562, opened 2019-01-10
add :exit-on example for define-interactive-keymap

szos:
adds an example of overriding the default exit keys in define-interactive-keymap which adds SPC to the exit commands. It could also be rewritten without the defparameter, but i found it more readable this way.


dbjergaard:
Thanks! Can you change the bhaz to baz? I prefer the latter spelling.

szos:
most definitely!

PR #565, opened 2019-02-06
add command 'remove-sibling'

szos:
adding the command remove-sibling, which recursively removes the sibling* frame of the current frame. I'm not sure if this functionality exists, but I couldn't find it. I think that having the ability to remove the sibling frame is useful and worth having in stumpwm, but if its better suited to the ~/.stumpwm.d/init.lisp let me know. I generally use this command as a more malleable 'only' command - executing it on a frame removes all other frames in that branch of the frame tree. 

*sibling here refers to whats on the other side of a split, not the frame focused by the sibling command (which seems to always be the 0th frame)

### DIFF
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -1055,9 +1055,8 @@ space."
 (defcommand (remove-sibling tile-group)
     (&optional (group (current-group))
 	       (frame (let ((window-type (type-of (current-window))))
-			           (if (eq window-type 'float-window)
-			               nil ;; (focus-all (current-window))
-			               (window-frame (current-window))))))
+			           (unless (eq window-type 'float-window)
+			             (window-frame (current-window))))))
     ()
   "removes the closest sibling frame to the specified frame, which defaults
 to the current frame. if the closes sibling is split, we call 
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -1052,6 +1052,39 @@ space."
             (show-frame-indicator group))
           (run-hook-with-args *remove-split-hook* l frame)))))
 
+(defcommand (remove-sibling tile-group)
+    (&optional (group (current-group))
+	       (frame (let ((window-type (type-of (current-window))))
+			           (if (eq window-type 'float-window)
+			               nil ;; (focus-all (current-window))
+			               (window-frame (current-window))))))
+    ()
+  "removes the closest sibling frame to the specified frame, which defaults
+to the current frame. if the closes sibling is split, we call 
+remove-split-sibling, otherwise we just remove the sibling frame. "
+  (when frame
+    (labels ((remove-split-sibling (tree group)
+	       (let ((left (car tree))
+		     (right (second tree)))
+		 (when (listp left)
+		   (remove-split-sibling left group))
+		 (when (listp right)
+		   (remove-split-sibling right group))
+		 (when (frame-p left)
+		   (remove-split group left))
+		 (when (frame-p right)
+		   (remove-split group right)))))
+      (let ((sibling
+	     (closest-sibling
+	      (list (tile-group-frame-head group (frame-head group frame)))
+	      frame)))
+	     (cond ((listp sibling) ;; sibling is split
+	            (remove-split-sibling sibling group))
+	           ((frame-p sibling) ;; sibling is a frame
+	            (remove-split group sibling))
+	           (t ;; something went wrong, show the current frame.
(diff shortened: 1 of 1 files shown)


tile-group.lisp
```diff
@@ -1052,6 +1052,39 @@ space."
             (show-frame-indicator group))
           (run-hook-with-args *remove-split-hook* l frame)))))
 
+(defcommand (remove-sibling tile-group)
+    (&optional (group (current-group))
+	       (frame (let ((window-type (type-of (current-window))))
+			           (if (eq window-type 'float-window)
+			               nil ;; (focus-all (current-window))
```
dbjergaard: This should be an unless form instead of if since the true case is nil

tile-group.lisp
```diff
@@ -1052,6 +1052,39 @@ space."
...
+			               (window-frame (current-window))))))
+    ()
```
dbjergaard: Is there a reason we have such a complicated (&optional) form for the arguments?

tile-group.lisp
```diff
@@ -1052,6 +1052,39 @@ space."
...
+	       (let ((left (car tree))
+		     (right (second tree)))
+		 (when (listp left)
+		   (remove-split-sibling left group))
+		 (when (listp right)
+		   (remove-split-sibling right group))
+		 (when (frame-p left)
+		   (remove-split group left))
+		 (when (frame-p right)
+		   (remove-split group right)))))
```
dbjergaard: is there a reason this isn't a cond?

dbjergaard:
Thanks so much for your PR.  I'm a bit confused what the wider use case is? 

On the one hand, it seems appropriate as an extension, on the other hand, its only one function so maybe adding it is best?

Also, it needs an entry in stumpwm.info.in so that it gets into the manual properly.

tile-group.lisp
szos: I had originally used a function (current-frame), but realized as I was submitting the pull request that that function was in my init file, not stumpwm itself.

tile-group.lisp
szos: This should be, yes. This was hastily tacked on at the last minute to replace a call to (current-frame), which I had implemented and forgotten wasnt a stumpwm function. this is why there is a commented out call to focus-all.

tile-group.lisp
szos: When I write this as 
`(cond ((frame-p left) (remove-split group left)) 
           ((frame-p right) (remove-split group right)))`
and try to call it on a sibling frame that has been split, it removes only the split in the sibling frame, not the sibling frame itself, wheras when its implemented using two when statements, it removes both the split in the sibling frame, and the split that has created the sibling frame. I'm sorry im having a hard time expressing this well.

szos:
the use case comes from constantly switching to frames just to remove them. if I'm in a frame, and want to have it take up more of the screen, I have to switch to the frame and remove it. if the sibling frame is split, I have to remove each split myself. I originally tried cobbling it together with the `sibling` command, but it always wanted to select the 0th frame. 

I hadnt implemented it as a module for just that reason (thats what you mean by extension right?)

additionally, where could I find more info about building a manual entry for this? When I browse the manual source I see lots of things that I assume are used to link between source code and the manual (like all the @ symbols), and I'm not sure how exactly to go about adding a (assumedly dynamic) manual entry.

tile-group.lisp
```diff
@@ -1052,6 +1052,38 @@ space."
...
+(defcommand (remove-sibling tile-group)
+    (&optional (group (current-group))
+	       (frame (let ((window-type (type-of (current-window))))
+			           (unless (eq window-type 'float-window)
+			             (window-frame (current-window))))))
+    ()
+  "removes the closest sibling frame to the specified frame, which defaults
+to the current frame. if the closes sibling is split, we call 
+remove-split-sibling, otherwise we just remove the sibling frame. "
+  (when frame
```
dbjergaard: Please make the doc string more descriptive of how to use this and what it is for. This is what will get compiled into the manual when its produced.  Include the rationale that you used to justify the pull request so that new users will understand what its for.

To add this to the manual, edit stumpwm.texi.in, find the appropriate place (I would put it near remove-split) and add this:
```
!!! remove-sibling
```

szos:
after thinking about this some more, I think this is best suited for the rc file. its one function, and the use case is niche at best. I dont want to add any unnecessary clutter.

PR #569, opened 2019-02-12
exit keymap via specific bindings.

szos:
This adds the command call-and-exit-kmap, which allows us to run a command, and then run the exit command for the appropriate interactive keymap. It also modifies define-interactive-keymap so one can specify specific bindings that exit the keymap when run, by virtue of passing them to call-and-exit-kmap. This action is triggered by appending any true value to the end of the key binding. 
in the below example, pressing 'n' will run 'meta n', while pressing m will run 'meta x' and then exit the interactive keymap. 
(define-interactive-keymap example ()
  ((kbd "n") "meta N")
  ((kbd "m") "meta x" t))

this modification of define-interactive-keymap preserves the original usage, and doesnt break existing code using it.

### DIFF
--- a/interactive-keymap.lisp
+++ b/interactive-keymap.lisp
@@ -38,6 +38,9 @@
 
 (defcommand call-and-exit-kmap (function exit-map) ((:command "command to run: ")
                                                    (:command "keymap to exit: "))
+  "This command effectively calls two other commands in succession, via run-commands.
+it is designed for use in the define-interactive-keymap macro, to implement exiting
+the keymap on keypress. "
   (run-commands function exit-command))
 
 (defmacro define-interactive-keymap
--- a/interactive-keymap.lisp
+++ b/interactive-keymap.lisp
@@ -36,6 +36,10 @@
   (message "~S finished" name)
   (pop-top-map))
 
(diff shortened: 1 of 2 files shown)


dbjergaard:
thanks for the PR, can you help me understand your use case a little better?

szos:
my use case comes from the differences between Emacs kbd function and StumpWMs kbd function. I want to be able to have key bindings like "C-x u", and currently I do that with (make-sparse-keymap), and hang that map off of C-x. Taking my interactive-keymap for moving around frames, windows, and groups, I have some bindings to undo what I've done, and when I call those, id like to exit the keymap, so my workflow would be:  move frames around -> undo and exit VS move frames around ->undo -> exit. there are several interactve-keymaps that I want to add this functionality to, such as my keymap for managing floating windows in tiled groups. 

When i was playing around with implementing this, I first tried adding a :one-shot key variable, to indicate that it was to exit after one key press, but found that to limiting. this is probably a niche use case.

dbjergaard:
Add the defcommand to the manual with an updated docstring and I'll merge this.

jorams:
I just pushed d6057a9bcbf299b7013bfdc0d093406871a7c98d to fix the definition of `call-and-exit-kmap`. It was causing a style warning during compilation and an error at runtime.

PR #570, opened 2019-02-13
Generate the heads using Xrandr instead of Xinerama.

ralt:
Xinerama has several drawbacks, including that it doesn't support
dynamic reconfiguration of monitors by default (e.g. the list of
numbers is random).

Using RandR API, we notably gain the ability to get the EDID of the
monitors, which means we'll be able to store where frames were between
monitors plugs and restore them when it makes sense, for example.

Given that older machines might not support RandR (it's fairly
"young"), Xinerama support is still left in the existing
implementation.

PS: I also sneak in a trailing whitespace removal, sorry :stuck_out_tongue:

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -103,7 +103,7 @@
         ((equalp old-heads new-heads)
          (dformat 3 "Bogus configure-notify on root window of ~S~%" screen) t)
         (t
-         (dformat 1 "Updating Xinerama configuration for ~S.~%" screen)
+         (dformat 1 "Updating Xrandr or Xinerama configuration for ~S.~%" screen)
          (if new-heads
              (progn (head-force-refresh screen new-heads)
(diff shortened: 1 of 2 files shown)


ralt:
Also, I did see https://github.com/stumpwm/stumpwm/pull/328 but `(xinerama:xinerama-query-screens *display*)` was returning the wrong results, e.g. it still returned 2 heads after unplugging my external monitor.

ralt:
Related: I tried to use `(loop ... collect)`, but it doesn't seem to have a way to play with `multiple-value-bind`.

dbjergaard:
THANK YOU SO MUCH! This is awesome!

ralt:
Lol, glad you're liking it, and thanks for the merge!

Issue #564, opened 2019-01-31
select-from-menu alist documentation


noogie13:
> Prompt the user to select from a menu on SCREEN. TABLE can be
a list of values or an alist. If it's an alist, the CAR of each
element is displayed in the menu. What is displayed as menu items
must be strings.

This is part of the current documentation of the `select-from-menu` function, but if you pass in an alist `'(("something" . "else"))` (a form I'm using with else being a URI), searching is messed up. The reason is that the `typing-action` method uses `(second table-item)` which doesn't work with an alist. 

Changing the list form to `'(("something" "else"))` works fine, just thought it might be worth a mention, because the documentation doesn't seem exactly verbose.

sdilts:
Thanks for catching this! I'm calling this an error in our documentation. This behavior is present in `select-from-menu` before the menu commands were heavily modified in #470 . For reference, I tested this at cec7fd9e963e5da59b05ba0ffbe292dc94fd2947, the latest commit before this pull request. In addition, we use a list in multiple places including:
https://github.com/stumpwm/stumpwm/blob/cec7fd9e963e5da59b05ba0ffbe292dc94fd2947/group.lisp#L494-L498
and
https://github.com/stumpwm/stumpwm/blob/cec7fd9e963e5da59b05ba0ffbe292dc94fd2947/window.lisp#L972-L979

sdilts:
There is more to this than just the alist part. The documentation also claims that a list of values also works as input, but this crashes when you attempt to search through the table.Run
``` lisp
(select-from-menu (current-screen) (list '("a" "f") "b" "c"))
```
and type any character to cause this crash.

noogie13:
I see. Thanks!

PR #571, opened 2019-02-14
Update select-from-menu documentation to reflect reality

sdilts:
The `table` argument to `select-from-menu` uses a nested list, not an alist. See issue #564 for more details.


PuercoPop:
Thanks for updating the documentation. :tada:

PR #573, opened 2019-02-14
head: use normal counters for xrandr heads.

ralt:
It turns out that the EDID property returned by CLX is not unique at
all.

To fix this bug, we fallback to the previous behavior of using a
simple counter to assign the head number.

That said, because it can be useful later on, a new NAME property is
added to the head struct. _That_ property is unique, but turning it
into a number requires hashing it, which we might want to avoid as a
first step. (So, right now, it is only useful for introspection purposes.)

### DIFF
--- a/head.lisp
+++ b/head.lisp
@@ -38,56 +38,33 @@
              :height (xinerama:screen-info-height screen-info)
              :window nil))
 
-(defun make-screen-randr-heads (root)
+(defun output->head (output count)
   (multiple-value-bind
-        (timestamp config-timestamp crtcs outputs)
-      (xlib:rr-get-screen-resources root)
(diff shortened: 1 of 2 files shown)


PuercoPop:
@ralt I've pushed a commit refactoring to make the code less dense. One thing that may be controversial is that to avoid going over the 80-char limit for the first m-v-b in `output->head` I used the throwaway names (`_0`, `_1`, etc) for the variables we are going to ignore anyway. Let me know what you think @ralt @dbjergaard @jorams

ralt:
I totally love it! Just a small detail: the first m-v-b defines `_` but you ignore `_0`.

PuercoPop:
Thanks for catching. Merged :tada:

ralt:
Thanks!

Issue #252, opened 2016-05-06
mode-lines are updated on each mousewheel event


NicolasPetton:
When scrolling with the mouse wheel, all mode-lines are updated (many, many times, since many events are triggered when scrolling).

This makes scrolling really slow with some applications like chromium, and consumes a lot of cpu, depending on how many elements are drawn in the mode-line.

It looks like an event `:BUTTON-PRESS` is triggered on mouse wheel, which results in calling `UPDATE-ALL-MODE-LINES` for each one of them.

dbjergaard:
I'll look into this.  I can see that this would be very frustrating.  The modeline should definitely not refresh itself when the mouse wheel triggers the event.  Unfortunately :BUTTON-PRESS is a generic event for any button, so it can make sense to have the modeline update itself when it gets one of these events in case you destroy a window or trigger some other change to the modeline.

Without delving too deep into the code, I assume this choice was made to make the modeline responsive to user changes without having complex code paths signaling when the modeline should re-draw.

NicolasPetton:
@dbjergaard any news about this issue?  Can I help in some way?

dbjergaard:
Sorry, no news. I took a brief look at the `:button-press` event handler and it looks like it doesn't directly call update-all-mode-lines.  That event handler should only fire when you mouse over the mode-line.  Can you let me know how you determined which event is triggered by the mousewheel?

NicolasPetton:
> Sorry, no news.

Can you reproduce the issue?

dbjergaard:
No I can't I opened google-chrome and scrolled a long webpage but stumpwm didn't become unresponsive.  Can you give me a breakdown of:
- how you start stumpwm/post your stumpwmrc
- you computer specs
- a bullet proof procedure for reproducing

jorams:
I did some quick testing. The `group-button-press` method for tiling groups calls `update-all-mode-lines`. Because of that you can trigger a ton of updates by scrolling in an empty frame. I can't, however, get it to trigger when scrolling in a window.

NicolasPetton:
Joram Schrijver <notifications@github.com> writes:

> I did some quick testing. The `group-button-press` method for tiling
> groups calls `update-all-mode-lines`. Because of that you can trigger
> a ton of updates by scrolling in an empty frame. I can't, however, get
> it to trigger when scrolling in a window.

I was using NixOS back when I reported this issue, and I'm now using
Fedora, but I will try to reproduce it and give you a recipe.

[...]

PR #577, opened 2019-02-15
Make the mode-line faster and less intrusive.

ralt:
Essentially, instead of redrawing the mode-line every time it is asked
to, a new thread is spawned for each mode-line, and that thread
redraws the mode-line every *mode-line-refresh-interval* (every 200ms
by default).

In practice, 200ms is not noticeable by the human eye. If you have
great eyes, it is very easy to reduce the interval.

Testing with this change, and the following mode-line:

```
(ql:quickload :local-time)

(defun get-date-modeline ()
  (local-time:format-timestring
   nil
   (local-time:now)
   :format '(:short-weekday #\space :short-month #\space
	     (:day 2 #\space) #\space (:hour 2) #\: (:min 2)
	     #\: (:sec 2))))

(setf *screen-mode-line-format*
      (list "^n^b %g ^n^b %W ^> "
            "  "
            "^n^b%B "
            '(:eval (get-date-modeline))))
```

I now have:

- The Slack web application no longer lags when scrolling. I believe
  it could solve the issue mentioned in #252 for a few people.
- The mode-line updates the time with the new second constantly.

I understand that this is a fairly consequential change, so I don't
expect that to be merged without discussion :-)

Also worth noting that the code in this PR takes special care in not breaking BC for the existing function signatures.

### DIFF
--- a/mode-line.lisp
+++ b/mode-line.lisp
@@ -27,6 +27,8 @@
           *mode-line-pad-y*
           *mode-line-position*
           *mode-line-timeout*
+          *mode-line-refresh-interval*
+          *mode-line-throttle-interval*
           *screen-mode-line-format*
           *screen-mode-line-formatters*
           add-screen-mode-line-formatter
@@ -118,6 +120,12 @@ timer.")
 (defvar *mode-line-timer* nil
   "The timer that updates the modeline")
 
+(defvar *mode-line-refresh-interval* 0.3
+  "The interval at which the mode lines should be refreshed.")
+
+(defvar *mode-line-throttle-interval* 0.04
+  "The interval at which the mode lines should be throttled.")
+
 ;;; Formatters
 
 (defun add-screen-mode-line-formatter (character fmt-fun)
@@ -146,13 +154,6 @@ timer.")
 
 (defun mode-line-gc (ml)
   (ccontext-gc (mode-line-cc ml)))
-
-(defun turn-on-mode-line-timer ()
-  (when (timer-p *mode-line-timer*)
-    (cancel-timer *mode-line-timer*))
-  (setf *mode-line-timer* (run-with-timer *mode-line-timeout*
-                                          *mode-line-timeout*
-                                          'update-all-mode-lines)))
 
 (defun maybe-cancel-mode-line-timer ()
   (unless *mode-lines*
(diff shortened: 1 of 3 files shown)


ralt:
cc @PuercoPop because you mentioned that approach here https://github.com/stumpwm/stumpwm/issues/252#issuecomment-427936639

sdilts:
I just found another solution to this; you can actually check to see which button is pressed on each button event. Another proposed fix is just to not update the modeline on scroll button events. Of course, there are other benefits to updating the modeline every `n` seconds. A pull request is forthcoming.

dbjergaard:
This was going to be my proposed solution (an update thread with some default refresh interval). Let me review it and we’ll get this merged soon. 

    David

ralt:
@dbjergaard I changed the approach a little bit: 200ms was noticeable when switching windows, i.e. the highlighted window was lagging behind.

I kept the 200ms refresh (actually, made it 300ms), but also let the (update-mode-lines) calls go through, with some throttling. The throttling is 0.04s because we can in theory see 24 frames per second.

dbjergaard:
This looks great! You remove mode-line-timer.  Is there a reason for this? It seems like users might want to hang stuff off of that, and removing it could make it harder to customize the mode-line.

dbjergaard:
Another question: Does this just set up a queue for mode-line refresh events and throttles them if the thread gets spammed?

ralt:
I removed `*mode-line-timer*` because it is dead code. I am probably missing a reason on why it should be kept?

And no, the code doesn't queue anything, it just ignores throttled requests, but they will be handled after <300ms.

ralt:
Said another way: the waitqueue is only used to wake up the main mode-line thread. If the thread was last woken up in the previous 40ms, it doesn't do anything (that's the throttling), otherwise it refreshes the mode-line. If nothing happens for 300ms, the thread refreshes the mode-line anyway.

ralt:
But you're right, I'll make it so that if an event is skipped, we only wait for 40ms instead of 300ms, this way the missed events will get a quick update. After no event is skipped, we go back to waiting 300ms.

ralt:
PR has been updated.

jorams:
I'd like to discuss the option of reverting this change, as it's seriously affecting the stability of the mode-line. None of the code doing formatting for the mode-line expects to run in a multi-threaded environment, and there's a lot of global state being read and manipulated all over the place, without any locks being held. The issues referencing this PR are some examples. The MPD module is also completely broken when multiple mode-lines are opened.

Maybe it's possible to throttle the refreshes of the mode-line without running it in a separate thread?

jorams:
If there are no objections I'm going to go ahead and revert this in a few days.

dbjergaard:
Please revert immediately. Let’s start a new issue/project for revamping the modeline. It is something that needs to be addressed. 

    David

jorams:
Reverted by #599. Thanks for doing the actual work @PuercoPop :)

PR #578, opened 2019-02-16
Don't update the modeline on mouse scroll events

### LINKED ISSUE #252: mode-lines are updated on each mousewheel event

sdilts:
I don't have any lag on my system to actually check if this fixes #252, but I can confirm that the modeline no longer updates on each scroll event. Even if we go with the solution written by @ralt in #577, the first commit in this series should probably stick around.

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -589,18 +589,32 @@ the window in it's frame."
         (focus-all win)
         (update-all-mode-lines)))))
 
+(defun decode-button-code (code)
+  "Translate the mouse button number into a more readable format"
+  (ecase code
+    (1 :left-button)
+    (2 :middle-button)
+    (3 :right-button)
+    (4 :wheel-up)
+    (5 :wheel-down)
+    (6 :wheel-left)
+    (7 :wheel-right)
+    (8 :browser-back)
+    (9 :browser-front)))
+
 (define-stump-event-handler :button-press (window code x y child time)
-  (let ((screen (find-screen window))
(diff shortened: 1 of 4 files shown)


ralt:
I've been running this branch for an hour without any issue so far.

events.lisp
```diff
@@ -589,18 +589,37 @@ the window in it's frame."
...
 (define-stump-event-handler :button-press (window code x y child time)
-  (let ((screen (find-screen window))
+  (let ((button (decode-button-code code))
+        (screen (find-screen window))
         (mode-line (find-mode-line-by-window window))
         (win (find-window-by-parent window (top-windows))))
     (cond
       ((and screen (not child))
-       (group-button-press (screen-current-group screen) x y :root)
+       (group-button-press (screen-current-group screen) button x y :root)
```
ralt: I'm wondering if you shouldn't bail out right there instead of in `group-button-press`? I don't think triggering `click-hook` makes sense if the call is ignored in the end, for example.

ralt: Although I guess it's still focusing the window, so nevermind me.

PR #585, opened 2019-03-14
Add *click-hook*

erjoalgo:
Support running custom code after a click event.


erjoalgo:
I am wondering if there is a better or existing way to do this? I needed something like this to support capturing click events to define a scrot bounding box in environments where I don't have a mouse.

dbjergaard:
I’m a bit concerned about adding this. Cab we discuss your use case and see if there’s a better way?

erjoalgo:
Basically, I would like to run some custom code whenever there is a click event. Is there another way to do this?

PR #590, opened 2019-03-19
Improve search of modifier keys in get-modifier-map

glv2:
This patch allows detecting that ```ISO_Level3_Shift``` is ```mod5``` on my machine.

I needed another modification to make the AltGr-key combinations return the right symbols: replacing ```(base (if altgr-p 2 0))``` by ```(base (if altgr-p 4 0))``` in the ```code-state->key``` function.
However I didn't put it in this pull request, as I have no idea if 4 is the right value for all mappings or if there is a way to detect/compute what the right value for ```base``` is.

For information, I'm running XFCE 4 as desktop manager with stumpwm instead of xfwm4 as window manager, with a french keyboard layout.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -763,7 +763,9 @@ input (pressing Return), nil otherwise."
 
 (defun get-modifier-map ()
   (labels ((find-mod (mod codes)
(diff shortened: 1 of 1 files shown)


PuercoPop:
Thanks for the PR!

PR #599, opened 2019-04-10
Revert threaded modeline

PuercoPop:
Aside from the initial commit I'm reverting a couple more that fix some of the issues reported.


mtlll:
We should still destroy all mode lines on soft-restart, as this is still broken with mode lines enabled, and apparently has nothing to do with the threaded version. So maybe re-apply the DESTROY-ALL-MODE-LINES change?

jorams:
@mtlll Thank you, you seem to be correct. I've reverted the revert of that commit in 319e22c693a37857118349ad6a03e4308c86a573.

PR #603, opened 2019-04-12
Update README.md

KoltPenny:


jorams:
Thank you for your contribution. Since the only thing `autogen.sh` does is call `autoconf`, I don't think this change is necessary.

PR #607, opened 2019-05-02
Alter window-raising behaviour slightly (fixes #469)

### LINKED ISSUE #469: How to keep a (notification) window on top?
I'm using [dunst](https://dunst-project.org) to get notification popups (for things like IRC mentions, calendar events, etc).  It mostly works with StumpWM, with one problem.

The notifications appear in the top right frame (good), but if I focus that frame while the notification window is up, then whatever window is in that frame will raise on top of the notification window.  If more notifications come in before the others have gone away, they get put into that same notification window, which is now below the other window and so I can't see them.  Once the notifications have timed out (or I clear them) any new ones coming in will create a new window, and the whole cycle will start again.

Is there a way I can tell StumpWM to keep the dunst notification window on top of everything else no matter what?  Or alternatively, is there some other workflow for popup notifications that StumpWM users tend to use

goose121:
This commit makes it so that rather than raising a window to the absolute top of the window stack, `raise-window` raises it above the highest-raised window in the current group which was established by `raise-window`, preventing unmanaged windows (e.g. those created by dunst) from being covered up.


goose121:
Wait hold on; I'm finding an intermittent bug where the window doesn't raise properly; I'm going to close this until I'm sure that my patch actually works

goose121:
Okay, I've fixed it now

PR #608, opened 2019-05-18
add *global-pending-top*

falood:
Thanks for the project, I tried several tiling window manager and this is the one I like best!
But when I tried to use stumpwm with [polybar](https://github.com/polybar/polybar), sometimes the polybar may be covered by the window, so I tried to add a `*global-pending-top*` variable to make a space for polybar.
I'm still learning common lisp, this pr just works but I'm not sure it's good enough as a part of the project, any suggestion is welcome and I'll try to make it better.
Also I'll add `*global-pending-bottom*` if it's good to you.
Thanks again!


dbjergaard:
Let me take a look and I’ll let you know. I hope we get better panel support so I appreciate the code.

dbjergaard:
Hi, the code looks good to merge but I remember something about reserving space for taskbars in the manual. I don’t want to add code that’s already been implemented elsewhere. Let me do some more digging once I’m on a laptop.

aricav96:
I ran some quick searches on the info manual and the closest thing to this says: 

> Alternatively, external panel applications such as the GNOME panel
> and KDE's kicker may be used.  Simply starting one of these programs is
> enough to set it as the mode line of the head it would like to be on (if
> the panel is XRandR aware) or whichever head is available.  In order to
> avoid problems displaying menus, configure your panel application for
> positioning at the top or bottom of the head rather than relying on
> *MODE-LINE-POSITION*

Granted, I did this through C-s in Emacs and used loose terms like "status bar", "mode-line" etc. (this is from part 7 of the manual, by the way), so a more thorough search might be required. However, I have also been trying to set up polybar and I did find that modifying the `update-mode-line-position` function yielded favorable results. The code as of the latest release has a comment that states that the function is meant for external mode lines too, so I figure that that must be the problem.

dbjergaard:
do we still need this PR? It looks like you found what you needed from update-mode-line-position?

aricav96:
To be clear, what I did was a hack hahaha. I simply changed the last `setf`form in the function to read `(setf (mode-line-position mode-line) :bottom)`. I don't really know which calculations are made (I postponed grokking this out of urgency) so I do think it is worthwhile to check this PR out.

dmb2:
Sorry this has languished for so long, I admit I haven’t had a chance to really dig in. I don’t want this to sit gathering dust waiting on me. One problem with merging this is the vocabulary used. pending-top doesn’t really describe to me what it’s doing. Mode-line-padding our panel-padding would be better. Also what units is this in? Making it in pixels is probably the most sensible. @PuercoPop could you take a look? Otherwise I’ll probably have some time during the upcoming holidays.

falood:
I switched to Macos recently, my new company doesn’t provide Linux laptop.
So I may won’t use stumpwm recently, feel free to close this pr.

Thanks for your awesome jobs!

Issue #610, opened 2019-05-24
Crash related to xrandr


jmbr:
The latest stumpwm crashes whenever I try to disable a screen using the xrandr command:

    xrandr --output DP-0 --off

The crash is triggered by a condition signalled by the form (xlib:rr-get-crtc-info *display* crtc (get-universal-time)) in the function output->head within head.lisp. It turns out that the crtc value returned by one of the outputs coming from (xlib:rr-get-screen-resources root), called from make-screen-randr-heads is equal to zero and this triggers the error.

PuercoPop:
Apparently writing the word issue between Close and the issue number prevents auto-closing from working :woman_shrugging:

PR #611, opened 2019-05-24
Work-around for issue #610.

jmbr:
Ditto. This prevents the crash reported in #610


PuercoPop:
@jmbr Thanks for submitting the fix. We where wondering at `#stumpwm` what does CRTC represent exactly but since it is a XID we can agree that whatever it is it shouldn't be 0. Commit with minor changes, using `PLUSP` as it is the built-in function (although technically XID must be integers)

Issue #576, opened 2019-02-15
unable to compile: RR-GET-OUTPUT-INFO not in :XLIB package


szos:
after git pull-ing StumpWM, I'm unable to recompile. All is well until running make, where it complains about XLIB::RR-GET-OUTPUT-INFO not existing in the XLIB package. The specific lisp form that is complained about is in head.lisp, on line 44, as part of a multiple-value-bind:
`(xlib::rr-get-output-info *display* output (get-universal-time))`
the debugger invocation message is:
`debugger invoked on a UIOP/LISP-BUILD:COMPILE-FILE-ERROR in thread
 #<THREAD "main thread" RUNNING {hexadecimal number}>:
       COMPILE-FILE-ERROR while compiling <CL-SOURCE-FILE "stumpwm" "head">`

I think this is related to pull request #570, as the line in question ends up getting called from the randr branch of the cond statement found on line 71 of head.lisp. 

I got it to compile by commenting out the randr branch of the cond statement alongside the functions `make-screen-randr-heads` and `output->head`. However I havent attempted to run it. I assume that it will still work, as the xinerama branch is still there and the pull request that added this functionality is recent. 
this is on Manjaro linux, with SBCL.

PuercoPop:
What version of CLX are you using?

szos:
In `quicklisp/dists/quicklisp/software/` I have `clx-20170516-git`, `clx-xembed-20120909`, and `clx-truetype-20160825`. I'm not sure how to update these, I've always just relied on `ql:quickload` to get things.

brettgilio:
Hi, I am able to reproduce this issue using ASDF

(UIOP/LISP-BUILD:CHECK-LISP-COMPILE-RESULTS NIL T T "~/asdf-action::format-action/" ((#<ASDF/LISP-ACTION:COMPILE-OP > . #<ASDF/LISP-ACTION:CL-SOURCE-FILE "stumpwm" "head">)))

 error finding frame source: Bogus form-number: the source file has probably
                               changed too much to cope with.
   source: NIL
0] 15

The regression was introduced on the first February merge. So, this definitely isn't isolated.

I am using GuixSD as my GNU/Linux system building against SBCL.

sdilts:
@szos You can update your quicklisp distributions by running `(ql:update-all-dists)` to update all of them, or individually with `(ql:update-dist dist-name)`,

[...]

PR #625, opened 2019-06-29
updated README.md with Ubuntu SBCL install instructions

M4rkD:
I'm not sure if this is useful, but here are the documented steps I took to install SBCL from source on Ubuntu. This is based on the extremely helpful advice of @PuercoPop (thanks!) in issue #576 

If you would rather it was in the wiki I can take a look the info there (or updating the Ubuntu install docs maybe).

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,9 @@ hackable desktop experience, look no further.
 * cl-ppcre
 * alexandria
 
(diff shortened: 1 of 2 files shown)


PuercoPop:
@M4rkD Thanks for the PR. I've committed a slightly reworded version of the note in the README. The specific instructions of how to build SBCL in Ubuntu belong in the wiki as you mention. Please update the Wiki instead :pray:.

PR #629, opened 2019-07-11
introduce toggle for remapped-keys

abertschi:
As a former EXWM user I really appreciated the feature to disable
remapping of keys temporarily. As far as I know stumpwm provides
`send-raw-key` to temporarily ignore remapping for a single key.

This pull request introduces a toggle behavior to temporarily disable
remapped-keys for more than a single key

## Usecase

```lisp
  (defun remap-keys-allow (win)
    (every (lambda(w) (string-not-equal w (window-class win)))
           '(list "jetbrains-idea"
             "Emacs"
             "jetbrains-pycharm"
             "jetbrains-studio"
             "android-studio2")))

  (define-remapped-keys
      `((,(lambda (win) (remap-keys-allow win))
          ("C-a"   . "Home")
          ("C-e"   . "End")
          ("C-n"   . "Down")
          ("C-p"   . "Up")
          ("C-f"   . "Right")
          ("C-b"   . "Left")
          ("M-b"   . "C-Left")
          ("M-f"   . "C-Right")
          ("M-v"   . "Prior") 
          ("M-w"   . "C-c")
          ("C-w"   . "C-x")
          ("C-y"   . "C-v")
          ("C-_"   . "C-z")
          ("M-<"   . "C-Home")
          ("M->"   . "C-End")
          ("C-M-b" . "M-Left")
          ("C-M-f" . "M-Right")
          ("C-k"   . ("S-End" "C-x"))
          ("M-a"   . ("C-Home" "C-S-End"))
          ("M-k"   . "C-k")
          ("C-d"   . "Delete"))))

  (defcommand toggle-remapped-keys () ()
              (if (eq *remap-keys-enable* t)
                  (progn
                    (setq *remap-keys-enable* nil)
                    (message "disable remapped-keys"))
                  (progn
                    (setq *remap-keys-enable* t)
                    (message "enable remapped-keys"))))

  (define-key *top-map* (kbd "s-i") "toggle-remapped-keys")

```

What do you think about this? Thanks for your awesome project :)

### DIFF
--- a/remap-keys.lisp
+++ b/remap-keys.lisp
@@ -28,7 +28,9 @@
 (export '(define-remapped-keys remap-keys-enable))
  
 (defvar *remap-keys-window-match-list* nil)
-(defvar *remap-keys-enable* t)
+
+(defvar *remap-keys-enable* t
+  "Toggle remap-keys on/off ")
 
 (defun find-remap-keys-by-window (window)
   (first
--- a/remap-keys.lisp
+++ b/remap-keys.lisp
@@ -25,9 +25,10 @@
 
 (in-package #:stumpwm)
 
-(export '(define-remapped-keys))
-
+(export '(define-remapped-keys remap-keys-enable))
+ 
 (defvar *remap-keys-window-match-list* nil)
+(defvar *remap-keys-enable* t)
 
 (defun find-remap-keys-by-window (window)
   (first
@@ -73,7 +74,9 @@
          (keys (cdr (assoc (print-key raw-key) keymap :test 'equal))))
     (when keys
       (dolist (key keys)
-        (send-fake-key window key))
+        (if (eq *remap-keys-enable* t)
+            (send-fake-key window key)
+            (send-fake-key window raw-key)))
       t)))
 
 (defun define-remapped-keys (specs)
--- a/remap-keys.lisp
+++ b/remap-keys.lisp
@@ -25,12 +25,12 @@
 
 (in-package #:stumpwm)
 
-(export '(define-remapped-keys *remap-keys-enable*))
- 
+(export '(define-remapped-keys *remapped-keys-enabled-p*))
+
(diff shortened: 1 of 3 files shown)


remap-keys.lisp
```diff
@@ -25,10 +25,13 @@
 
 (in-package #:stumpwm)
 
-(export '(define-remapped-keys))
-
+(export '(define-remapped-keys *remap-keys-enable*))
+ 
 (defvar *remap-keys-window-match-list* nil)
 
+(defvar *remap-keys-enable* t
```
PuercoPop: Could you rename this to `*remapped-keys-enabled-p*` so it aligns with `*hooks-enabled-p*`?

remap-keys.lisp
```diff
@@ -73,7 +76,9 @@
          (keys (cdr (assoc (print-key raw-key) keymap :test 'equal))))
     (when keys
       (dolist (key keys)
-        (send-fake-key window key))
+        (if (eq *remap-keys-enable* t)
```
PuercoPop: Could you move the conditional 'into' the arguments of `send-fake-key`. Also it is not necessary to `(eq t ...)`. Just use the value directly.

```
(send-fake-key window (if *remapped-keys-enabled-p*
                                           key
                                           raw-key)
```

Or bind the key value to use in an outer let, whichever you prefer

```
(let ((key (if *remapped-keys-enabled-p*
                   key
                   raw-key)))
  (send-fake-key window key)
```

remap-keys.lisp
abertschi: ok

remap-keys.lisp
abertschi: ok

abertschi:
I implemented your feedback. Let me know if there is still something to change.

dbjergaard:
I’m good, @Puercopop merge when ready!

    David

PR #634, opened 2019-08-12
Update group-raised-window properly in geometry-hints (fixes #632) 

### LINKED ISSUE #632: New windows are displayed below inactive fullscreen windows
New windows that are in a frame where there exists an inactive fullscreened window will have focus but won't be visible.

Reproduce:

In a single empty frame, launch mpv or vlc and fullscreen the window. Launch xterm in that same frame. It should display fine. When xterm is the active window, launch another xterm in that frame. Instead of the new xterm displaying, the fullscreen window will be raised although the new xterm will have focus.

This change in behavior was introduced recently by this PR by @goose121 
https://github.com/stumpwm/stumpwm/pull/607

I don't know what the proper fix is.

If the PR change can't be made to support the old behavior while still supporting dunst then I'd prefer it be reverted.

goose121:
The reason the issue occurs is because `geometry-hints` raises fullscreen windows manually rather than using `raise-window`, so the `group-raised-window` never got updated, which then confuses `raise-window`, as it raises windows to the "top" by placing them above a window which isn't actually the top window. These commits make `geometry-hints` behave similarly to `raise-window`, in that it updates `group-raised-window` and raises the fullscreen window above the old `group-raised-window` rather than to the top of the window stack.

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -119,7 +119,8 @@ than the root window's width and height."
                y (frame-y head)
                width (frame-width head)
                height (frame-height head)
-               (xlib:window-priority (window-parent win)) :above))
+               (xlib:window-priority (window-parent win)) :above
+               (group-raised-window (window-group win)) win))
        (return-from geometry-hints (values x y 0 0 width height 0 t)))
       ;; Adjust the defaults if the window is a transient_for window.
       ((find (window-type win) '(:transient :dialog))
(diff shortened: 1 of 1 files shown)


scottjad:
Some comments [here](https://github.com/stumpwm/stumpwm/issues/632#issuecomment-521786107)

PuercoPop:
Merged with some style changes in 0a104b41bbee5163f6c4deedcf06801b4ce09ad1

Thanks for your contribution!

PR #635, opened 2019-08-16
Improve default font selection

lokedhs:
The default font is set to be `9x15`. If that font can't be found, StumpWM gets a list of all fonts and picks the first one. Choosing the fist font is rarely the right thing to do, and a better default is to use `fixed`.

This change updates StumpWM to choose `fixed` in this case, and only reverts back to the first available font if it `fixed` cannot be found (which should be a very unlikely scenario).


PuercoPop:
@lokedhs Thanks for following up with the PR!

PR #636, opened 2019-08-23
Fix a crash when splitting windows in some specific cases

lokedhs:
In some cases, an off by one error caused the frame width to be negative,
resulting in a crash in CLX


PuercoPop:
Thanks :tada:

PR #640, opened 2019-09-06
allow overriding of makeinfo command

zmyrgel:
OpenBSD has old makeinfo as part of base system which errors in stumpwm info
generation. This allows to override the makeinfo command so that later version from the OpenBSD ports tree can be used instead (gmakeinfo).


alexanderbarbosa:
everything ok here! :)

PR #642, opened 2019-09-07
Add XWIN-WM-NAME

PuercoPop:
We can't use XLIB:WM-NAME as it doesn't conform to the ICCM
standard. It doesn't use the type of the property to decode the
octets.

In the case the property type is STRING there is some additional
translation needed which XLIB:CARD8->CHAR implements. Reuse that.

  STRING as a type or a target specifies the ISO Latin-1 character set
  plus the control characters TAB (octal 11) and NEWLINE (octal
  12). The spacing interpretation of TAB is context dependent. Other
  ASCII control characters are explicitly not included in STRING at
  the present time.

from https://tronche.com/gui/x/icccm/sec-2.html#s-2.7.1

REF: https://tronche.com/gui/x/icccm/sec-4.html#s-4.1.2.1

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -332,9 +332,23 @@ _NET_WM_STATE_DEMANDS_ATTENTION set"
   (when-let ((name (xlib:get-property win :_NET_WM_NAME)))
     (utf8-to-string name)))
 
+(defun xwin-wm-name (win)
+  (multiple-value-bind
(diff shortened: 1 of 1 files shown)


JMC-design:
@PuercoPop perhaps submit the fix to CLX?

Issue #641, opened 2019-09-07
StumpWM spontaneous crash in XWIN-NAME


shamazmazum:
Hello. After system update (I've updated both xorg libraries and quicklisp repo) I am constantly receiving spontaneous crashes at random time. As I lose connection to my X environment, I cannot copy-paste backtrace for you, so I attach a picture (very sorry for this):
[Backtrace photo](https://i.postimg.cc/RV3qmPLx/IMG-20190907-093609.jpg)

How can this be fixed? 

UPD: As you can see, my stumpwm is in local-projects. I use my own fork of stumpwm, which has truetype fonts patch included and supports wallpapers. It is almost up to date with upstream.

UPD2: Seems like a bug in clx, but I report here anyway, because I do not know how to reproduce it without stumpwm. Also, maybe bug appears when I have non-ASCII characters in window names? Could be. Will check it

shamazmazum:
I was right. The bug appears when there are some windows with non ASCII-characters (like russian letters) in their names. Strange, this bug has showed up only recently.

PuercoPop:
> Strange, this bug has showed up only recently.

It is likely due to the increased safety in clx's compilation options, which is a good thing.

> How can this be fixed?

I think I've spotted the issue. If I'm reading the code for `xwin-name` correctly only reading the window name from :_NET_WM_NAME is utf8 aware. Can you try modifying [xwin-name](https://github.com/stumpwm/stumpwm/blob/8c6bf7e5f9e4905dad83c7b6c4c7f3aab2a69652/window.lisp#L337) slightly?

```lisp
(defun xwin-wm-name (win)
  "Return the netwm wm name"
  (when-let ((name (xlib:get-property win :WM_NAME)))
    (utf8-to-string name)))

(defun xwin-name (win)
  (escape-caret (or (xwin-net-wm-name win)
                    (xwin-wm-name win)
                    "")))
```

let me know if it works

shamazmazum:
@PuercoPop, Hi.

Reporting. StumpWM no longer crashes, but decodes some window titles (which are decoded by xwin-wm-name, obviously) incorrectly.

For example it returns "С????к ??амб??ка в XI Ча??ов - Снежн??й ????д" instead of "Стук Бамбука в XI Часов - Снежный Мёд".

This string converted by function (lambda (str) (map 'vector #'char-code str)):

[...]

PR #643, opened 2019-09-08
Catch TYPE-ERROR when decoding window names.

shamazmazum:
Look at explanation [here](https://github.com/stumpwm/stumpwm/issues/641). To put it short: uxterm on FreeBSD puts UTF-8 encoded string in property WM_NAME of type STRING. This causes StumpWM to crash while decoding this string.

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -332,6 +332,11 @@ _NET_WM_STATE_DEMANDS_ATTENTION set"
   (when-let ((name (xlib:get-property win :_NET_WM_NAME)))
     (utf8-to-string name)))
 
+(defun safely-decode-x11-string (string)
+  (handler-case
+      (map 'string 'xlib:card8->char string)
+    (type-error () nil)))
+
 (defun xwin-wm-name (win)
   (multiple-value-bind
         (name encoding)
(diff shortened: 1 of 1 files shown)


PuercoPop:
@shamazmazum Sorry about the brain fart, indeed we need to handle the case when clients mistakenly put utf-8 strings w/o setting the appropriate property type.

Instead of making the handler case cover more forms could you please define a function named safely-decode-string that wraps the only the first branch of the if in and handles the type error?

shamazmazum:
@PuercoPop Is it ok now? I've moved handler cases into two separate functions (I think, the other wants its own function too ;))

shamazmazum:
Also, is StumpWM supported on other common lisp implementations or only on SBCL? Is it correct to use SBCL-specific functions? Flexi-streams should provide platform-independent octets-to-string

PuercoPop:
> (I think, the other wants its own function too ;)

I think just calling `utf8-to-string` would be better. Less code, less problems (Sorry im asking you to remove code I just wrote :shame:

> Is StumpWM supported [...] only on SBCL?

Yes

> Is it correct to use SBCL-specific functions? 

Yes

shamazmazum:
I am sorry. Pushed wrong code. This is obviously an error:

>(if (eq encoding :x11-string)
          (safely-decode-x11-string name)
          (sb-ext:octets-to-string string :external-format encoding)))

Already fixed it in my branch but it was too late

shamazmazum:
@PuercoPop Can you please fix it? Or I need to open another pull request? The line

> (sb-ext:octets-to-string string :external-format encoding)

needs to be replaced with

> (utf8-to-string name)

Issue #647, opened 2019-09-12
StumpWM crashes frequently when switching monitors


sjl:
Sorry for this long, rambling issue, but I don't really understand what's happening and/or know how to debug it further.

I use StumpWM on two different machines:

* A desktop PC that's always attached to a single monitor.
* A laptop that is sometimes used on its own, and other times used with an external monitor.

The desktop setup is fine.  The laptop, however, has gotten extremely flaky when switching between the laptop monitor and the external monitor.  I have two commands set up in my `stumpwmrc` to run the necessary `xrandr` commands to switch between the setups:

```lisp
(defcommand screen-laptop () ()
  (loop with laptop = "eDP1"
        with extern = "DP1"
        for (output commands) in `((,laptop ("--auto"))
                                   (,laptop ("--primary"))
                                   (,extern ("--off")))
        do (uiop:run-program `("xrandr" "--output" ,output ,@commands))))

(defcommand screen-external () ()
  (loop with laptop = "eDP1"
        with extern = "DP1"
        for (output commands) in `((,extern ("--auto"))
                                   (,extern ("--primary"))
                                   (,laptop ("--off")))
        do (uiop:run-program `("xrandr" "--output" ,output ,@commands))))
```

These used to work fine, but last week I updated StumpWM and SBCL and now running these commands seems to crash StumpWM pretty often (I'd say ~50% chance of crashing).  I was previously on StumpWM commit 6115981 and am now on commit ae97bf9 and I'm using SBCL 1.5.6.

Here are some photos of the kinds of errors I get (the `copy-unhandled-error` command mentioned in the text box doesn't seem to work, so phone pictures are all I've got, sorry).  First, the error I usually see is this:

![Photo Sep 11, 13 38 06](https://user-images.githubusercontent.com/58365/64815755-490ca980-d574-11e9-9df1-b5650de808ff.jpg)

I tried sniffing around in the code and hacking in a fix:

```diff
diff --git a/tile-group.lisp b/tile-group.lisp
index eeea233..296b8cd 100644
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -128,9 +128,10 @@
 (defmethod group-lost-focus ((group tile-group))
   ;; If this window had the focus, try to avoid losing it.
   (let ((frame (tile-group-current-frame group)))
-    (setf (frame-window frame)
-          (first (remove-if 'window-hidden-p (frame-windows group frame))))
-    (focus-frame group frame)))
+    (when frame
[...]

PR #651, opened 2019-10-12
Adds a config to conditionally enable RANDR in make-screen-heads

kriyative:
Workaround for #647


dmb2:
Please simply fix the bug with a comment that this is upstream to us in clx, and (bonus points) include a link to the issue opened with clx so we can track when it gets fixed in clx and remove the comment.

PR #652, opened 2019-10-31
WHERE-IS: Style update

PuercoPop:


dmb2:
It would be nice if we captured a unit test demonstrating this is not a breaking change as we do these style updates to the codebase. Well never get 100% coverage but at least we’ll have fixed points for the code we’ve touched. Just a thought.

PR #665, opened 2019-12-01
place window in closest frame when unfloating

szos:
changes unfloat-window to utilize the new function calculate-appropriate-frame-for-window, which returns the frame closest to the window sent in. also adds two helper functions: calculate-window-centerpoint and calculate-frame-centerpoint.

This modifies the current behaviore of unfloat-window, which currently places an unfloated window into the first frame of the group.

### DIFF
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -1358,7 +1358,7 @@ frame, then returns the frame with the shortest distance. "
       (cdr shortest))))
 
 (defun unfloat-window (window group)
-  (let ((frame (calculate-appropriate-frame-for-window window)))
+  (let ((frame (calculate-appropriate-frame-for-window window group)))
     (change-class window 'tile-window :frame frame)
     (setf (window-frame window) frame
           (frame-window frame) window
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -1315,9 +1315,50 @@ direction. The following are valid directions:
         (message "There's only one frame.")
         (balance-frames-internal group tree))))
 
+(defun calculate-window-centerpoint (win)
+  "window x and y are calculated from the top of the screen. thus the x/y are the upper left hand 
+corner. to find the center x/y we do: xc = x + (w ÷ 2) and yc = y + (h ÷ 2)"
+  (let ((x (window-x win))
+        (y (window-y win))
+        (w (window-width win))
+        (h (window-height win)))
+    (let ((center-x (+ x (/ w 2)))
+          (center-y (+ y (/ h 2))))
+      (list center-x center-y))))
+
+(defun calculate-frame-centerpoint (frame)
+  "Calculates the centerpoint of a frame, not including borders. "
+  (let ((x (frame-x frame))
+        (y (frame-y frame))
+        (w (frame-width frame))
+        (h (frame-height frame)))
+    (let ((center-x (+ x (/ w 2)))
+          (center-y (+ y (/ h 2))))
+      (list center-x center-y))))
+
+(defun calculate-appropriate-frame-for-window (&optional (win (current-window))
+							  (group (current-group)))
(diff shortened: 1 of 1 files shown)


tile-group.lisp
```diff
@@ -1315,9 +1315,50 @@ direction. The following are valid directions:
...
+  (let ((x (window-x win))
+        (y (window-y win))
+        (w (window-width win))
+        (h (window-height win)))
+    (let ((center-x (+ x (/ w 2)))
+          (center-y (+ y (/ h 2))))
+      (list center-x center-y))))
+
+(defun calculate-frame-centerpoint (frame)
+  "Calculates the centerpoint of a frame, not including borders. "
```
PuercoPop: s/Calculates/Returns/

tile-group.lisp
```diff
@@ -1315,9 +1315,50 @@ direction. The following are valid directions:
...
+  (let ((x (frame-x frame))
+        (y (frame-y frame))
+        (w (frame-width frame))
+        (h (frame-height frame)))
+    (let ((center-x (+ x (/ w 2)))
+          (center-y (+ y (/ h 2))))
+      (list center-x center-y))))
+
+(defun calculate-appropriate-frame-for-window (&optional (win (current-window))
```
PuercoPop: Instead of the generic word appropriate it is better to say what is the specific criteria we are using for selecting the frame. In this case it is the `closest-frame`.

There is no reason for the arguments to be `&optional`.

tile-group.lisp
```diff
@@ -1315,9 +1315,50 @@ direction. The following are valid directions:
...
+							  (group (current-group)))
+  "Takes a window and calculates the distance from its centerpoint to the centerpoint of every 
+frame, then returns the frame with the shortest distance. "
+  (flet ((square (n) (* n n)))
+    (let* ((window-centerpoint (calculate-window-centerpoint win))
```
PuercoPop: Use `destructuring-bind` to bind the window centerpoint points.

tile-group.lisp
```diff
@@ -1315,9 +1315,50 @@ direction. The following are valid directions:
...
+           (winx (car window-centerpoint))
+           (winy (cadr window-centerpoint))
+           (shortest nil))
+      (loop for frame in (group-frames group)
+         do (let* ((frame-alist (calculate-frame-centerpoint frame))
```
PuercoPop: Move the bindings from the `let*` form into `for/=` clauses. An additional benefit is that we can use destructuring to simplify the bindings.

szos:
Wow, loop can do way more than I thought! The final look you posted looks a lot cleaner; I had forgotten all about destructuring bind after spending the last while in scheme world. 

Im not very sure about whether we should hide the iteration like that - I generally prefer recursion to iteration, it just makes more sense to my brain. I think i prefer the first version - the second one is difficult for me to parse, and when running it, the function continually returns nil instead of the frame. If I understand correctly, we initialize `frame` `frame-x` `frame-y` `distance` and `closest`, and via the `then` clause providing logic to set `closest` every iteration - either to the new distance/frame or the old one. However I fail to see where we are returning the frame, sorry im not to familiar with loop beyond a basic (loop for x in y do z). 

Ill look over the rest of the comments tomorrow and implement the changes; its past midnight here and this will be best approached tomorrow. 

Thanks for taking a look over and showing me some good tips/syle/loop info. Cheers

szos:
Unfortunately, I have found no way to add a commit to this pull request. Perhaps im missing something completley obvious, but I have deleted and reforked my stumpwm repo, and now will open a new pull request with the suggested changes added. Sorry for making things more complicated than they need to be.

PuercoPop:
No problem, btw an alternative option is to send a patch to the mailing list if you don't want to deal with GitHub for any reason.

PR #667, opened 2019-12-03
Place window in closest frame when unfloating

szos:
Here is the new  pull request. as the old one (#665) wouldnt let me add new commits for some reason. apologies again.


PuercoPop:
Merged with minor stylistic changes. I think the reason why you couldn't add new commits to the other PR is because you had committed them to your master branch.

Thanks for working on this!

PR #670, opened 2019-12-14
Refactor keysyms code. Store keysym docstrings and deprecation data.

equwal:
I would like to be able to auto-generate documentation about keybindings, but I need that information to be accessible. Specifically, if a key is unbound then I want to return some information about it. So `C-i h k g` should show show something like
`g is a unicode character named LATIN_G_OR_SOMETHING with keycode #xFAD`.
There were about 1000 comments in `keysyms.lisp`, so I included them in a struct. This is why the changed line count is so large. Having these like this makes it possible to warn about deprecated keys, and to use the docstrings in key documentation messages.

Doing this required some changes to the code, so I tried to uncomplicate anything I encountered.

 There are three commits here with long messages that might be helpful to read.
- Keysyms are a struct, with optional documentation and deprecated slots. This could be useful, for example, to store the new preferred key in the `deprecated` slot so it can be suggested somehow. Also, it removed the need for custom accessors, and less hash tables.
- More conditions for `parse-key` to fail with, including reporters. The reporter is now a little more helpful in showing why things broke.
- Option of wrapping your keys with `define-keys` instead of repeated calls to `define-key`. I think this is cleaner.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -757,7 +757,7 @@ input (pressing Return), nil otherwise."
 
 (defun get-modifier-map ()
   (labels ((find-mod (mod codes)
-             (let* ((keysym (keysym-name->keysym mod))
+             (let* ((keysym (keysym-code (get-keysym mod)))
                     (keycodes (multiple-value-list (xlib:keysym->keycodes *display* keysym))))
                (intersection keycodes codes))))
     (let ((modifiers (make-modifiers)))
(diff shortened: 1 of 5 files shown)


equwal:
This also touches on #479, since a few conditions have been made and possibly submitted by `get-keysym` and handled by `parse-key`

equwal:
In order to get this to pass that test, I really should do a separate pull request for the `parse-key` condition system changes first.

equwal:
I'm going to close this and reopen with only the `parse-key` edits in here, no `defstruct` or anything.

PR #671, opened 2019-12-14
Make describe-key show more helpful information

equwal:
Try: C-t h k C-i ? which will pop out: 
"? shows the bindings for the prefix map under C-t."

Previously describing a key would only show what command it ran, but now it provides the docstring also:
https://drive.google.com/file/d/1ibdQmd2BXBThWqPAhh_-SXNc5AjHo-rQ/view

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -73,3 +73,4 @@ Herbert Jones       jones dot herbert at gmail.com
 Daniel Oliveira     drdo at drdo.eu
 Panji Kusuma        epanji at gmail dot com
 Andrin Bertschi     hi at abertschi dot ch
+Spenser Max Truex   web ate spensertruex.com
--- a/help.lisp
+++ b/help.lisp
@@ -70,6 +70,18 @@
     (message-no-timeout "~{~a~^~%~}"
                         (columnize data cols))))
 
+(defun final-key-p (keys class)
+  "Determine if the key is a memeber of a class"
+  (member (lastcar keys) (mapcar #'parse-key class) :test #'equalp))
+
+(defun help-key-p (keys)
+  "If the key is for the help command."
+  (final-key-p keys '("?" "C-h")))
+
+(defun cancel-key-p (keys)
+  "If a key is the cancelling key binding."
+  (final-key-p keys '("C-g")))
+
 (defcommand describe-key (keys) ((:key-seq "Describe Key: "))
   "Either interactively type the key sequence or supply it as text. This
   command prints the command bound to the specified key sequence."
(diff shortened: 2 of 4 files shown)


dmb2:
Can you please show a demo of how this will work for very long doc-strings? I'm not a format guru, but it doesn't look like the format strings will word-wrap doc-strings longer than 80 columns, how hard would that be to implement?

equwal:
> Can you please show a demo of how this will work for very long doc-strings? 

Introducing: `long-doc`
```
(defcommand long-doc () ()
           "This command is to demonstrate a very long docstring without any newlines whatsoever. As you can see, it is quite long and needs wrapping! Just in case you have an ultra-wide monitor, these sentences are repeated a few times. This command is to demonstrate a very long docstring without any newlines whatsoever. As you can see, it is quite long and needs wrapping! Just in case you have an ultra-wide monitor, these sentences are repeated a few times. This command is to demonstrate a very long docstring without any newlines whatsoever. As you can see, it is quite long and needs wrapping! Just in case you have an ultra-wide monitor, these sentences are repeated a few times. This command is to demonstrate a very long docstring without any newlines whatsoever. As you can see, it is quite long and needs wrapping! Just in case you have an ultra-wide monitor, these sentences are repeated a few times."
           t)
I'm not a format guru, but it doesn't look like the format strings will word-wrap doc-strings longer than 80 columns,

Indeed, it does not do that. I also would like things ending in `C-g` to return something other than `not bound`.

> how hard would that be to implement?

Hopefully not too hard. I'll give it a go.

**Comment by equwal** [comment]
I also would like things ending in `C-g` to return something other than `not bound`.

**Comment by dmb2** [comment]
Follow emacs's convention for this please.  I'm very supportive of including more context in the help messages, and emacs provides and excellent template.

**Comment by dmb2** [comment]
I forget where I saw it tonight (there's been a ton of comments to go through) but I'm opposed to printing superfluous information about a key if it isn't bound.  I don't want C-t h k f print out information about the character 'f'.  Anything besides 'C-g' should respond that it is not bound.  C-g needs special treatment, and better than its getting now so I'm open to changes there.

**Comment by equwal** [comment]
> I forget where I saw it tonight (there's been a ton of comments to go through) but I'm opposed to printing superfluous information about a key if it isn't bound. I don't want C-t h k f print out information about the character 'f'. Anything besides 'C-g' should respond that it is not bound. C-g needs special treatment, and better than its getting now so I'm open to changes there.

I know where you saw it. [Here it is!](https://github.com/stumpwm/stumpwm/pull/670) After doing that work I'm inclined to agree, so I closed that pull.

**Comment by equwal** [comment]
> Follow emacs's convention for this please. I'm very supportive of including more context in the help messages, and emacs provides and excellent template.

That is going to require some major edits. Here is an emacs message:
```
C-f runs the command evil-scroll-page-down, which is an interactive compiled
Lisp function in ‘evil-commands.el’.

It is bound to C-f, <visual-state> C-f, <normal-state> C-f.

(evil-scroll-page-down COUNT)

Scrolls the window COUNT pages downwards.
```

So here one must implement:
- finding the file where a key is bound at
- finding all bindings for a given command
- the lambda-list for a given command
- show the docstring

**Comment by equwal** [comment]
> how hard would that be to implement?

This hard:
```common-lisp
(defmacro wrap (words max-col &optional stream)
  "Word wrap at the max-col"
  ;; Format insanity edited from Gene Michael Stover's "Advanced Use of Lisp's
  ;; FORMAT Function (2004)"
  `(format ,stream ,(concatenate 'string "~{~<~%~1,"
                             (with-output-to-string (s) (princ max-col s) s)
                             ":;~A~> ~}")
           ,words))

(wrap (split-string "This command is to demonstrate a very long docstring without any newlines whatsoever." " ") 80 *standard-output*)
```

equwal:
@dmb2 I believe this is ready for another review. Here is a screenshot:

https://spensertruex.com/static/new-documentation.png

I works for describe-key and describe-command.

equwal:
@dmb2 Here is `long-doc` with the max-columns set at `120` as an example:
https://spensertruex.com/static/long-doc.png

equwal:
This is the current version I have: https://spensertruex.com/static/stumpwm-describe-pretty.png

I'm open to other considerations. The idea is to only show the `lambda-list` when is is not `'()`, and to emphasize the arguments (not the function's name). This is similar to how emacs does it with italics.

equwal:
This has the added benefit of making `C-t h w` `where-is` redundant. It could be removed or aliased to `C-t h c` if that is desired.

equwal:
The latest push reduces redundancy a bit by removing the first line [Here's a screenshot of it](https://spensertruex.com/static/describe-command-short.png).

dmb2:
This looks really awesome!

equwal:
Thanks for merging all of these @dmb2.

Issue #675, opened 2019-12-17
Command "quit" does not quit everything.


equwal:
Running `C-t : quit` only closes the mode line, but does not totally close everything. For this, we need `(sb-ext:exit)` or similar.

equwal:
This is caused by my mode line which uses some `contribs`. Not sure which one yet.

PR #676, opened 2019-12-17
Add quit keybinding, ask to quit, and use sb-ext on SBCL (#675)

equwal:
(sb-ext:exit) provides a sure-fire way to quit everything on SBCL.

Running `C-t : quit` only closes the mode line, but does not totally close everything. Also, adding the keybinding `C-t q` is helpful, since many users of StumpWM cannot use it as their only window manager yet. Since it is easy to accidentally hit it (thereby closing all your windows and possibly losing a lot of work) it asks a one-character prompt: `Really close StumpWM? (y or n)`.

### DIFF
--- a/bindings.lisp
+++ b/bindings.lisp
@@ -183,6 +183,7 @@ is a tile group.")
   (kbd "F")       "curframe"
   (kbd "-")       "fclear"
   (kbd "Q")       "only"
+  (kbd "q")       "quit-confirm"
   (kbd "Up")      "move-focus up"
   (kbd "Down")    "move-focus down"
   (kbd "Left")    "move-focus left"
--- a/user.lisp
+++ b/user.lisp
@@ -217,6 +217,8 @@ such a case, kill the shell command to resume StumpWM."
 
 (defcommand-alias abort keyboard-quit)
 
+(defcommand quit-confirm (&optional (confirm t)) ((:y-or-n "Really close StumpWM? "))
+  (when confirm (quit)))
 
 (defcommand quit () ()
 "Quit StumpWM."


bindings.lisp
```diff
@@ -181,6 +181,7 @@ is a tile group.")
   (kbd "f")       "fselect"
   (kbd "F")       "curframe"
   (kbd "-")       "fclear"
+  (kbd "q")       "quit"
```
dmb2: I'm not sure I'm convinced binding quit by default is a good idea.

user.lisp
```diff
@@ -218,10 +218,14 @@ such a case, kill the shell command to resume StumpWM."
...
 (defcommand-alias abort keyboard-quit)
 
+(defvar *quit-prompt* "Really close StumpWM? "
+  "The prompt asked before quitting StumpWM.")
 
 (defcommand quit () ()
-"Quit StumpWM."
-  (throw :top-level :quit))
+  "Quit StumpWM."
+  (when (y-or-n-p *quit-prompt*)
```
dmb2: I don't mind the yes or no prompt.  I rarely suggest doing this, but I would like the ability to configure this behavior.  When I want stumpwm to quit, ':quit' means quit, I don't want to deal with another prompt.  Could you make a global called *prompt-on-quit* and set it to nil by default and then set this to (and *prompt-on-quit* (y-or-n-p *quit-prompt*))

user.lisp
```diff
@@ -218,10 +218,14 @@ such a case, kill the shell command to resume StumpWM."
...
+    #+sbcl (sb-ext:exit)
+    #-sbcl (throw :top-level :quit)))
```
dmb2: We don't support anything besides sbcl, so you can get rid of the sbcl hints.  I don't like the use of (sb-ext:exit) because this is the posix extension which implements the 'exit()' call from C which is "quit now, no backsies, sorry if you were in the middle of saving something we're dead now bye".  The way this is currently handled is that it throws a top-level condition that allows stumpwm the opportunity to exit gracefully.  

If you've got a  vanilla mode-line that is causing stumpwm not to exit cleanly then that's a bug, otherwise it sounds like you might have something configured that is causing stumpwm to misbehave.

bindings.lisp
equwal: I understand this concern, which is why it also is set to prompt by default. Otherwise a user poking around in StumpWM could accidentally close everything and have no idea what happened.

user.lisp
equwal: Would it be better to define a special function for interactive use (along with the variable setting) so that non-interactive funcitons can call `quit` and be sure it works, but a user can call `quitsafe` (or something) to be able to have/configure the prompt?

bindings.lisp
equwal: I also find myself having to quit StumpWM fairly often, because stuff breaks and configuration involves testing things.

user.lisp
dmb2: Well documented variables the user can tweak from their config file is more maintainable from my perspective than keeping multiple functions around that do almost the same thing.

user.lisp
equwal: It looks like the mode line configuration (yet again!) has caused stumpwm to misbehave.t

jorams:
I must say I rarely use the quit command, so binding it by default does not seem particularly useful to me. When I do want to use it, the confirmation prompt would only be annoying.

That said, I do think it might be a good addition for new users, because it would display a way to quit in the help popup. The default startup message tells users to press `C-t ?`, and immediately providing a way to quit makes sense.

Perhaps we should instead add a command (perhaps `maybe-quit`, or `quit-after-confirmation`) that gets bound by default? That way the behaviour of the `quit` command remains consistent and clear, while offering a more forgiving option for keybindings.

equwal:
> I must say I rarely use the quit command, so binding it by default does not seem particularly useful to me. When I do want to use it, the confirmation prompt would only be annoying.
> 
> That said, I do think it might be a good addition for new users, because it would display a way to quit in the help popup. The default startup message tells users to press `C-t ?`, and immediately providing a way to quit makes sense.
> 
> Perhaps we should instead add a command (perhaps `maybe-quit`, or `quit-after-confirmation`) that gets bound by default? That way the behaviour of the `quit` command remains consistent and clear, while offering a more forgiving option for keybindings.

@jorams 
I can get behind this. The new push adds `quit-confirm` to `C-t q` using stumpwm's `:y-or-n` argument parser,

Issue #673, opened 2019-12-17
Clipboard integration issues with xclip and vim.


equwal:
Neither xclip nor vim seem to have access to my clipboard, which tends to be annoying.

`echo "Not in StumpWM clipboard :(" | xclip`
`xclip -o` => `Not in StumpWM clipboard`.
`vim` (try to paste something from the StumpWM clipboard with `p`.

So both these programs seem to be lacking the integration I'd like. I still have a clipboard in other programs. Emacs, for example, works fine. So I find myself doing things like this:
```bash
some command >~/.tmp # cannot use xclip
emacs ~/.tmp # yank it in emacs and kill
```

kriyative:
@equwal - you might need to specify the X selection - xclip defaults to XA_PRIMARY, most apps seem to use XA_CLIPBOARD. See if the following works better:
```
echo "cut" | xclip -selection clipboard
```
and try pasting in vim.

equwal:
> @equwal - you might need to specify the X selection - xclip defaults to XA_PRIMARY, most apps seem to use XA_CLIPBOARD. See if the following works better:
> 
> ```
> echo "cut" | xclip -selection clipboard
> ```
> 
> and try pasting in vim.

I see. This indeed works. The functions `getsel` and `putsel` also exist, but are pretty much useless.

equwal:
The xclip suggestions works in general, but vim is still storing it's kill ring elsewhere.  I don't think this is a problem with StumpWM, so I am closing.

kriyative:
BTW, `putsel` and `getsel` also default to the `primary` selection, and that could be changed. Also, the primary selection can be accessed using the middle mouse button paste in most apps. It's just that most paste keybindings refer to the `clipboard` selection

equwal:
> BTW, `putsel` and `getsel` also default to the `primary` selection, and that could be changed. Also, the primary selection can be accessed using the middle mouse button paste in most apps. It's just that most paste keybindings refer to the `clipboard` selection

How could `getsel` and `putsel` be changed to just be generic copy/paste functions? I think if they were changed that way, and to use the `clipboard` selection, that using StumpWM's clipboard could be a lot more intuitive (eg. by putting them on `C-i y` and `C-i M-w`).

equwal:
> BTW, `putsel` and `getsel` also default to the `primary` selection, and that could be changed.

Changing this breaks the `show-clipboard-history-contrib`, though that could be changed also. My view is that it is worth doing.

[...]

PR #678, opened 2019-12-17
Make get-x-selection and set-x-selection have configuration choices for clipboard vs. primary selections #673

equwal:
One may desire to use the `clipboard` instead of the `primary` selection, or even clobber both.

Addresses https://github.com/stumpwm/stumpwm/issues/673

### DIFF
--- a/selection.lisp
+++ b/selection.lisp
@@ -29,9 +29,13 @@
           *default-selections*))
 
 (defvar *default-selections* '(:primary)
-  #.(format nil "~@{~A~}"
-          "Either a list of keyword X selections (like :primary or :clipboard)"
-          ", or a single one."))
+  #.(format
+     nil "~@{~A~%~}"
+     "A keyword or list, one of:"
+     ":primary or '(:primary) uses only the \"primary\" selection"
+     ":clipboard or '(:clipboard) uses only the \"clipboard\" selection"
+     "Both can be specified in a list like '(:primary :clipboard). In this case,"
+     "set-x-selection will clobber both, and get-x-selection will default to the first item."))
 
 (defun export-selection (selection)
   (let* ((screen (current-screen))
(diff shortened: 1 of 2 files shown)


xsteve:
What most applications expect depends on the applications you use.
My experience is that established unix tools use the primary selection.
Only few applications that I use use clipboard.
So I would love to keep primary the default.

One thing that could be done is adding a variable that defaults to primary - you can set it to clipboard.

equwal:
> What most applications expect depends on the applications you use.
> My experience is that established unix tools use the primary selection.
> Only few applications that I use use clipboard.
> So I would love to keep primary the default.
> 
> One thing that could be done is adding a variable that defaults to primary - you can set it to clipboard.

Is there a reason why we should not clobber both? I can do that also.

dmb2:
Hi, I see you've made a number of really great contributions recently.  One thing I try hard to do (which constrains what makes it into the main stumpwm branch) is to avoid opinionated defaults.  In this case, the current behavior is as close to the standard as we can be.  At times its inconvenient an not intuitive, but its the best we've got.

I like the idea of making this configurable.  Could you set the default to the old behavior and add the new variables to the manual?

equwal:
> Hi, I see you've made a number of really great contributions recently. One thing I try hard to do (which constrains what makes it into the main stumpwm branch) is to avoid opinionated defaults. In this case, the current behavior is as close to the standard as we can be. 

>At times its inconvenient an not intuitive, but its the best we've got.
> I like the idea of making this configurable. Could you set the default to the old behavior and add the new variables to the manual?

This is a fair point. Arbitrarily clobbering two selection spaces by default is kind of strange. On the other hand, I wonder at where my selections are going sometimes.

selection.lisp
```diff
@@ -25,7 +25,13 @@
...
 
 (export '(get-x-selection
-          set-x-selection))
+          set-x-selection
+          *default-selections*))
+
+(defvar *default-selections* '(:primary)
+  #.(format nil "~@{~A~}"
+          "Either a list of keyword X selections (like :primary or :clipboard)"
+          ", or a single one."))
```
dmb2: Please document all possible values and what they do.

selection.lisp
```diff
@@ -102,24 +106,25 @@
...
-          (xlib:convert-selection selection :utf8_string (screen-input-window (current-screen)) :stumpwm-selection)
-          ;; Note: this may spend longer than timeout in this loop but it will eventually return.
-          (let ((time (get-internal-real-time)))
-            (loop for ret = (xlib:process-event *display* :handler #'wait-for-selection :timeout timeout :discard-p nil)
-                  when (or ret
-                           (> (/ (- time (get-internal-real-time)) internal-time-units-per-second)
-                              timeout))
-                  ;; make sure we return a string
-                  return (or ret "")))))))
+  (let ((selection (if (listp selection) (car selection) selection)))
```
equwal: #L109 is the only non-whitespace change.

dmb2:
Good to go once you fix merge conflicts.

PR #680, opened 2019-12-18
Remove trailing whitespaces from everything.

equwal:
Like the great sage once said

> [So a language that makes source code ugly is maddening to an exacting programmer, as clay full of lumps would be to a sculptor](http://www.paulgraham.com/pypar.html)

And although the README does say 
> [do not] change whitespace

I think it is worth it in this case, as it is all that is done here.


dmb2:
It says in the readme don’t do this. There’s a reason.

PR #681, opened 2019-12-18
bind and document *movement-map*

equwal:
This symbol was exported, but not bound or documented. I think it is best to try to make use of it at some point.

### DIFF
--- a/bindings.lisp
+++ b/bindings.lisp
@@ -26,7 +26,6 @@
 
 (export '(*groups-map*
           *group-top-maps*
-          *movement-map*
           *help-map*
           set-prefix-key))
 
--- a/dynamic-window.lisp
+++ b/dynamic-window.lisp
@@ -37,7 +37,9 @@ area."
   (declare (ignore win))
   (let* ((windows (group-windows group))
          (num-win (length windows)))
-    (only)
+    ;; Only try to make this the only frame if it isn't already.
+    (unless (only-one-frame-p)
+      (only))
     (recursive-tile (min *expose-n-max* num-win) group)))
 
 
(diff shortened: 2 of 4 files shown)


equwal:
I believe there are two conflicting statements here:
> It looks like movement map was made whole-cloth for this PR? Also I don’t see it being bound by default.

This was the _bug_ that I was _fixing_. I did not invent `*movement-map*`, it was already there.

> I general our defaults are fixed. Yes there’s commands that are very useful that aren’t bound but I don’t want to come up with bindings just because.

I didn't put it there, but I documented it since it was exported.

> I’m open to suggestions on how to make this stuff more discoverable, but remember: one of the goals is to get the window manager out of the way of your work.

dmb2:
I’m only looking at the diffs on github, I only have time to look at source on my laptop once or twice a week. According to the diff, the defvar was added in this pr and didn’t exist before. I also don’t see a default binding. 

Documentation is welcome and encouraged, new default bindings are a different animal.

jorams:
The symbol `*movement-map*` was exported from `bindings.lisp` before this PR, but is otherwise entirely non-existent. Perhaps a better approach is to just stop exporting the unused symbol?

equwal:
> The symbol `*movement-map*` was exported from `bindings.lisp` before this PR, but is otherwise entirely non-existent. Perhaps a better approach is to just stop exporting the unused symbol?

Either option is better than exporting a symbol that does not otherwise exist.

dmb2:
Let’s delete it entirely.

equwal:
> Let’s delete it entirely.
> 
> Edit: assuming some contrib doesn’t reference it for a good reason.

None of the published ones do, according to the almighty `grep`. The new commit just removes it and should be ready @dmb2.

jorams:
Merged in 382be2c4bcc105b9c12af936a359bdf52e5effa1.

The commit contained some unrelated changes.

dmb2:
Thanks @jorams

PR #682, opened 2019-12-18
&&& CONDITION and +++ GENERIC-FN manual markup character sets

equwal:
I wanted to have `&&& CONDITION` in the manual. To do this I wrote the `defdoc` macro. This required writing `manual.lisp`, but I think it is worth it since it reduces lots of code duplication. All that needs to be done to add a new documentation form is adding a `defdoc` for it (~5 lines) and nothing else.

I tried to make the history as clean as possible for review purposes.

- Whitespace is not meaningful on line items, so `@@@func` is the same as `@@@      func` now.
- print progress on `*debug-io*`, so generating the documentation is less prone to error.
- Added a new documentation section "Conditions and restarts" which uses the new `&&&` line item.
- `+++ generate` is used to document itself.
- Added new documentation section `17.2 Adding Documentation and Editing This Manual` which gives some basic information about how the `texi` is generated, and how to write new documentation with the macro-expanders.

### DIFF
--- a/manual.lisp
+++ b/manual.lisp
@@ -24,7 +24,9 @@
 
 (in-package #:stumpwm)
 
-(require :sb-introspect)
+(require :sb-introspect) ; Function lambda lists.
+;; FIXME: Why does SBCL not know how to REQUIRE sb-mop?
+;; (require :sb-mop); Class slots lists.
 
 (defvar *debug-doc* nil
   "When T, will print extra debugging information from the doc generator.")
@@ -49,7 +51,7 @@
   (:documentation "Generate a texi.in documentation line."))
 
 (defmacro defdoc ((specializer
-                   (out-stream-var line-var name-var)
+                   (out-stream-var line-var name-var symbol-var)
                    marker dprint-label)
                   &body body)
   "Define a document generating method."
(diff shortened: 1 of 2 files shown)


dmb2:
Hmmm. I’m not a huge fan of refactoring the doc functions into a macro. 

I’m glad you documented how the manual is generated, so that can stay, I don’t want the new conditional stuff. Do you have to have +++ as well?

equwal:
> Hmmm. I’m not a huge fan of refactoring the doc functions into a macro.

I think you should reconsider. If you would like I can refactor them into a more conservative macro than the one currently there. There is a lot of repeated code that does not need to be there if no macro is used.

> I’m glad you documented how the manual is generated, so that can stay, I don’t want the new conditional stuff. Do you have to have +++ as well?

Not really, but there is no way to get the documentation for `defmethod` definitions. So defining it as a `defun` would be misleading. There weren't any generic functions in the source before this pull, so there previously was no want for `+++`.

equwal:
> I’m glad you documented how the manual is generated, so that can stay, I don’t want the new conditional stuff.

equwal:
I'm splitting this up into smaller pull requests so you can more easily accept/reject things one at a time.

dmb2:
@PuercoPop what do you think?

equwal:
A better macro would allow definitions like this:
`(defdoc ("%%%" "defmac" function) "{~a} ~{~a~^ ~}~%~a")`
or similar one-line definitions.

equwal:
> @PuercoPop what do you think?

I'm glad you mentioned him, since he has [this project](https://github.com/stumpwm/stumpwm/projects/2) which is suggestive of this kind of change.

PuercoPop:
First @equwal let me thank you for taking to time and effort to work on improving the Manual.

> Hmmm. I’m not a huge fan of refactoring the doc functions into a macro.
> @PuercoPop what do you think?

I agree with @dmb2 in this case, as the code is a ~100 LoC 'shell script'. There isn't much boilerplate to be shaved in this case. I would only go for it if the macro was super straight forward. I also have some objections to how the implementation itself which I'll list later.

> I'm glad you mentioned him, since he has this project which is suggestive of this kind of chang

I think you meant to link this instead https://github.com/stumpwm/stumpwm/commit/a481f2e19007752208b9e9262ded3e463853a414. Although the goal there was mainly to signal a warning when a symbol is not found instead of signaling an error which has affected our *BSD users in the past.

The current scope of this PR is too large, I think the best way move forward is to break into smaller PRs with a single purpose so we can reach consensus on a point by point basis. As it stands the PR is doing

- [ ] Adds support for conditions (Using the `&&&` prefix).
- [ ] Adds support for generic functions (Using the `+++` prefix).
- [ ] Refactor the fall-through 'case' code so that it uses a generic function instead
- [ ] Reduces boilerplate by introducing the defdoc macro
- [ ] Some miscellaneous cleanups. 

I'm in favor of adding support for conditions to the manual

Unless I'm missing something I don't understand why we would want to give special treatment to generic functions so I would be against adding that prefix.

I can see using a generic function leading to more readable code, although for a 'closed' system (not expected to be extended by 3rd parties) I think the approach in a481f2e19007752208b9e9262ded3e463853a414 would lead to simpler and more straightforward code. If you plan on using this as a base for a general documentation system then yeah, a generic function + syntax sugar on top is the way to go. Minor nit, the name generate is too generic.

Finally in the case of defmacro I think the best argument in favour of it would be a single patch showing the LoC it saves, but I'm guessing it won't be many. Or how much logic it centralizes to keep things DRY. But it appears it only takes care of variable binding according to a ppcre, as stands it looks like a specialized variant of lambda-match.

Finally the code seems to be building the regular expressions by concatenating strings. That barbarism would be fine if this was Perl, but in Lisp we have more civilized ways. CL-PPCRE has [create-scanner](http://edicl.github.io/cl-ppcre/#create-scanner2) which lets you build the regexp using s-exps, similar to emacs's rx.el

equwal:
>     * [ ]  Adds support for conditions (Using the `&&&` prefix).

I'd like to see more proper use of conditions in StumpWM, but I don't want to
make a mess of things. So the best thing to do is probably to:
- add this `&&&` sigil.
- Move conditions to a file called `conditions.lisp`.
- make a manual section specifically for conditions.

>     * [X]  Some miscellaneous cleanups.

I think this one has been covered by https://github.com/stumpwm/stumpwm/pull/684 

Although for the third I think a better plan is to use a hook with lambda-functions (no generics). So make it:

* [ ] Refactor doc generators to be functions in a hook.

I'll not do these, since I've been convinced that it is a fools errand:
>     * [ ]  Reduces boilerplate by introducing the defdoc macro
>     * [ ]  Adds support for generic functions (Using the `+++` prefix).

Thanks for helping sort these things out. I'll close this now and reference each of these points when I have a pull for them.

PR #684, opened 2019-12-19
Helpful output while generating the documentation (dprint).

equwal:
This implements `dprint` (which was already defined, but with the implementation commented out). It sends everything to debug-io, and does a freshline afterward (which looks nice, since otherwise it says `Formatting manual for the var *top-level-error-action*...stumpwm.texi`. 

Since this is only ran at compile-time, there is not much of a use case for turning it off.

Was a part of (closed) https://github.com/stumpwm/stumpwm/pull/682

### DIFF
--- a/manual.lisp
+++ b/manual.lisp
@@ -26,11 +26,9 @@
 
 (require :sb-introspect)
 
-;; handy for figuring out which symbol is borking the documentation
 (defun dprint (type sym)
-  (declare (ignorable type sym))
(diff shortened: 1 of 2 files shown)


equwal:
I imagine the rationale behind closing this is as follows @dmb2:

> I will not be accepting any modifications to how the doc-strings are currently generated.

dmb2:
No I closed this one because the function is vestigial and should probably be deleted rather than uncommented.

PR #685, opened 2019-12-19
Whitespace not meaningful in documentation macroexpander

equwal:
Makes the macros in the doc not depend on whitespace, so these lines are all the same:
```
@@@ format @c the old way
@@@format @c no space required
@@@      format @c any whitespace accepeted
```
If desired, it is possible to issue a `STYLE-WARNING` when there is anything but a single space.

Was a part of https://github.com/stumpwm/stumpwm/pull/682

### DIFF
--- a/manual.lisp
+++ b/manual.lisp
@@ -33,7 +33,7 @@
   )
 
 (defun generate-function-doc (s line)
-  (ppcre:register-groups-bind (name) ("^@@@ (.*)" line)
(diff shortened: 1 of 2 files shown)


dmb2:
Being sensitive to white space enforced consistent style. This makes blending. Two styles possible which I don’t want to encourage. I don’t want to relax something that is already enforced when we already have a “don’t allow style warnings” guideline.

PR #686, opened 2019-12-19
No warning when calling expose

equwal:
Fix bug where the `expose` command warns "There's only one frame." This can be
distracting, since there are likely to be multiple frames.

Now it only warns when `expose` is called on a single window.

# Expose Bug Reproduction:
- Open a few terminal emulators
- `C-t Q` (only)
- `C-t ; expose`
- See warning on the top-right corner.


equwal:
I don't think this is the best approach to solving this, and it still warns `there is only one frame` after "exposing" the only current frame and making the user select it.

Issue #687, opened 2019-12-20
`expose` displays message "There's only one frame." when it should not.


equwal:
`expose` calls `(only)` twice. First it calls `(only)` so that it can start splitting it. If the frame was already maximized then the message is displayed. Then it calls `fselect`, which can be annoying if accidentally calling `expose` when there was only one frame open (since the user must select that frame -- even if it is the only one). Finally, it calls `(only)` again, potentially displaying the warning message again.

# Expose Bug Reproduction:
- Close all applications except for one. `C-t Q` to make it the only frame.
- `C-t ; expose`
- See warning on the top-right corner.
- wait for it to timeout.
- be forced to select the only window on the screen.
- see the warning again.

# Expected behaviour
- Close all applications except for one.  `C-t Q` to make it the only frame.
- `C-t ; expose`
- See the warning. No `fselect`, just stop trying to expose.

PR #688, opened 2019-12-20
Don't expose with only one application or only one frame

equwal:
`expose` made two unconditional calls to the "only" command, which results in
spurious messages "There's only one frame." Also, it called "fselect" on only
one frame, forcing the user to select the only frame.

The new function `only-one-frame-p` is defined and wraps these to ensure the user
is only shown `fselect` or the error messages from `only` when it makes sense.

Fixes the bug in issue #687


dmb2:
Fix merge conflicts and this is ready.

PR #695, opened 2019-12-22
Use a hook to expand the documentation sigils %%%, ###, etc.

equwal:
Push anonymous functions onto a hook, then expand it by executing them. This approach has the benefit of reducing code duplication using StumpWM's own `hooks` implementation. I've decided this `defdoc` macro is straigtforward enough; it is about as complicated as `defun`.

### DIFF
--- a/manual.lisp
+++ b/manual.lisp
@@ -32,61 +32,69 @@
   ;(format t "~&Doing ~a ~a..." type sym)
   )
 
-(defun generate-function-doc (s line)
-  (ppcre:register-groups-bind (name) ("^@@@ (.*)" line)
-                              (dprint 'func name)
-                              (let ((fn (if (find #\( name :test 'char=)
-                                            ;; handle (setf <symbol>) functions
-                                            (with-standard-io-syntax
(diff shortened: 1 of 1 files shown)


dmb2:
My original and continued opposition to this is that it takes the DRY (don't repeat yourself) idiom to the extreme.  The current/original implementation of these doc functions are all structured the same.  They scream for DRY to be applied, BUT there are only a handful of doc-string "sigils" and having a redundant implementation for them isn't a big deal.  Once you read one, you understand the rest, and we don't need to modify/maintain them because they have all proved semantically correct.  

Doing anything to refactor them adds intellectual overhead to the maintainer reading the code, and does nothing to make the code more maintainable since you aren't squashing bugs or making bugs less likely. The code paths are straightforward and well tested as is.  

I will not be accepting any modifications to how the doc-strings are currently generated.

dmb2:
@PuercoPop Sorry I missed your comments in the other thread.  I didn't realize you suggested breaking things out into smaller PRs.  If you feel this should have been merged the way it was addressed please reopen and merge (or bug me and I'll re-open and merge it).

PR #696, opened 2019-12-22
Add generate-condition-doc `&&&`

equwal:
Adds `&&&` and documents it in the manual. Does not actually document any conditions, just implements the functionality.

[Here is an image showing the condition `stumpwm-error`](https://spensertruex.com/static/condition-example.png). Note that `stumpwm-error` has no slots, so none are shown (but it is implemented).

Need this merged to write a docs section for conditions, thanks.

### DIFF
--- a/manual.lisp
+++ b/manual.lisp
@@ -88,6 +88,16 @@
         t)
       (warn "Symbol ~A not found in package STUMPWM" name))))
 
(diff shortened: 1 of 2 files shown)


dmb2:
This type of documentation is out of scope for the manual.  It is only relevant to people reading source code, and those people have the technical competency to understand source-code comments.

Issue #697, opened 2019-12-22
gselect 1, 2, 3... should make a new group if there is not one there already


equwal:
Trying to open one of these groups returns a message "No Such Group."

I propose a condition `no-such-group` being defined, which can be used in this exceptional situation to be handled as `gnew "workspace N"` instead.

In fact there are a lot of [calls to `throw 'error`](https://github.com/stumpwm/stumpwm/search?q=throw+error&unscoped_q=throw+error&type=Code) that could possibly be made useful.

Also, these prompt the user for a group, which makes no sense. They should just open the group at the number without prompting. Try: `C-t <F1>` so see the prompt.

Also, `gmove` and should do the same thing.

dmb2:
No, this changes default behavior.  This is a case where your expected behavior isn't how it works, but it isn't clearly what everyone might expect.  Opening/selecting a non-existant group should return said message.

Opening a new group should prompt the user for the group name unless it was supplied from a non-interactive call (this is currently correctly handled).

PR #698, opened 2019-12-22
Move conditions to the file conditions.lisp

equwal:
* Makes it easier to use conditions.
* Conditions *must* be declared before use, so they can be made the
first file loaded.

Having things this way will make https://github.com/stumpwm/stumpwm/issues/697 easier.


dmb2:
Conditions are contextually dependent on what is happening around them. Refactoring them to their own file breaks this context.

PR #704, opened 2019-12-23
Use color to emphasise the confirm-quit prompt

equwal:
The rationale is that it could be *really bad* to quit accidentally.
Modeled on the `gkill` color scheme when closing a non-empty group. [Here is an image of this prompt](https://spensertruex.com/static/really-quit-or-die.png).

I'm against flashy colours in most cases, but this one is special.

### DIFF
--- a/user.lisp
+++ b/user.lisp
@@ -217,8 +217,12 @@ such a case, kill the shell command to resume StumpWM."
 
 (defcommand-alias abort keyboard-quit)
 
-(defcommand quit-confirm (&optional (confirm t)) ((:y-or-n "Really close StumpWM? "))
-  (when confirm (quit)))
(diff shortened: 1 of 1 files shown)


user.lisp
```diff
@@ -217,8 +217,12 @@ such a case, kill the shell command to resume StumpWM."
 
 (defcommand-alias abort keyboard-quit)
 
-(defcommand quit-confirm (&optional (confirm t)) ((:y-or-n "Really close StumpWM? "))
-  (when confirm (quit)))
+(defcommand quit-confirm () ()
+  (when (y-or-n-p #.(format nil "~@{~a~^~%~}"
```
PuercoPop: I don't think this warrants read time evaluation.

PR #707, opened 2019-12-23
Improve describe-function and variable with pretty emacs-like output

equwal:
* Using SBCL's `describe` for this purpose has some problems, and is much uglier than this. For example, it is possible to color the function/variable names, control width and height, remove unhelpful information (for a StumpWM user) and add an elipsis to the current value if it is too large.
* describe-var shows the value of the variable, but limits it to `*help-max-height*` lines, when it makes an elipsis (see screenshot).

Mandatory screenshots:
- [describe-function](https://spensertruex.com/static/describe-function.png)
- [describe-variable](https://spensertruex.com/static/describe-variable.png)

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -357,7 +357,7 @@ then describes the symbol."
   (multiple-value-bind (sym pkg var)
       (lookup-symbol (argument-pop-or-read input prompt))
     (if (fboundp sym)
-        (symbol-function sym)
+        sym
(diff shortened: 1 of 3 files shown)


equwal:
Exporting `*message-max-width` and `*help-max-height*` seems like the wrong way to do this. Instead, this pull should detect the screen size somehow.

PR #708, opened 2019-12-25
Debug prompts for describe-* commands

equwal:
Remove extra spaces, and extra colon ':' from describe-key.


dmb2:
Do you have a good stylistic justification for this? If not, I prefer to keep things as they are.

equwal:
> Do you have a good stylistic justification for this? If not, I prefer to keep things as they are.

Yes, it is preferable to have the same spacing every time from the prompt. I think that it should be one space only.

PR #709, opened 2019-12-25
Change `commands` function to fuzzy search on 3 columns

equwal:
This makes `describe-command` and `commands` the same, except `commands` shows
more commands at once.

This makes `commands` a useful function. I felt inclined to allow selecting multiple commands somehow, then calling `describe-command` on all of them, but I think the UI for such a thing would be too complicated and clunky.

Alternatives to this:
- Alias `commands` to `describe-command`, abandoning the three column layout.
- Alias `describe-command` to `commands`, abandoning the single column layout.
-  Just remove `commands` entirely. If this is the choice, I think it would be a nice compromise to make `describe-command` use a two-column layout.
- I don't think leaving `commands` as-is is a very good alternative, since it is mostly useless as-is.

[screenshot](https://spensertruex.com/static/commands.png)

### DIFF
--- a/help.lisp
+++ b/help.lisp
@@ -64,14 +64,22 @@
                         (or (columnize data cols) '("(EMPTY MAP)")))))
 
 (defcommand commands () ()
-"List all available commands."
+  "List all available commands in multiple columns.
+Runs describe-command on the selected command."
   (let* ((screen (current-screen))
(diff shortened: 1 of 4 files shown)


equwal:
I think using a two-column version and aliasing is a better plan.

PR #710, opened 2019-12-25
Change `commands` and `describe-command` to fuzzy search on 2 cols

equwal:
This makes `describe-command` and `commands` the same function, and
`commands` more useful.
[screenshot](https://spensertruex.com/static/commands-2col.png)
Since `columns` is now a menu-entry, it should be easier to use this for changing the `C-t ?` display to also use completion (since it is a 4 column thing).


equwal:
Should really use a proper column guessing mechanism instead.

PR #711, opened 2019-12-25
Remove `commands` since `describe-command` does that already

equwal:
`describe-command` has fuzzy completion of command names, so a command that does
a dumb display of all the commands available is not required.

### DIFF
--- a/help.lisp
+++ b/help.lisp
@@ -63,16 +63,6 @@
                         (print-key-seq key-seq)
                         (or (columnize data cols) '("(EMPTY MAP)")))))
 
-(defcommand commands () ()
-"List all available commands."
-  (let* ((screen (current-screen))
-         (data (all-commands))
-         (cols (ceiling (length data)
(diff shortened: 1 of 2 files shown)


PuercoPop:
I like the idea of removing obsolete commands but we can't break backwards compatibility.

@dmb2 wdyt about declaring commands as an alias to describe-command and adding a TODO comment to remove in 2 years time?

equwal:
> I like the idea of removing obsolete commands but we can't break backwards compatibility.
> 
> @dmb2 wdyt about declaring commands as an alias to describe-command and adding a TODO comment to remove in 2 years time?

This might be a good time to adopt some kind of deprecation mechanism, if that is desirable. `all-commands` already exists for non-interactive purposes.

dmb2:
@PuercoPop I agree, just make sure the TODO has a future or current date so that maintainers know when to remove the redundant code.

PR #712, opened 2019-12-26
Additional columns for describe functions

equwal:
Adds a column to `colon`(with lambda-lists) and `describe-command` (with a keybinding list). 

This doesn't affect the completion, since a new function for completion is used. Thus, they are only there to make it easier to use the system.
# Screenshots:
- [describe-command with keybindings](https://spensertruex.com/static/desc-command-kb.png)
- [colon with lambda-lists](https://spensertruex.com/static/describe-cmd-2.png)

It should be straightforward to add new informational columns to other commands using `add-column`.

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -593,13 +593,28 @@ delimiting them with double quotes. A backslash can be used to escape
 double quotes or backslashes inside the string. This does not apply to
 commands taking :REST or :SHELL type arguments."
   (let* ((commands (all-commands))
-         (cmd (select-from-menu (current-screen)
-                                (mapcar #'list commands)
-                                ": "
-                                (or (position initial-input
-                                              commands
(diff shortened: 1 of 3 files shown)


dmb2:
its not clear from the screenshots if the user can select a possible completion.  Is this possible? Also, it looks like this PR breaks the build as is.

equwal:
> its not clear from the screenshots if the user can select a possible completion. Is this possible? Also, it looks like this PR breaks the build as is.

Yes, the user can select the completion. The completion for the command is separated at the first space.

dmb2:
Fix merge conflicts and I'll merge this.

equwal:
Unfortunately this goes against the old completion system. 

If I can add that completion as an alternative in `input.lisp`, then I will be able to do something like this.

Issue #713, opened 2019-12-27
^-t ; command box broken or changed behaviour


kaarejens:
It seems that "ctrl-t - ;" has changed behaviour after recent commit. 

Both history (arrow up) and tab completion is now broken, while the new "list-commands-as-you-type" is the new behaviour. Also, ex. "run-shell-command urxvt" does not work, so now one must type "run-shell-command" hit enter, and then type the options. 

Not sure if this is on purpose, or if it is a bug, but at least for me, this is not what I expect.

This commit seems to be related: https://github.com/stumpwm/stumpwm/commit/41f0f70eaeec76e982d6dc5663ca73abcb44a267

mtlll:
Perhaps it would be better to keep the old colon and add a new command colon-menu with the new behaviour. While this new behaviour is nice in theory, I personally never type a command into colon unless I already know how to type it, and often typing and tab completing is faster than selecting from the list.

For instance, now if I wanted to run 'next' using colon, even if I typed all of "next", I still have to press down 3 times to get past fnext, gnext, and gnext-with-window. The muscle memory reactiong of just typing next and pressing enter would run fnext. Additionally, pressing tab causes garbage characters to be input instead of completing. Perhaps SELECT-FROM-MENU should be changed to tab-complete on tab, because that's quite jarring for people who habitually hit tab to complete, like me.

kaarejens:
My thoughts exactly. Though it would of course be nice to have a hybrid - best of both worlds, but until someone can produce such a code, it seems more convenient to put this on ex. ^-t  ,  or some other meaningful combo for now. 

I really like the idea of this, but if I would have to choose, I would say the old way, was quicker and more valuable. It also feels more aligned with the WM in general (remembering lots of key combos for efficiency and so on).

PuercoPop:
> just realized the suggested ^-t , only makes sence with nordic keyboard layouts

Don't worry I think it was clear you meant <prefix>-t. 

I think it is reasonable to support the old functionality while still 'moving forward'. What do y'all think about restoring the old functionality and instead adding a new command _but_ updating the default binding for ^-t to the new command?

cc: @dmb2

emacsomancer:
@PuercoPop I think restoring the old functionality and adding a new command is good. I don't have a firm opinion on the binding.

equwal:
[...]

PR #714, opened 2019-12-30
Add input line completion with fuzzy complete

equwal:
* Set the input line for scrolling one element at a time.
* Tab-complete on the top item.
* safe-car in describe-command (bugfix).
* Add deletion command C-DEL to clear the prompt line.
#713

The intended way to use this is to type a partial command, move to a completion with `C-n` and `C-p`, using `C-TAB` to complete and `RET` to just run it.

### DIFF
--- a/menu-definitions.lisp
+++ b/menu-definitions.lisp
@@ -32,6 +32,9 @@
           select-from-batch-menu
           command-menu))
 
+(defun car-safe (lst)
+  (if (consp lst) (car lst) lst))
+
 (defun entries-from-nested-list (lst)
   (mapcar (lambda (x)
             (make-instance 'menu-entry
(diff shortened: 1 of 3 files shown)


menu-definitions.lisp
```diff
@@ -32,6 +32,9 @@
           select-from-batch-menu
           command-menu))
 
+(defun car-safe (lst)
+  (if (consp lst) (car lst) lst))
+
```
dmb2: Please do not use car-safe.  There's a reason why this isn't built into the language.

menu-definitions.lisp
```diff
@@ -216,6 +226,12 @@ unmark the entry at the selected point."
   "Default action is to do nothing"
   (declare (ignore key-seq)))
 
+(defun match-p (table-item menu)
+  (funcall (single-menu-filter-pred menu)
+           (car table-item)
+           (second table-item)
+           (single-menu-current-input menu)))
+
```
dmb2: Why did you pull this out of the labels form? Please don't break methods that are internal to other methods. It pollutes the local namespace.

equwal:
Closing this pull, since it builds on the completion system I was asked to revert [here](https://github.com/stumpwm/stumpwm/issues/713).

PR #715, opened 2019-12-30
Completion sorting by length and first characters

equwal:
Sort menu items by length, and start the regex search with `^` for a more intuitive experience.

Addresses #713

### DIFF
--- a/menu-definitions.lisp
+++ b/menu-definitions.lisp
@@ -32,6 +32,12 @@
           select-from-batch-menu
           command-menu))
 
+(defun menu-sort-dwim (tbl-items)
+  "Sort the table in order to Do What I Mean."
+  (sort tbl-items
+        (lambda (tbl1 tbl2)
+          (< (length (first tbl1)) (length (first tbl2))))))
+
 (defun entries-from-nested-list (lst)
   (mapcar (lambda (x)
             (make-instance 'menu-entry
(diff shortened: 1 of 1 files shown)


dmb2:
I'm struggling to understand the context of some of your pull requests.  They do not contain sufficient justification for why the changes are necessary and do not lay out a good case for why the changes should be made.

What is intuitive to you may not be intuitive to everyone else, especially when we've grown accustomed to how StumpWM behaves as it is currently configured. 

I do not want any "do what I mean" functions in stumpwm proper because they reflect the author's idea of what they mean.  This type of behavior is perfectly in-scope for a contrib module.  If you need help getting one of the ground let me know, and we can export whatever symbols you need as long as they're reasonable parts of the API to crystallize.

equwal:
> I'm struggling to understand the context of some of your pull requests. They do not contain sufficient justification for why the changes are necessary and do not lay out a good case for why the changes should be made.
> 
> What is intuitive to you may not be intuitive to everyone else, especially when we've grown accustomed to how StumpWM behaves as it is currently configured.
> 
> I do not want any "do what I mean" functions in stumpwm proper because they reflect the author's idea of what they mean. This type of behavior is perfectly in-scope for a contrib module. If you need help getting one of the ground let me know, and we can export whatever symbols you need as long as they're reasonable parts of the API to crystallize.

This added some sorting suggested in the issue, which was intended as the context. Since I am reverting the use of `select-from-menu` in the command input, there is no need for this pull at this time.

dmb2:
Since a lot of your recent pull requests have to do with revamping the completion system, would you like to put this into a project with some defined milestones that we can brainstorm so that we don't have this kind of mis-communication? That way we can get your ideas merged in without having to revert/break default behavior.

equwal:
> Since a lot of your recent pull requests have to do with revamping the completion system, would you like to put this into a project with some defined milestones that we can brainstorm so that we don't have this kind of mis-communication? That way we can get your ideas merged in without having to revert/break default behavior.

That would be great.

PR #716, opened 2019-12-30
Add tab-in-indent to .gitattributes

equwal:
Prevents a very annoying whitespace problem.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.lisp diff=lisp
+*.lisp whitespace=tab-in-indent,trailing-space,space-before-tab


dmb2:
Again, you are not providing any justification for why the changes are needed, or what the changes actually do.

equwal:
Thanks for looking at this. Since modifying whitespace is against the guidelines, it is a good idea to have a stricter whitespace policy for git (which is what this does).

dmb2:
Explain what the new options do and why they're needed, ideally citing the corresponding documentation from git, or referencing other lisp projects that use them.

PuercoPop:
@equwal as @dmb2 says giving further context about why the change is needed and what it does goes a long way in making it easier for the PR to be understood and either merged or give clear feedback. For example you mention this `a stricter whitespace policy for git (which is what this does)` but fail to mention what that policy is. Is the new policy that git will error when trying to commit a diff with trailing whitespace? That trailing whitespace will be highlighted git displays a diff?

Below is an example of how the initial submission could have been written. Note that specific headers titlted Problem Solution are not necessary but it is important that they are always clearly stated. 

<hr />

### Problem

People of all git proficiency levels submit changes to PR which results in patches with trailing whitespace, tabs or a combination of both. As we discourage whitespace changes in our patches it is important to catch this problems early on.

### Solution

Configure git to prevent committing patches with whitespace problems (I would expand here but I'm unsure what this PR does :grimacing: )

From [`man 1 git-config`](https://git-scm.com/docs/git-config)

- `tab-in-indent`: treats a tab character in the initial indent part of the line as an error (not enabled by default).

- `trailing-space`: is a short-hand to cover both blank-at-eol and blank-at-eof.

- `space-before-tab`: treats a space character that appears immediately before a tab character in the initial indent part of the line as an error (enabled by default).

<hr /> 

As I mentioned above I'm not entirely sure what this PR does. Is setting `core.whitespace` enough or do we also what to set `apply.whitespace` to fix? What about `diff.whitespace`? Is the default behaviour to error? To fix? What will happen if one tries to commit changes with trailing whitespaces after merging this?

Hope this helps you improve your future sumbmissions!

equwal:
Thanks for the very in-depth reply @PuercoPop. The [gitattributes `whitespace` value](https://git-scm.com/docs/gitattributes#_checking_whitespace_errors) uses the same syntax as [`core.whitespace` in git-config](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corewhitespace). It does this to whitespace detection:

> git diff will use color.diff.whitespace to highlight them, and git apply --whitespace=error will consider them as errors.

I advocate detecting as many as possible to prevent whitespace errors from sneaking in undetected and becoming permanent.

Either way it is a minor change for sure.

dmb2:
I'm not opposed to this change if it can be clearly articulated what it is doing and why.  @PuercoPop's response is a great example of how to format this info.  You don't need to strictly follow this format, but doing so may organize your PR statements in a way that would make it much easier for me to assess them.

I don't want to make any changes that impose my opinions on other people, and I don't want to make changes that make it more difficult to contribute.

The no-whitespace change rule is a soft rule, if white space changes in the course of editing a file, and its easy to see past them in the diff during code review I don't care.  I don't want specific PRs whose only purpose is white space changes, nor do I want it to be a general rule that each person can impose their idea of whitespace when they submit a PR.  For me, its a judgement call when I review a PR and for the most part it has worked well in the past.

PR #718, opened 2019-12-31
Show completions with input function

equwal:
Adds a bare-minimum display of possible completions for input. The input line works the same as before, but displays possible completions rather than keeping them invisible. History, tab completion, etc. are not affected.

[screenshot](https://spensertruex.com/static/input-complete.png)

Since it does not do regex search, but instead just shows the completions in order, it won't interfere with "muscle memory" or any such thing. Just makes it easier to spell things correctly and learn what is available.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -342,6 +342,7 @@ match with an element of the completions."
                                     (match-input))
                                 (return (input-line-string input))
                                 (draw-input-bucket screen prompt input "[No match]" t)))))))
+      (draw-input-bucket screen prompt input)
       (setup-input-window screen prompt input)
       (catch :abort
         (unwind-protect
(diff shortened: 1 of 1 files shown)


impaktor:
I discoverd this feature just now when upgrading my stump. I expected it to work like the minibuffer in Emacs, where one can C-n/C-p to select in the list (at least if using [swiper/councel](https://github.com/abo-abo/swiper), or maybe Helm?), I don't know if that's do-able with the current stumpwm code base, but would be a nice additional feature to have, if anyone is looking to do more work on this.

equwal:
> I discoverd this feature just now when upgrading my stump. I expected it to work like the minibuffer in Emacs, where one can C-n/C-p to select in the list (at least if using [swiper/councel](https://github.com/abo-abo/swiper), or maybe Helm?), I don't know if that's do-able with the current stumpwm code base, but would be a nice additional feature to have, if anyone is looking to do more work on this.

For historical reasons we are not doing that. There were recent commits that did that, but we were asked to revert those.

Issue #722, opened 2020-01-10
Build error after commit 2f073d8


Jehops:
Here is a full build log: http://pkg.awarnach.mathstat.dal.ca/data/12amd64-default/2020-01-10_11h22m35s/logs/errors/stumpwm-19.11.73_1.log

I believe the error started after upgrading SBCL.  If the issue is not obvious, over the weekend I can try testing different SBCL versions and/or dissecting StumpWM commits.

PuercoPop:
Hey thanks again for. The error is not particularly obvious. It seems similar to the previous one, where there was a function that was only defined for Linux using the `#+linux` feature expression, but I checked the manual to see if readline-from-sysfs had crept into it once again but it doesn't appear to be the case.

It would be great if you could run a git bisect to find out the offending commit. This would be the 3rd time this year we have broken the build on OpenBSD so I think we should take proactive measures to avoid doing so in the future.

I've heard good things about the OpenBSD support of builds.sr.ht so we should setup the CI to build every PR on OpenBSD as well as Linux. I have a [build manifest for ubuntu](https://builds.sr.ht/~puercopop/job/135249) working. If you could provide me with the sequence of shell commands you use to build StumpWM I can write a build manifest for OpenBSD.

@dmb2 Are you ok with us using builds.sr.ht?

Jehops:
I'll try to dig deeper this weekend.  I am using the _FreeBSD_ port to build it, but I can also try quicklisp and different SBCL versions to try and get to the bottom of the problem.

PuercoPop:
Ok, lets do FreeBSD then. I run into some issues installing quicklisp on the sbcl from ports. Any tips? https://builds.sr.ht/~puercopop/job/135275

Jehops:
It seems commit 7fc3def introduces the build failure.  The previous commit, 747b6c3 builds fine with SBCL 2.0.0.  I can also build the latest commit (currently bf0b06a) with `stumpwm.texi.in` reverted to 747b6c3: http://pkg.awarnach.mathstat.dal.ca/data/12amd64-default/2020-01-12_11h49m01s/logs/stumpwm-19.11.75.log

@PuercoPop, hooking CI in to one or both of the BSDs would be much appreciated.  I mention FreeBSD, because I think OpenBSD and FreeBSD could be sufficiently different.  I can really only offer information about FreeBSD, since I'm only a rare OpenBSD user.

[...]

PR #728, opened 2020-01-15
Removing Linux-only function reference from stumpwm.texi.in

Jehops:
Since the read-line-from-sysfs function is 1) a temporary workaround for
GNU/Linux and 2) causing build problems outside of GNU/Linux, remove it from
stumpwm.texi.in. This solves issue #722.

Function documentation:

   On GNU/Linux some contribs use sysfs to figure out useful info for the
   user. SBCL up to at least 1.0.16 (but probably much later) has a problem
   handling files in sysfs caused by SBCL's slightly unusual handling of files
   in general and Linux' sysfs violating POSIX. When this situation is resolved,
   this function may be removed.

### DIFF
--- a/stumpwm.texi.in
+++ b/stumpwm.texi.in
@@ -2153,7 +2153,6 @@ multiplexer to actually run it.
 !!! getsel
 !!! putsel
 !!! copy-unhandled-error
-@@@ read-line-from-sysfs
 @@@ define-stumpwm-command
 !!! set-contrib-dir
 


PuercoPop:
Merged in https://github.com/stumpwm/stumpwm/commit/92692a621d36206636d01bdb0d4eaffd89262859 with an additional modification so that the build doesn't fail the next time since this is the 3rd time read-line-from-sysfs slips back in. Now the build won't fail when the function is not found but instead a warning will be signaled.

PR #734, opened 2020-01-24
Suppress displaying completions of "" and note truncated completion list

sjl:
This patch does two (possibly controversial) things:

1. Suppress showing *all* possible completions when the user hasn't typed anything at all yet.
2. If the completion list is truncated to fit `*maximum-completions*`, replace the last completion with `... and N more`.

I added item 1 is because I find it very disconcerting to immediately have a giant list of words immediately thrown at me when I open a prompt.  It *is* cut down to fit `*maximum-completions*`, but that means what I end up seeing is the first N alphabetical items every single time, which doesn't seem super helpful.  This patch delays showing the list of completions until the user has at least typed *something*.

I added item 2 because if I get a list of completions that looks roughly like it's around the maximum length I've set, I can't be sure whether I'm seeing *all* my options or whether there are some I'm not seeing because they were truncated off.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -42,10 +42,12 @@
 ;;; General Utilities
 
 (defun take (n list)
-  "Returns a list with the first n elements of the given list."
-  (loop repeat n
-        for x in list
-        collect x))
+  "Returns a list with the first n elements of the given list, and the
+remaining tail of the list as a second value."
+  (loop for l on list
+        repeat n
+        collect (car l) into result
+        finally (return (values result l))))
 
 ;; This could use a much more efficient algorithm.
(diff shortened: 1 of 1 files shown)


PuercoPop:
Thanks for this :tada: . I don't think your changes are controversial, they are polishing the current implementation of the recent feature, which although it has its rough edges I think it is a step in the right direction.

This PR brings something else to light. We should start thinking be moving a way of using linked lists for completions to at least a dequeue, maybe even a trie. f/e `(append (butlast completions) (list ..))` is n² as butlast also has to traverse the whole list. Because n maximum value is `*max-completions*` (100 by default) we don't run into problems but it will comeback to bite us eventually.

Issue #747, opened 2020-03-09
Listing exec completions is too slow


brandoninvergo:
I'm not sure exactly when the feature was added, but now when running the `exec` command, when typing an executable's name a big list of possible matches pops up, narrowing down as you type more.  The problem is that the initial generation of this list is too slow and while it is loading, the next keypress is ignored.

So, as an example, I'll run `exec` and start typing "firefox". After typing "f", it loads a big list of all executables that start with "f". Meanwhile, I'm continuing typing in "irefox". That "i" is missed, though, so by the time I'm done typing, the prompt says "frefox". Incidentally, if focus is on a window with text input, the "i" is actually entered in that window's text box instead of in StumpWM's input box.

equwal:
It is straightforward to replace this with historical completion only from this session.

equwal:
Also, I think this is a duplicate of #741.

brandoninvergo:
Ah very sorry, I somehow completely missed #741 (I swear, I did check before sending the bug report). Apologies for the noise.

PuercoPop:
A well written bug report is not noise, even if it was a duplicate. Thanks for taking the time to submit it. Closing in favor of 741

Issue #741, opened 2020-02-23
New default `exec` UI lags after the first char, discards the following chars


madand:
## Problem
The new "enhanced" UI for the `exec` command consistently exhibits the following annoying behavior. It discards the second (and third if you type fast enough) typed character, leaving you with only the first typed character and long list of candidates.

## Steps to reproduce
1. Open the `exec` widget with e.g. `Ct !`.
2. In the opened input widget rapidly type some 2- or 3-character prefix for some shell command. For example you can type `em` for Emacs.

## Expected behavior
Have all the character typed present in the input widget.

## Observed behavior
Second character typed is not present, due to a lag caused by filtering the list of candidates. Actually the candidates are cached the first time a prefix is used, but lag is still sufficient enough to cause character "loss" even when the cached candidates list is re-used.

## Proposed solution
Make the new and fancy stuff **optional** and **opt-in**.

PuercoPop:
I just tried to reproduce and the observed behaviour is worse. The second character is not lost, it is sent to the currently selected window. Try to reproduce but before opening the exec widget have your cursor in the text area at the bottom of the issue and you'll see!

You can also reproduce the issue with other text inputs like `C-t ;`. This points to the underlying issue being  a bug in the code that handles that handles the input window.

madand:
I confirm the "lost" chars are indeed sent to the active window.

equwal:
#749 is a workaround for this.

m-g-r:
> I just tried to reproduce and the observed behaviour is worse. The second character is not lost, it is sent to the currently selected window.

Thanks for pointing this out! I noticed that the StumpWM input handling is too slow for me typing on the two machines where I use it but haven't noticed so far that the input goes somewhere else. Maybe that explains some spurious characters in some files I noticed at "commit time".

PR #749, opened 2020-03-09
Shell completion workaround *verbose-shell-completion* toggle

equwal:
It makes sense to only do historical completion, unless StumpWM was going to
implement full shell completion support.

Affects all `:shell` commands, like `run-shell-command` or its alias `exec`.

Addresses issue #747

>Proposed solution
>Make the new and fancy stuff optional and opt-in

# Completion 
Completion with `*verbose-shell-completion* t`.
shell-progs

With `*verbose-shell-completion nil` (history only)
shell-hist

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -33,8 +33,12 @@
           defcommand-alias
           define-stumpwm-type
           run-commands
+          *verbose-shell-completion*
           %interactivep%))
 
+(defvar *verbose-shell-completion* nil
+  "When T, show very verbose shell completion. Known bug: Can cause problems
+losing character events.")
 (defstruct command-alias
   from to)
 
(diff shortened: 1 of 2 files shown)


parjanya:
I just found out about this completion to `exec`, and I like it just as it is. Thanks a lot for that! Adding an option for history only would be very nice, but I don’t regard the current behaviour as an error.

equwal:
> I just found out about this completion to `exec`, and I like it just as it is. Thanks a lot for that! Adding an option for history only would be very nice, but I don’t regard the current behaviour as an error.

That is a good point. I don't have the problems stated in #741, probably because my computer is fast enough to handle the completion without missing character events.

A solution is to add an opt-in `*verbose-shell-completion*` variable so that you can enable it, but it won't hinder users desiring a more minimal choice.

dmb2:
Hi, please don't refer to issue numbers in the docstrings.  Also, this is not a fix to issue 741, at best its a workaround.  We need to understand the underlying reason why input is being dropped/ignored, and if this the the only possible solution (making a user-configurable switch to shut off buggy behavior) then we should use it as a last resort.

PuercoPop:
> That is a good point. I don't have the problems stated in #741, probably because my computer is fast enough to handle the completion without missing character events.

I don't think it is an issue of computer speed but of typing speed. Try typing the characters almost at the same moment.

PR #752, opened 2020-03-10
Where-is alias keybindings support

equwal:
This builds on the following PR: #750 which is tagged 19.11-doc-v1
whereas this PR is tagged 19.11-doc-v2

Where-is is made more useful by showing aliased names and keys

Displays the aliases to a command, even if it is not itself bound to any keys.
This makes for a much more useful where-is message for tracking down the
relationship between a particular action and a key.

This makes for a more verbose message, but in my experience I usually wanted
that extra information, since that is why I did a where-is in the first place.
In most cases it makes no difference (since most commands are not aliased), and
it normally just adds a few lines, and the messages are still small and
non-obtrusive.

# Old behaviour
where-is-old

# New behaviour

where-is-new

### DIFF
--- a/help.lisp
+++ b/help.lisp
@@ -192,7 +192,7 @@
               (mapcar 'print-key-seq bindings))
       (format stream "Command \"~a\" is not currently bound." cmd))))
 
-(defcommand where-is (cmd) ((:rest "Where is command: "))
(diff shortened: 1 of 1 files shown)


dmb2:
can we use the language 'is aliased to' rather than on?

PR #755, opened 2020-03-15
Make SPLIT-SEQ work with a nil separator

goose121:
As far as I can tell, `split-seq` is never actually called with a nil argument currently, but the code at the moment will completely ignore `nil` as a separator due to using `find`, which is an undocumented potential footgun.

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -771,7 +771,7 @@ positive direction."
             min))))
 
 (defun split-seq (seq separators &key test default-value)
-  "split a sequence into sub sequences given the list of seperators."
+  "Split a sequence into subsequences given the list of seperators."
   (let ((seps separators))
     (labels ((sep (c)
                (position c seps :test test)))
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -774,7 +774,7 @@ positive direction."
   "split a sequence into sub sequences given the list of seperators."
   (let ((seps separators))
     (labels ((sep (c)
-               (find c seps :test test)))
+               (position c seps :test test)))
       (or (loop for i = (position-if (complement #'sep) seq)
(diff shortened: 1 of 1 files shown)


equwal:
> As far as I can tell, `split-seq` is never actually called with a nil argument currently, but the code at the moment will completely ignore `nil` as a separator due to using `find`, which is an undocumented potential footgun.

I'm not sure if this is what you meant, but this test works as I would expect without change:
```
> (split-seq "  " nil)
(" ")
> (split-seq nil " ")
NIL
```
Can you provide a test case?

equwal:
May as well fix this docstring while doing this:
-  "split a sequence into sub sequences given the list of seperators."
+  "Split a sequence into sub sequences given the list of separators."

goose121:
Here's a test case:
```
> (split-seq #(1 nil 2) nil)
(#(1 NIL 2))
```

equwal:
> Here's a test case:
> 
> ```
> > (split-seq #(1 nil 2) nil)
> (#(1 NIL 2))
> ```

I see no difference here with or without the patch.

primitives.lisp
```diff
@@ -771,7 +771,7 @@ positive direction."
             min))))
 
 (defun split-seq (seq separators &key test default-value)
-  "split a sequence into sub sequences given the list of seperators."
+  "Split a sequence into subsequences given the list of seperators."
```
equwal: s/seperators/separators/

goose121:
> Here is what I think you want for a test case:
> 
> ```
> > (split-seq #(1 nil 2) '(nil))
> (#(1) #(2)) ; Only with your fancy patch.
> ```

Oh yeah, that's what I meant

PR #756, opened 2020-03-19
Allow escaped symbols to be used as commands

eihli:
The codebase allows commands to be symbols, but, for no reason I could
find, it did not support escaped symbols (|some symbol|).

Escaped symbols can sometimes make code more readable and with no good
reason to not support them it would be good for the code to not have
this hidden surprise.

There is also a small performance/clarity update where getting a value
from a hashmap by iterating the entries and checking equality is
replaced by `gethash`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -74,3 +74,4 @@ Daniel Oliveira     drdo at drdo.eu
 Panji Kusuma        epanji at gmail dot com
 Andrin Bertschi     hi at abertschi dot ch
 Spenser Max Truex   web ate spensertruex.com
+Eric Ihli           eihli at owoga com
--- a/command.lisp
+++ b/command.lisp
@@ -168,10 +168,7 @@ alias name for the command that is only accessible interactively."
 (defun dereference-command-symbol (command)
   "Given a string or symbol look it up in the command database and return
 whatever it finds: a command, an alias, or nil."
-  (maphash (lambda (k v)
-             (when (string-equal k command)
-               (return-from dereference-command-symbol v)))
-           *command-hash*))
+  (gethash (read-from-string command) *command-hash*))
 
 (defun command-active-p (command)
   (typep (current-group) (command-class command))
@@ -227,11 +224,12 @@ only return active commands."
   "Pop the next argument off."
   (unless (argument-line-end-p input)
     (flet ((pop-word (input start)
-             ;; Return the first word of INPUT starting from START and
+             ;; Return the first symbol of INPUT starting from START and
              ;; its end position.
              (let* ((p1 (position #\space input :start start :test #'char/=))
-                    (p2 (or (and p1 (position #\Space input :start p1))
-                            (length input))))
+                    (p2 (multiple-value-bind (text p2)
+                            (read-from-string input nil nil :start start)
+                          p2)))
                ;; we wanna return nil if they're the same
                (unless (= p1 p2)
                  (values (subseq input p1 p2)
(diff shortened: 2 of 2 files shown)


eihli:
This won't work. Commands are stored in the hashmap as namespaced keys. So `(string-equal "foo" 'stumpwm::foo)` is true, and we'd do `(gethash 'stumpwm::foo *map*)`, but then we'd be asking people to use namespaced keys for commands, which is a pretty big behavior change.

Since we don't support getting things that clash out of the command map, we might as well not allow putting things in that clash. The way things are, it's possible to put both `'stumpwm::foo` and `:my-stuff::foo` in the command map but it's arbitrary which when we'd get out when we do `(dereference-command-symbol "foo")`. That's a bug that I think I'll update the code to address by just putting non-namespaced keys in the command map and adding some notes around the code.

eihli:
Just making a few more notes for reference. I'm viewing the keys in `*command-map*` and see some keys are package-qualified and some are not and I haven't yet figured out why.

`(maphash (lambda (k v) (print k)) stumpwm::*command-hash*)`

```
WHERE-IS 
WHICH-KEY-MODE 
MODIFIERS 
DUMP-GROUP-TO-FILE 
STUMPWM::DUMP-GROUP 
DUMP-SCREEN-TO-FILE 
STUMPWM::DUMP-SCREEN 
DUMP-DESKTOP-TO-FILE 
STUMPWM::DUMP-DESKTOP 
RESTORE-FROM-FILE 
RESTORE 
```
I am guessing it has something to do with how Common Lisp expands and evaluates macros and when it interns symbols. Can anyone help explain? I was trying to make a change so that none of the keys are qualified.

jorams:
Commands are named by symbols, and they can be "escaped symbols" as well. You can also refer to them using strings, and comparisons are done case-insensitively.

```lisp
(defcommand |escaped-symbol-test| () () (echo "test succeeded"))
(|escaped-symbol-test|) ; echoes "test succeeded"
(run-commands "escaped-symbol-test") ; echoes "test succeeded"
(run-commands "EsCaPeD-SyMboL-TeSt") ; echoes "test succeeded"
```

Every symbol in that hash table is "package qualified", in that they all have a home package.

When you `print` them, some are displayed without a package prefix because you are in the `stumpwm-user` package, which `:use`s the `stumpwm` package. Symbols that are exported from the `stumpwm` package are thus accessible in `stumpwm-user` without explicitly specifying the package, so they are printed as such. Symbols that are not (like `dump-group`) are only accessible if you explicitly specify the package, so they are printed with `stumpwm::`.

If you switch to, for example, the `cl-user` package. Most of these symbols would be printed with `stumpwm:`.

```lisp
(in-package :cl-user)
(maphash (lambda (k v) (print k)) stumpwm::*command-hash*)
; ...
; STUMPWM:WHERE-IS
; STUMPWM::DUMP-GROUP 
; ...
```

I agree it's not ideal that there can be clashes. There can even be clashes between symbols in the same package, if they only differ by case. For example:

```lisp
(in-package :stumpwm)
(defcommand |where-is| () () (echo "haha, you fool"))
(run-commands "where-is") ; could run either WHERE-IS or |where-is|
```

I don't think that's something we can change without massive breakage.

eihli:
Thank you for explaining so thoroughly. Perhaps instead of changing code I'll just add comments documenting the clash possibility of command names and that escaped command names cannot contain spaces.

I should have worded things differently to make it clear that it wasn't the escapedness of the symbol that was broken, but that it was escaped symbols with spaces in them that were broken.

dmb2:
@eihli I'm going to close this for now, but welcome updated documentation!

PR #757, opened 2020-03-21
Document file writing behaviour of dump/restore

equwal:
`dump-to-file` and `restore-from-file` write to XDG_CONFIG_HOME, and add `.dump`
file extensions.


dmb2:
Can you make it $XDG_... to make it clear that it is an environment variable?

equwal:
@dmb2 I think that is what you meant.

PR #762, opened 2020-04-07
Exec and show

m-g-r:
hi again, what do you think about this?
Sometimes I just want to execute something *and* see the result for a short while in a message window.

I tried to follow the existing exec which is on `<prefix>-!`
So this would add run-shell-command-and-show-output, exec&show, and bind it to `<prefix>-M-!`

Also adding me to AUTHORS which I forgot in the last PR https://github.com/stumpwm/stumpwm/pull/759. 

Cheers, Max

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -74,3 +74,4 @@ Daniel Oliveira     drdo at drdo.eu
 Panji Kusuma        epanji at gmail dot com
 Andrin Bertschi     hi at abertschi dot ch
 Spenser Max Truex   web ate spensertruex.com
+Max-Gerd Retzlaff   m.retzlaff at gmx.net
--- a/bindings.lisp
+++ b/bindings.lisp
@@ -89,6 +89,7 @@ is a tile group.")
   (kbd "a")   "time"
   (kbd "C-a") "time"
   (kbd "!")   "exec"
+  (kbd "M-!") "exec&show"
   (kbd "C-g") "abort"
   *escape-fake-key* "send-escape"
   (kbd ";")   "colon"
--- a/user.lisp
+++ b/user.lisp
@@ -167,7 +167,13 @@ such a case, kill the shell command to resume StumpWM."
       (run-prog-collect-output *shell-program* "-c" cmd)
       (run-prog *shell-program* :args (list "-c" cmd) :wait nil)))
 
+(defcommand run-shell-command-and-show-output (cmd) ((:shell "/bin/sh -c "))
+  "Call run-shell-command, collect the output and show as message"
+  (message (run-shell-command cmd t)))
+
 (defcommand-alias exec run-shell-command)
+
+(defcommand-alias exec&show run-shell-command-and-show-output)
 
 (defcommand eval-line (cmd) ((:rest "Eval: "))
   "Evaluate the s-expression and display the result(s)."


m-g-r:
If there is a simpler way, just tell me. Also, if you folks think this is not of general interest, I can just add it to my local system only.

m-g-r:
Update: someone had a go at this already, see fist entry of https://github.com/stumpwm/stumpwm/wiki/Tips-And-Tricks but this PR is simpler.

Maybe we should at some timeout handling?

m-g-r:
Also I wonder, how the command line parsing of `stumpwm:run-commands` or rather `stumpwm::eval-command` is supposed to work, and if it shouldn't be possible to specify the `collect-output-p` option of `run-shell-command` somehow with that? I notice that you bind keys with define-keys to command strings. Those are command lines that are parsed, right? Not just bare names of commands.

dmb2:
I think this is the perfect candidate for a module.  You can make it more elaborate, or even incorporate pieces from the tips-and-tricks wiki page. 

We're happy to review the code and help you get things squared away too!

m-g-r:
hey @dmb2, do you really think this is big enough for a module and not of more general interest?
I've updated it and added timeout-handling and it is still tiny, see https://github.com/m-g-r/stumpwm/pull/2/files


Btw. looking at the comment of RUN-SHELL-COMMAND (https://github.com/stumpwm/stumpwm/blob/master/user.lisp#L161) that a not-stopping command hangs stumpwm I think it would be a good idea to add timeout handling to that one as well.

But in this case just a sb-ext:with-timeout around its body without any handling of that error as RUN-SHELL-COMMAND is used in a lot of modules and it is expected that it returns a string. If that one times out, this needs to be an error and a caller might handle that error or it will be an unhandled error which would be the correct behavior. (Mind you, this would not kill the running command in an uncontrolled fashion. It would just unwind the calling of RUN-SHELL-COMMAND, the process would be kept running, just preventing a long hanging stumpwm.)
A proposal might be:
```
(defcommand run-shell-command (cmd &optional collect-output-p (timeout 30)) ((:shell "/bin/sh -c "))
  "Run the specified shell command. If @var{collect-output-p} is @code{T}
then run the command synchonously with a timeout and collect the output.
After TIMEOUT seconds it stops waiting and a condition of type SB-EXT:TIMEOUT
is signaled. The command itself will continue to run until it is finished.
Be careful, the shell command will hang StumpWM until it returns or timeouts."
  (if collect-output-p
      (sb-ext:with-timeout timeout
        (run-prog-collect-output *shell-program* "-c" cmd))
      (run-prog *shell-program* :args (list "-c" cmd) :wait nil)))
```

Downside: modules might break as authors might ignore conditions
Upside: modules with non-finishing calls do not hang stumpwm

sdilts:
I can't speak for dmb2, but here are my two cents on why this is a great module but not a good addition to stumpwm proper, despite it's simplicity.

At least from my perspective, StumpWM is, and should only ever be, a window manager. The fact that it has commands like `run-shell-command` and the modeline is because they either existed in ratpoison or it was easier to implement those as part of the core rather than do them "properly" as a separate application. There is no compelling reason to get rid of those components now, as many people rely on them and would miss them if they were gone. The thing is, we also don't want to add new components like them, as if we ever get the manpower to actually do them properly, it is just more things to replace. In addition, drop-down terminals like [tilda](https://github.com/lanoxx/tilda) can provide something like this with little additional effort.

The contrib repo exists for two reasons. The first is that some things like audio controls and shutdown/reboot menus are very system specific, and we don't want that code to not work on someone's system just because they use don't use pulseaudio or on they use a platform that doesn't use systemd. The second reason it exists is because the particular functionality is good and useful, but it doesn't fit into the WM portion of the project. I use most of the modeline modules and a few of the util modules to get a working DE, and would consider using your code if I didn't already use a drop-down terminal. Your code is of general interest, but it just doesn't belong in the main StumpWM codebase. 

BTW, your changes to the `run-shell-function` would probably be a good thing to have. Why don't we discuss them on gitter? It's a better place to discuss new features than on an unrelated pull request.

m-g-r:
@sdilts thanks for your explanations. I do something on run-shell-function. Actually, handling a timeout there correctly and translating it to an error makes the second function run-shell-command-and-show-output  almost not existing.

Yes, there are always the design principles and trying to achieve loose coupling and separation of concerrns as well as coherence. So, there is a separation of concerns, that the main stumpwm should be just the wm, and contrib other additional things. That sounds good and reasonable. In this particular case, the two functions will not be loosely coupled, so this would mean that one function is in core stumpwm and the other, although highly dependent on the concrete behavior of the first, will be in different modules. It actually sounds like when I follow your arguments that both commands should rather reside in a module.

Staring with the European Lisp Symposium (els2020) last week I am a bit in the #stumpwm irc channel on freenode. But I am not so keen on using gitter in addition to that.

PR #768, opened 2020-04-19
fix a small typo in sample rc file macro.

Ramapitecus:
Need to use `nsubstitute' to mutate the string.


jorams:
I'm quite surprised this problem stuck around for 16 years, since it's quite obviously incorrect.

Thanks for the fix!

Ramapitecus:
It's funny. I was looking to see if it was reported already and found a bug I reported in the same macro 5 years ago. But also didn't see this. 

You're welcome.

dmb2:
I think this just proves that almost *no one* uses the example init file :)


On Monday, April 20, 2020 22:27 CEST, Halim G Naim H <notifications@github.com> wrote:
 
 
It's funny. I was looking to see if it was reported already and found a bug I reported in the same macro 5 years ago. But also didn't see this.
You're welcome.

PR #773, opened 2020-04-24
add recursively-add-to-load-path

m-g-r:
"like INIT-LOAD-PATH first recursively build a list of paths
that contain modules, then add them to the load path"

Open for suggestions to rename it or refactor it in a different way. But I just want to be able to add another modules directory that is handled just like the default modules directory by INIT-LOAD-PATH.

### DIFF
--- a/module.lisp
+++ b/module.lisp
@@ -32,8 +32,7 @@
           init-load-path
           set-module-dir
           find-module
-          add-to-load-path
-          recursively-add-to-load-path))
+          add-to-load-path))
 
 (defvar *module-dir*
   (pathname-as-directory (concat (getenv "HOME") "/.stumpwm.d/modules"))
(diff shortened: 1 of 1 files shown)


dmb2:
I'd rather just export build-load-path and have the user run mapcar themselves from the init file.

m-g-r:
Yeah, that's what I had before but I don't like the asymmetry of add-to-load-path and init-load-path... instead of getting rid of the new function we could actually go the other direction.. wait, I'll commit something...

m-g-r:
See commit msg of last commit, copied here for convenience:
```
replace INIT-LOAD-PATH by definition of RECURSIVELY-ADD-TO-LOAD-PATH 


This new definition is also more general and can be used to add more module
directories in the user's init file. Also, this should be more save then
the original definition of INIT-LOAD-PATH.

It gets rid of the dreaded SYNC-ASDF-CENTRAL-REGISTRY that uses UNION.
This is not correct as ASDF:*CENTRAL-REGISTRY* is an ordered list and
the lisp spec says in the documentation for UNION:
  "The order of elements in the result do not have to reflect the ordering
   of list-1 or list-2 in any way."
Here is an example that shows that also SBCL's definition of UNION whould
mess with the order in SYNC-ASDF-CENTRAL-REGISTRY (as SBCL for short lists
just adds to the shorter list):
  (union (list 1 2) '(1 2 3))
  => (3 1 2)
I actually consider this behavior of SYNC-ASDF-CENTRAL-REGISTRY a bug and
this commit fixes it as the new definition uses the existing ADD-TO-LOAD-PATH
which pushes to the front of the lists.
```

m-g-r:
Thanks,  @dmb2.

PR #774, opened 2020-04-25
add signal handler for POSIX signal SIGHUP to force a hard stumpwm restart

m-g-r:
_I gather this will be slightly more controversial... but I still suggest as it gives a means to salvage your X session on heavy problems._

FORCE-STUMPWM-RESTART will immediately force a restart outside stumpwm's
loops and is kind of harsh.

Sometimes when debugging I cannot invoke the command RESTART-HARD anymore.
But you cannot call RESTART-HARD from a swank as it needs to be in the
context of the CATCH :TOP-LEVEL.

The :HUP-PROCESS in function STUMPWM gave me the idea that stumpwm should
even do this when receiving SIGHUP and this is kind of nice so that you
can restart stumpwm even when you cannot invoke commands properly and you
do not have a running swank server anymore (which sometimes happens for me).

### DIFF
--- a/stumpwm.asd
+++ b/stumpwm.asd
@@ -16,7 +16,8 @@
                #:cl-ppcre
                #:clx
                #:sb-posix
-               #:sb-introspect)
+               #:sb-introspect
+               #:cffi)
   :components ((:file "package")
                (:file "primitives")
                (:file "wrappers")
--- a/stumpwm.lisp
+++ b/stumpwm.lisp
@@ -277,9 +277,25 @@ further up. "
   ;; what should the top level loop do?
   :quit)
 
+(defun force-stumpwm-restart (&key (close-display t))
+  (when close-display
+    (xlib:close-display *display*))
+  (apply 'execv (first sb-ext:*posix-argv*) sb-ext:*posix-argv*))
+
+(defmacro set-signal-handler (signo &body body) ;; from Andrew Lyon at https://stackoverflow.com/a/10442062
+  (let ((handler (gensym "HANDLER")))
+    `(progn
+       (cffi:defcallback ,handler :void ((signo :int))
+         (declare (ignore signo))
+         ,@body)
+       (cffi:foreign-funcall "signal" :int ,signo :pointer (cffi:callback ,handler)))))
+
 ;; Usage: (stumpwm)
 (defun stumpwm (&optional (display-str (or (getenv "DISPLAY") ":0")))
   "Start the stump window manager."
+  (set-signal-handler sb-posix:sighup
+    (dformat 0 "SIGHUP received: forcing immediate restart of stumpwm~%") ;; debug level 0 to "force" logging
+    (force-stumpwm-restart))
   (let ((*in-main-thread* t))
     (setf *data-dir*
           (make-pathname :directory (append (pathname-directory (user-homedir-pathname))
@@ -298,7 +314,7 @@ further up. "
               ;; the process because otherwise we get errors.
               ((eq ret :hup-process)
                (run-hook *restart-hook*)
-               (apply 'execv (first sb-ext:*posix-argv*) sb-ext:*posix-argv*))
+               (force-stumpwm-restart :close-display nil))
               ((eq ret :restart)
                (run-hook *restart-hook*))
               (t


m-g-r:
Ok, what do you think? I have this now for  a good week on my dev system. And it just helped me again to safe a X session by sending STUMPWM a SIGHUP to force restart itself.

PuercoPop:
@m-g-r I haven't had the time to look into it, but I vaguely recall reading something about SBCL not playing well with some signals. I didn't want to comment before I have had the time to look into what are the caveats, if any. I On principle I agree with handling SIGHUP.

A couple of things though:
- It is not necessary to use CFFI for this. You can use SBCL's own FFI for this. 
- Because the body of the handler is always the same there is no need to use a macro like `set-signal-handle`. Instead define the callback as the top-level and call `cffi:foreign-funcall` directly where set-signal-handle` is currently called.

m-g-r:
Hi @PuercoPop,

Yes, signals are a bit problematic in general with SBCL, You can read about it in http://www.sbcl.org/sbcl-internals/Signal-handling.html and I looked it up before writing this. I think if you are not especially careful you can botch the GC. But the thing is, that this is a last resort mechanism that will almost immediately replace the running lisp image with a new one via EXECV. So the old image will not be running anymore,  and in this circumstances that should be ok with SBCL.

When your StumpWM is hungup internally and you cannot control it anymore to send a proper RESTART-SOFT or RESTART-HARD from its command shell and you cannot connect via swank anymore, then you can send SIGHUP to trigger a forceful restart. (I also think sending a SIGHUP signal is something that only a more advanced user does.)

The problematic signal handling is also, why I tried to have the implementation of FORCE-STUMPWM-RESTART short and sweet. Still, we have to close the X display first, otherwise X stops (at least in my tests). In my tests it also actually worked to just close the display again, but I read that this is in general not safe to repeat this call so I've added the conditional `when close-display`. (But we could get rid of the conditional  by not calling it from function STUMPWM and leaving line 317 unchanged, https://github.com/stumpwm/stumpwm/pull/774/files#diff-3dcc4288b81c6a6fd4f9c728930cf8a7R317.)

To the other points:
* yes, I can also use SBCL's ffi not cffi if you prefer that. (I do use that one also sometimes.)
* yes, if this is the only signal we want to handle I can get rid of the macro

I could also make this just a module that a user can load only if he/she wants. But I guess module authors shouldn't be encouraged to write signal handling code. :)

PuercoPop:
Thanks for your detailed explanation. You seem to have covered all the bases. :+1: 

> - yes, I can also use SBCL's ffi not cffi if you prefer that. (I do use that one also sometimes.)

I do prefer it but I'm ok with using CFFI. What concerns me about adding a new dependency it would complicate the packaging of StumpWM, although I'm uncertain how big a complication it would be tbh. At least on ubuntu cffi is already packaged as a deb. Although not my case, I do know some users of StumpWM use their distribution's package/ports.

However if we are introducing a new dependency then we need to update the README and most importantly, we would need to wait for @dmb2's approval. Merging a change that introduces a new dependency would be above my pay grade.

> - yes, if this is the only signal we want to handle I can get rid of the macro

Yeah, after you change I this I would merge this PR (after @dmb2's approval if we decide to stick with CFFI)

> I could also make this just a module t

I think this belongs in StumpWM proper, not a module. And thanks for working on this :pray: :tada:

emacsomancer:
Just as a point of information, it looks like Ubuntu, Guix, Nix all have `cffi-cl` or `sbcl-cffi` or `lisp-cffi` packages, but neither Arch (even in the AUR) nor Void appear to have any cffi packages for cl. (but I don't think Arch has an official package for stumpwm, nor does Void). 

What are the drawbacks to using SBCL's own FFI over CFFI?

PuercoPop:
@emacsomancer I was thinking of our BSD users. They have 

> What are the drawbacks to using SBCL's own FFI over CFFI?

For this case I don't think they are any, but cffi has some extra features like [`define-foreign-library`](https://common-lisp.net/project/cffi/manual/html_node/Library-definition-style.html#Library-definition-style)

emacsomancer:
> @emacsomancer I was thinking of our BSD users.

But I think even on Linux there isn't universal packaging of CFFI, for what it's worth.

m-g-r:
I have a look at using SBCL's FFI. I don't like dependencies either, and as we are supporting only SBCL it is okay to directly use it.

m-g-r:
It's small but I haven't found the time, yet, as it is low priority for me. I'll see if I can do it the next two weeks.

dmb2:
no problem, my response times are longer than normal just now anyway.  @PuercoPop feel free to merge this once everything is straightened out.

ralt:
I know it's not particularly related to this issue, but I'd like to point out that execv will *not* close the currently opened file descriptors.

Whether that is an issue or not, I'll defer to other people.

dmb2:
I think we can probably let the file descriptors dangle since catching a signal is exceptional and probably not something you're going to allow to run for months after you catch the signal and reset.  

let's use sbcl's FFI if possible. This should be integrated in stumpwm proper, not a module.

dmb2:
Is this good to go?

Jehops:
I just read through the comments and thought that SBCL's FFI was preferred so that a new dependency would not be required.  Could you clarify?  I would like to be certain that adding this dependency was intentional before doing the work to update and create new OS packages.

Thanks.

m-g-r:
Hi, sorry, I have been busy. I can look again to replace it with sbcl's FFI if you still prefer that.

But I can confirm that I have to use it from time to time in order to rescue a stumpwm session. Say, once every one or two months. Am I the only one by the way? :)

(Just a few days ago when some window still had a progress bar open and would then close a window and I switched the stumpwm group already before it was finished and stumpwm didn't accept any commands anymore. Sadly, I forgot which program it was. Maybe I remember again....)

Max

Jehops:
Hi Max,

No need to apologize.  My question was as much to @dmb2 after 7c5a7a0.  I am not qualified to assess the merits of CFFI versus SBCL's FFI, but if more external dependencies aren't necessary, then ideally they wouldn't be required.  CFFI has its own dependencies like `Babel` and `trivial-features` and they also have new dependencies.

Jehops:
I also forgot to mention that I don't recall having to kill a locked session.  Maybe 5 years or so ago, I did see crashes with `(require 'swank)` and `(swank:create-server)` in `.stumpwmrc`, but I took those lines out back then and haven't had any major problems since.  Sometimes certain popup windows like browser notifications can cause StumpWM to get confused and not recognize commands.  In these situations, I've found using the mouse to set focus back on a window gets things back to normal again.

m-g-r:
Ok, see new PR https://github.com/stumpwm/stumpwm/pull/845

This was actually really easy as looking at CFFI's source code at https://github.com/cffi/cffi/blob/master/src/functions.lisp I realized I could just MACRO-EXPAND the call to set-signal-handler and then the internal calls to %foreign-funcall, and get a first SBCL alien version that I just had to clean up a bit...

Keep in mind that this is tested less. The old version was used for month on two machines, the new SBCL version only now on one machine several times. But then again, it is directly derived from the old definition (see previous paragraph).

m-g-r:
@dmb2 I apologize that I only made the rewrite now after you merged the branch. That is unfortunate.

PR #776, opened 2020-04-27
add missing keyword argument to docstring

Ramapitecus:
Was reading the docstring to use this macro, and noted that a keyword is not mentioned in the example.


sdilts:
Nice catch!

PR #781, opened 2020-05-08
modify `define-frame-preference` to handle floating windows

szos:
This pull request attempts to allow users a finer grained control of the default window type of a new window. 

It adds the hook `*auto-float-window-hook*`, as well as the custom calling function `auto-float-window-p`, and modifies the `group-add-window` method to force the creation of a new floating window if any function in `*auto-float-window-hook*` returns true. The custom calling function was necessary as `run-hook-with-args` doesn't return any values from the hooks functions. While I don't like straying away from the established hook functionality, I'm unsure how else to implement it without modifying `run-hook-with-args`. 

Motivation: 
One of the questions i have seen from new users and have had myself is how to force a window to be floating when in a tiling group. The two ways I have dealt with this are to A) hang a call to float-window on the *new-window-hook*, or B) switch to a floating group, create the window there, and then move it back to a tiling group. The problems with this are that A doesn't preserve the windows default size (useful for things like tor-browser, or some emacs child frames) and B takes a while and is frustrating.

Edit: If a hook doesn't seem like the right move, this could be implemented as a closure with a macro to force the launching of a window as a floating window. I threw together a small and incomplete example of this here: https://gist.github.com/szos/b0315747f2ae1abefbfd5feaf7d385a4


jorams:
I see this uses tabs for indentation, which causes problems. Could you run `untabify` on the changes? (assuming you're using Emacs)

I wonder if, instead of this specific solution, we could add something like `*new-window-preprocessing-hook*` (name TBD) that gets called before adding the window to the group. That could use the normal hooks machinery. We *would* still need a function to do the actual floating (basically `float-window` without the group), which might not be a good idea.

szos:
Apologies 'bout the tabs, should be fixed in the latest commit. 

I agree that using the normal hook machinery would be nice/best, but I'm unsure how `float-window` would work without the group. One of the problems is that we have to override the default behavior of `group-add-window` when called with a tile group. I figured the simple way of doing this is changing the `if` to a `cond` and adding an extra check, which requires some function to return true before hitting the t clause. Although as I write this we could add a function that checks if the window type is a floating window and if so do nothing. 

The problem of calling `float-window` without the group is an interesting one. The first thing to come to my mind is to add an extra `&optional pre-processing` argument to the function to determine whether we have processed the window or not. A solution off the top of my head would be: 
```
(flet ((float-existing-window (win grp)
	   (let ((frame (tile-group-current-frame grp)))
	     (change-class win 'float-window)
	     (float-window-align win)
	     (funcall-on-node (tile-group-frame-tree grp)
			      (lambda (f) (setf (slot-value f 'window) nil))
			      (lambda (f) (eq frame f)))))
	 (add-floating-window (win grp)
	   (change-class window 'float-window)
	   (float-window-align win)
	   (group-focus-window grp win)))
  (defun float-window (window group &optional pre-window-processing)
    (if pre-window-processing
	(add-floating-window window group)
	(float-existing-window window group))))
```

What do you think of that kind of solution?

sdilts:
I thought that `define-frame-preference` was supposed to be used for automatically floating windows, but as it doesn't actually do that, I'm really happy that this is getting implemented.

szos:
I've never been one for `define-frame-preference`, and hadn't thought of it when working on this implementation. I'm curious if it a better solution would be to check if the window matches any rules from within `group-add-window` and if the frame is `:float` to create it as a floating window, instead of adding a new hook and associated functions? The downside to this is that window rules would be iterated over twice when creating a window, but it could be a nicer interface. 

What do you think @jorams?

sdilts:
Frankly, I'm not really sure. I don't know how `define-frame-preference` works myself, and only brought it up because that was the first thing I reached for when trying to automatically float windows. I'd prefer not to add another hook or interface, but if we can't do this without major modifications to `define-frame-preference`, then I'd be perfectly fine with the current solution. I'll let someone who is more familiar with it decide.

szos:
I agree with you regarding adding hooks/interfaces, but I'm torn between that and wanting to be able to use arbitrary logic when determining whether to float a window.

jorams:
I don't know much about `define-frame-preference` and didn't think of it, but it does seem like the obvious place to implement this. Indeed there are some limitations to the rules available, but we could improve that for all preferences, not just the floating one. It seems like a good idea.

Adding it to the frame preferences would also allow the `remember` command to work for this (with a small modification to recognize floating windows), making it very easy to configure.

And indeed it looks like passing the type argument to `window-matches-properties-p` will always cause an error, so that's a bug that's been there for 12 years.

szos:
I'll take a look at `window-matches-properties-p` and `define-frame-preference` and see how they could be extended. Based on what the manual reports, I think what I'd like to see is that `target-group` gets extended to be able to match any group (ie the window stays in the group it was created in), `:frame` can have the value `:float` and it will create a floating window, and that some amount of arbitrary logic gets allowed. The last one is probably the most difficult. I'm thinking of adding a couple `&key` arguments that take a function and a keyword of either `:or` or `:and` which control how window properties and the function are counted (ie one must match or both must match).

szos:
Given that this pull request has gone a little off the rails, with multiple solutions being implemented and then reverted, I'm curious if this pull request should be closed and a clean one opened without all of these extra commits? 

After living with the modified `define-frame-preference` and accompanying functions I think that it is the best solution, as it doesn't add new functions and variables, but extends the preexisting machinery.

sdilts:
A new pull request with a clean history would be much appreciated. Thanks for implementing this!

PR #791, opened 2020-06-02
Don't lose focus when fetching completions

ralt:
There's this really annoying issue when you have a lot of completions:
while it's displaying them, the focus is lost to the underlying
window. So if you type "fi" for firefox (in `<prefix>-!`), the "i" is
sent to the underlying window.

It may be due to me using a truetype font which is slower to render,
or the fact that a single letter will yield a lot of completion
candidates.

Either way, making sure we keep the focus during the whole insertion
solves this issue.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -337,20 +337,20 @@ match with an element of the completions."
                                       (caar compls)
                                       (car compls))))))
              (key-loop ()
-               (loop for key = (with-focus (screen-input-window screen)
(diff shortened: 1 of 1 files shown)


PuercoPop:
Thanks for the PR. Merged in https://github.com/stumpwm/stumpwm/commit/7b7be7fd9bd97cac58976baa4701e57a56ac33cf

PR #792, opened 2020-06-07
Float windows in tiling groups using `define-frame-preference`

szos:
Cleaned up the commits and opening a new pull request to make `define-frame-preference` handle auto-floating windows in tiling groups. Previous pull request was #781.

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -1121,10 +1121,16 @@ add rules")
   "Create a rule that matches windows and automatically places them in
 a specified group and frame. Each frame rule is a lambda list:
 @example
-\(frame-number raise lock &key from-group create restore dump-name class instance type role title)
+\(frame-number raise lock &key from-group create restore dump-name class class-not
+instance instance-not type type-not role role-not title title-not 
+match-properties-and-function match-properties-or-function)
 @end example
 
 @table @var
+@item target-group
+When nil, rule applies in the current group. When non nil, @var{lock} determines 
+applicability of rule
+
 @item frame-number
 The frame number to send matching windows to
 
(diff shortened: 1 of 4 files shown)


window-placement.lisp
```diff
@@ -28,33 +28,61 @@
 (defun string-match (string pat)
   (ppcre:scan (get-or-create-rule-scanner pat) string))
 
-(defun window-matches-properties-p (window &key class instance type role title)
+
```
sdilts, line 31: You added an extra space here.

szos: Thanks for catching that, not sure how i missed it

primitives.lisp
```diff
@@ -1154,27 +1160,55 @@ When non-NIL the group is restored even if it already exists. This arg should
...
+regardless of whether the window properties match. Takes one argument, the window.
 @end table"
   (let ((x (gensym "X")))
     `(dolist (,x ',frame-rules)
        ;; verify the correct structure
-       (destructuring-bind (frame-number raise lock
-                                         &rest keys
-                                         &key from-group create restore class instance type role title) ,x
-         (declare (ignore from-group create restore class instance type role title))
+       (destructuring-bind (frame-number raise lock &rest keys
```
PuercoPop, line 1203: I vote to remove the `&key` parameters from the code. They are used as a way to document what keywords a frame-rule has but that is already documented in the docstring.

If we decide against it then please leave the `&rest` parameter on its own line. There is a method to the madness :grimacing:

tile-group.lisp
```diff
@@ -1347,12 +1352,12 @@ direction. The following are valid directions:
...
-	      :for distance := (sqrt (+ (square (- win-x frame-x))
-					(square (- win-y frame-y))))
-	      :unless shortest
+              :for (frame-x . frame-y) := (frame-centroid frame)
+              :for distance := (sqrt (+ (square (- win-x frame-x))
+                                        (square (- win-y frame-y))))
+              :unless shortest
                 :do (setf shortest (cons distance frame))
-	      :when (> (car shortest) distance)
+              :when (> (car shortest) distance)
```
PuercoPop, line 1360: Don't untabify code unrelated to the PR please:pray:

window-placement.lisp
```diff
@@ -66,8 +93,15 @@
...
-          (declare (ignore from-group lock class instance type role title))
+                             &key from-group create restore class class-not instance
+                               instance-not type type-not role role-not title title-not
+                               match-properties-and-function
+                               match-properties-or-function)
+            match
+          (declare (ignore from-group lock class class-not instance
+                           instance-not type type-not role role-not
+                           title title-not match-properties-and-function
+                           match-properties-or-function))
```
PuercoPop, line 104: Instead of adding variables to the ignore declaration, remove them from the d-bind.

tile-group.lisp
szos: Apologies! after having some issues with emacs and tabs I set up emacs to select the whole file and untabify it. I'll remove that from my .emacsrc and do it manually from now on, and fix this up.

primitives.lisp
szos: That works for me, I'll remove it unless theres a reason to keep them.

tile-group.lisp
```diff
@@ -66,26 +66,31 @@
...
-          (setf (frame-window (window-frame window)) window))
-        (when (and frame raise)
-          (setf (tile-group-current-frame group) frame
-                (frame-window frame) nil))
-        (sync-frame-windows group (window-frame window))
-        (when (null (frame-window (window-frame window)))
-          (frame-raise-window (window-group window) (window-frame window)
-                              window nil)))))
+  (cond ((typep window 'float-window)
+         (call-next-method))
```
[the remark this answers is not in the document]
szos: That seems like a good solution, I'll implement it.

tile-group.lisp
PuercoPop: >  I set up emacs to select the whole file and untabify it

Samesies. I have `(untabify (point-min) (point-max))` as a before-save hook. No need to remove that setting. Just don't stage the changes when committing. If using the CLI use `git add -p`, if you use magit you can select the region of the diff before pressing s. That way you only stage those changes.
I have `()`. You can just not stage those changes when making the commit.

szos: ah, i didnt know of that functionality, thanks!

window-placement.lisp
PuercoPop: I just realized this recommendation was wrong. I forgot to add that we have to add `&allow-other-keys` otherwise destructuring-bind will raise an error. Sorry about that. Will commit the fix soon

PR #796, opened 2020-06-13
Log to a file by default

PuercoPop:
The new mechanism logs to a file by default. It also keeps the two oldest
log files. This way when a user wants to enable persistent logging all they have to do is `(setf *debug-level* X)`. The previous `redirect-all-output` approach still works.

Additionally this commit adds basic support for the XDG base directory
specification. We only use it for the location of the log file, but it
could be used in the future to place the configuration file according
to the XDG base directory specification.

I would also like to slowly move things that aren't directly related to window managing responsibilities to a module named support. If you are familiar with Rails think along the lines of active support. Things like auto-complete strategies or run-prog would go there.

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -821,51 +821,6 @@ string which is split to obtain the individual regexps. "
          (post (subseq list nth)))
     (nconc pre (list item) post)))
 
-(defvar *debug-level* 0
-  "Set this variable to a number > 0 to turn on debugging. The greater the number the more debugging output.")
-
-(defvar *debug-expose-events* nil
-  "Set this variable for a visual indication of expose events on internal StumpWM windows.")
-
-(defvar *debug-stream* (make-synonym-stream '*error-output*)
-  "This is the stream debugging output is sent to. It defaults to
-*error-output*. It may be more convenient for you to pipe debugging
-output directly to a file.")
-
-(defun dformat (level fmt &rest args)
-  (when (>= *debug-level* level)
-    (multiple-value-bind (sec m h) (get-decoded-system-time)
-      (format *debug-stream* "~2,'0d:~2,'0d:~2,'0d ~2,' d " h m sec level))
-    ;; strip out non base-char chars quick-n-dirty like
-    (write-string (map 'string (lambda (ch)
-                                 (if (typep ch 'standard-char)
-                                     ch #\?))
-                       (apply 'format nil fmt args))
-                  *debug-stream*)
-    (force-output *debug-stream*)))
-
-(defvar *redirect-stream* nil
-  "This variable Keeps track of the stream all output is sent to when
-`redirect-all-output' is called so if it changes we can close it
-before reopening.")
-
-(defun redirect-all-output (file)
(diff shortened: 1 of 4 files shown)


equwal:
`fdump.lisp` and `stumpwm.lisp` could probably use these XDG utilities more cleanly than the current way compatibility is done for dump/restore, so whether the logging is added or not I'd like to see the XDG file added. Maybe split that off into another pull request.

https://github.com/stumpwm/stumpwm/pull/748

sdilts:
An implementation of the XDG spec is built into `uiop`, which is included with ASDF. Should we use that instead?

https://common-lisp.net/project/asdf/uiop.html#UIOP_002fCONFIGURATION

support/xdg-base-dirs.lisp
```diff
@@ -0,0 +1,47 @@
...
+(defun config-dir ()
+  (let ((config-home (or (sb-ext:posix-getenv "XDG_CONFIG_HOME")
+                         (merge-pathnames ".config/" (user-homedir-pathname)))))
+    (merge-pathnames "stumpwm/"
+                     config-home)))
+
+(defun ensure-config-dir ()
+  (let ((config-dir (config-dir)))
+    (unless (probe-file config-dir)
+      (sb-ext:run-program "mkdir" `("-m" "700" "-p" ,config-dir)))))
```
sdilts, line 36: It's not likely, but this will fail if the parent directory doesn't exist. Use [ensure-directories-exist](http://clhs.lisp.se/Body/f_ensu_1.htm) instead?

PuercoPop: I wasn't aware of `ensure-directories-exist`. It is certainly preferable to shelling out the mkdir, done.

dmb2:
I'm happy to merge this if/when its ready.  Two thoughts:
I prefer uiop to using getenv for the xdg directory.

Part of me wants to keep the '`.stumpwmrc`' and the '`.stumpwm.d`' directory because that's what I use/prefer, but the other part of me recognizes the cleanliness of following XDG as well as the compliance with desktop standards.

equwal:
> I'm happy to merge this if/when its ready. Two thoughts:
> I prefer uiop to using getenv for the xdg directory.
> 
> Part of me wants to keep the '`.stumpwmrc`' and the '`.stumpwm.d`' directory because that's what I use/prefer, but the other part of me recognizes the cleanliness of following XDG as well as the compliance with desktop standards.

A lot of programs support "partial" XDG support, where the choice to use XDG directories has to be manually configured.

https://wiki.archlinux.org/index.php/XDG_Base_Directory#Partial

I think this is the way StumpWM should go, and keep the old way of cluttering `~` by default.

PuercoPop:
> Part of me wants to keep the '.stumpwmrc' and the '.stumpwm.d' directory because that's what I use/prefer, but the other part of me recognizes the cleanliness of following XDG as well as the compliance with desktop standards.

After submitting this, I've come to agree that having a central .stumpwm.d directory would be better UX. I'll update the PR to remove the XDG implementation.

Issue #800, opened 2020-08-10
Unable to type for prompted argument on gravity and resize-direction


epanji:
When using command interactively without typing argument, command will pop a prompt asking for input. Error will occured when trying to type anything.

How to reproduced:
- Open colon __C-t ;__ and type command __gravity__ __\<enter\>__
- In prompt Gravity: __\<type anything\>__
- Error: The value ... is not of type ...
- Same error apply for command __resize-direction__

PR #801, opened 2020-08-10
Fixing stumpwm-type :gravity and :direction

epanji:
Wrong optional argument __completions__ passed on function __argument-pop-or-read__ causing error for issue https://github.com/stumpwm/stumpwm/issues/800#issue-676053690.


jorams:
LGTM :+1:

PR #813, opened 2020-09-27
Get dependencies from parsing stumpwm.asd

femnad:
This patch adds a Lisp file for listing dependencies defined in `stumpwm.asd` and changes the Makefile template to utilize that file to identify source files. This fixes a case make cannot locate dependencies because the dependency is in a subdirectory (like `support/debug.lisp`) but it is not easy to find that out by grepping source files in `stumpwm.asd`.


PuercoPop:
I completely forgot about the fact that we use grep to build the dependency list for the Makefile. But honestly that is an awful hack. I'd rather we stop doing that. We could use ASDF's build-op to make an executable. The only thing we would need to figure out is how to deal with extracting the docstrings for building the manual though. Or just keep the source tree flat.

What do y'all think?

femnad:
As someone not well-versed in Lisp ecosystem, the ability to use make to compile Stumpwm is really helpful. However, I would be just as happy if there was a script to achieve the same purpose.

tychoish:
I think it's possible to leave some kind of makefile/shim to make the project accessible to people who aren't as used to lisp, but I think that we should build the binary using asdf.

dmb2:
I agree with @tycoish.

PuercoPop:
to be clear I'm not proposing to stop using Make, just to stop using grep to extract the file dependencies for the Make rule. One way to do so would be what this PR is doing iirc, which is using ASDF to build a small executable to extract the dependencies. The other way would be to punt the problem to ASDF and have the rule be a `.PHONY` one like:

```make
stumpwm: stumpwm
	sbcl --eval '(asdf:make "stumpwm/build")'
```

The make file would stay either way.

femnad:
I'd be happy for this one to be closed to make way for #814 by the way, however, I think there is need for a band-aid until then as building/installing with make is not working on master. Alternative (less cumbersome?) approaches could be adding a symlink to `debug.lisp` in the repo root or removing the dependencies for the current `stumpwm` target in the Makefile and making it a `.PHONY`, similar to @PuercoPop suggested above.

femnad:
@PuercoPop based on your comment in #815 I realized that `sudo make install` wasn't working since we're trying to load the stumpwm system for listing the source files but the root user usually doesn't have the necessary Quicklisp initialization nor the required packages. So I added a shell script for listing source files. It's even more hacky and has many assumptions, however, at least it's a working band-aid this time.

PuercoPop:
@femnad As a bandaid I'd rather we list the files explicitly tbh. @dmb2 @tychoish What do you think?

PuercoPop:
After talking to @tychoish on IRC we decided to flatten the tree for the time being. [This commit](https://github.com/stumpwm/stumpwm/commit/14b203e2196016f923565a2b3f53cc76609b5dfd) should fix the issue. Can someone else confirm to be sure?

femnad:
Yep it works, closing this one.

PR #814, opened 2020-09-27
Add program-op to build StumpWM using only ASDF

PuercoPop:
This is a starting point for my suggestion in https://github.com/stumpwm/stumpwm/pull/813

To generate the executable in a REPL evaluate:

  (asdf:make "stumpwm/build")

This creates an executable named stumpwm on the current directory.

Before we could merge this we would still need to use `(uiop:command-line-arguments)` to have a CLI option to extract the manual instead. With this change we could simplify the stumpwm rule in the Makefile to be just stumpwm, the file.

### TODO

- [ ] Add CLI option to extract the docstrings to build the manual
- [ ] Update Makefile

### DIFF
--- a/main.lisp
+++ b/main.lisp
@@ -1,0 +1,30 @@
+;; Copyright (C) 2020 Javier Olaechea
+;;
+;;  This file is part of stumpwm.
+;;
+;; stumpwm is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


dmb2:
I like this idea, but I'm holding off on merging because I'm about to tag a new release.

dmb2:
I'm ready to merge this when you are @PuercoPop

WhyIsEvery4thYearAlwaysBad:
@PuercoPop @dmb2 Any updates?

PR #817, opened 2020-10-02
Fix invalid use of makunbound

madand:
This prevents TYPE-ERROR during StumpWM shutdown process.


PuercoPop:
Thanks for fixing this issue!

PR #818, opened 2020-10-03
Recover if randr version of make-screen-heads fails.

ydl7:
For example, on Android with XServer XSDL, randr capability is detected but cannot at present be used.


tychoish:
This seems reasonable change. I wish there were a test for things like this.

ydl7:
I do not think there can be a generic test to detect "things like this" where pessimistic failure modes are taken. Is the absence of such a test enough to stop the merge of this change?

PR #819, opened 2020-10-06
Don't call xlib:wm-size-hints-*-inc when hints are empty

### LINKED ISSUE #775: Error in command 'redisplay': No applicable method for generic function WM-SIZE-HINTS-WIDTH-INC
The only way I've been able to produce this error is install python arcade library (`pip install arcade==2.3.15`) and run the following python script:

    import arcade
    arcade.Window(800, 600, 'Foo', resizable=True)
    arcade.run()

With that window open and active, executing the "redisplay" command would result in the following error:

    Error In Command 'redisplay': There is no applicable method for the generic function
                                               #<STANDARD-GENERIC-FUNCTION XLIB:WM-SIZE-HINTS-WIDTH-INC (1)>
                                             when called with arguments
                                               (NIL).

This is in Ubuntu 18.04 with Python 3.7.5. StumpWM is at the latest master version as of now, and the `version` command reports: `19.11-140-g7f43263`.

PuercoPop:
Instead let them default to the default value of one. This fixes the
bug that occurs when (window-normal-hints window) return nil, which
results in the following error.

There is no applicable method for the generic function
   #<STANDARD-GENERIC-FUNCTION XLIB:WM-SIZE-HINTS-WIDTH-INC (1)>
   when called with arguments
   (NIL).

Fixes #775

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -620,13 +620,15 @@ and bottom_end_x."
 (defun window-width-inc (window)
   "Find out what is the correct step to change window width"
   (or
-    (xlib:wm-size-hints-width-inc (window-normal-hints window))
+   (when-let ((window-hints (window-normal-hints window)))
+     (xlib:wm-size-hints-width-inc (window-normal-hints window)))
     1))
 
 (defun window-height-inc (window)
   "Find out what is the correct step to change window height"
   (or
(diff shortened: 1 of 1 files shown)


PuercoPop:
Closed in favor of https://github.com/stumpwm/stumpwm/pull/824

PR #821, opened 2020-10-07
Initialize the data dir with 0700 permissions

femnad:
The previous permissions were missing the executable bit so `stumwm.log` couldn't be created since it's parent directory was missing the required permission to contain files.

Set the group and other permission to `0` as I think only giving them execute permissions won't enable those entities to access this directory.


PuercoPop:
Thanks for looking into this!

PR #824, opened 2020-10-08
Fix redisplay on certain windows

### LINKED ISSUE #820: Error In Command 'redisplay': The value NIL is not of type XLIB:WM-SIZE-HINTS
To reproduce open random image with `feh` and execute `redisplay` produces this error message:
```text
Error In Command 'redisplay': The value
                                           NIL
                                         is not of type
                                           XLIB:WM-SIZE-HINTS
```

Also `feh` repeats the image file because I guess it doesn't know corrently width and height of the window. To reproduce this issue:
* Open random file with `feh`. It repeats the image both vertically and horizontally. 
* Split the screen horizontally or vertically (`vsplit` / `hsplit`)
* Return back to single window (`only`).
* Now `feh` shows image without repeats.

Screenshot feh repeating issue:
ksnip_20201006-124307

My StumpWM version is = `19.11 Compiled On Tue Sep 29 2020 18:39:46`
My feh version is = `feh version 3.5
Compile-time switches: curl exif verscmp xinerama`

Related issue from another window manager: https://github.com/swaywm/sway/issues/3106

_Originally posted by @erenhatirnaz in https://github.com/stumpwm/stumpwm/issues/775#issuecomment-704162690_

CatherineTower:
When redisplay is called on windows with a nil normal-hints, redisplay
throws an error. This modifies window-width-inc and window-height-inc
so they always return a number

Fixes #820

### DIFF
--- a/window.lisp
+++ b/window.lisp
@@ -619,15 +619,15 @@ and bottom_end_x."
 
 (defun window-width-inc (window)
   "Find out what is the correct step to change window width"
-  (or
-    (xlib:wm-size-hints-width-inc (window-normal-hints window))
-    1))
+  (if (window-normal-hints window)
+      (xlib:wm-size-hints-width-inc (window-normal-hints window))
+      1))
 
 (defun window-height-inc (window)
   "Find out what is the correct step to change window height"
-  (or
-    (xlib:wm-size-hints-height-inc (window-normal-hints window))
-    1))
+  (if (window-normal-hints window)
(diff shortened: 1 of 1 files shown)


PuercoPop:
This is a dupe of https://github.com/stumpwm/stumpwm/pull/819, but I find your solution to be cleaner. Thanks :tada:

CatherineTower:
Oh shucks, I missed that. Happy to help out!

Issue #838, opened 2020-12-25
[Proposal] Add Customization System


sdilts:
# Proposal: Add Customization System

This is a proposal to add a customization system to StumpWM to make it easier view customization options and ensure that the values that the customizable variables are set to are correct. Inspired by the discussion in #832.

## Rational
StumpWM has a large amount of configuration options that are available to users. You can change things like mouse behavior, colors, the modeline, window appearance, and others. While most of these customization can be accomplished by setting global variables, a configuration system to ensure that these features are self-documenting, self-validating, and easy to use would make it easier for users to  configure their system.

The configuration system would allow users to do three main things: examine available settings, view each setting's default values, and set the values to a valid value.

### Exame Customizable Settings
It should be possible to retrieve a list of all the customizable settings using a function. This would return a list of all of the available customizable variables along with their valid values / validator functions. This function might have optional parameters that allow you to filter the returned list.

It should also be possible to examine an individual setting to view it's default value,  current value, and documentation. An added bonus would also be the ability to get the information that is used to validate variable's value. Two functions should exist, one that gets the information in a data structure and one to get the information as a printable string.

### Setting Customizations
There should be a way to set the customizable variables  so that the value is validated before the variable is changed. When the given value is incorrect, an error condition should be raised.

### Create new Customizable Settings
There should be at least one `defparameter` style macro that registers a setting, it's documentation, as well as validation information. Since it is probably a very common use case, it would be nice to have a version that accepts a list of possible values that the setting can hold.

### Grouping variables
While this feature isn't as important as the other ones and optional, it would be nice to be able to categorize settings to make them more searchable or separate them out on a per-module basis.

### Implemented in a separate package
This should be implemented in it's own package that doesn't depend on any of StumpWM's internals so that it is easier to reuse and not coupled to the current implementation.

szos:
I've started work on a more robust system than the one talked about in #832 ([located here, check branch functionize](https://github.com/szos/defcustom/tree/functionize)) and I'm running into some design issues. I would like to make this system more general and allow for validation of not only dynamic places but of slots and the like as well, both through accessors and slot-value/with-slots. The previous system works fine for symbols (and symbols only) and can easily be translated into the system currently being written, but when it comes to extending it like I mentioned everything falls apart. I'd love to get some input from people more knowledgeable than myself on how this should be designed.

montaropdf:
Waw, I am really impressed by the work done in such a small amount of time, after I open #832.

[...]

PR #839, opened 2020-12-26
Add basic configuration system

sdilts:
This PR adds a basic configuration utility package that covers all of the essential elements of a configuration system as described in #838. Before merging this commit, look at https://github.com/szos/defcustom/tree/functionize, which @szos is writting to solve the same problem. Comparing the two, it looks like szos's version has many more features and nice-to-haves, while the code here just meets the minimal requirements and is much simpler.

## Features

This PR adds the macros `define-config-parameter` and `define-config-var` for declaring variables that are meant to be modified by the system users. These macros register the variables and declare them using `defparameter` and `defvar` respectively. A validation function is given to this macro that is used for checking that the variable's value conforms to specific requirements. `define-config-enum` is also present, which works the same way as `define-config-parameter` except the the validation function is generated automatically.

To set configuration variables, you can use the `set-config` and `set-config-atomic` macros. These macros follow the same form as `setf` and verify that the values the variables are being set to are correct using the provided validation function. If the values aren't valid, an `invalid-datum-error` condition is raised. The difference between these two functions are that if an error condition is raised while `set-configuration-atomic` is being executed, all of the variables in form are reset to their values from before the form was executed.

You can retrieve information about a specific configuration variable by using the `get-configuration-info` function, and get a list of all of the known configuration variables with `all-config-info`. You can print information about individual variables or variables matching a pattern to a stream using `describe-config-info` and `describe-all-config-info`, respectively.

A `with-atomic-update` macro is also provided, which resets a given set of variables to their values before the form was executed if a condition was raised while the form was executing.

Since the code uses `defparameter` and `defvar` under the hood, it is completely backwards compatible with existing code.

### DIFF
--- a/config-system.lisp
+++ b/config-system.lisp
@@ -94,7 +94,8 @@
   (check-type name symbol)
   (with-gensyms (default-value is-valid func)
     `(progn
-       (,declare-type ,name ,default ,documentation)
+       (,declare-type ,name ,default ,@(when documentation
+                                         (list documentation)))
        (let* ((,default-value ,name)
               (,func ,validator)
               (,is-valid (funcall ,func ,default-value)))
@@ -104,21 +105,17 @@
                                      :validator ,func :doc ,documentation))
              (error 'invalid-datum-error :place (quote ,name) :value ,default-value))))))
 
-(defmacro define-config-parameter (name default validator documentation)
+(defmacro defconfig (name default validator &key documentation reinitialize)
   "Create and register a configurable variable with the given default value,
-validator function, and documentation using DEFPARAMETER"
-  (%generate-config-var-code 'defparameter name default validator documentation))
+validator function, and documentation"
+  (%generate-config-var-code (if reinitialize `defparameter 'defvar)
+                             name default validator documentation))
 
-(defmacro define-config-var (name default validator documentation)
-  "Create and register a configurable variable with the given default value,
-validator function, and documentation using DEFVAR"
-  (%generate-config-var-code 'defvar name default validator documentation))
-
-(defmacro define-config-enum (name default values documentation &key (test-fn #'equal))
-  "Same as DEFINE-CONFIG-PARAMETER, except the validation function
-is built by creating a function that checks if the set  value is in the VALUES list."
-  `(define-config-parameter ,name ,default (lambda (x) (member x ,values :test ,test-fn))
-                            ,documentation))
+(defmacro define-config-enum (name default values &key (test-fn #'equal) documentation reinitialize)
+  "Same as DEFCONFIG, except the validation function
+is built by creating a function that checks if the set value is in the VALUES list."
+  `(defconfig ,name ,default (lambda (x) (member x ,values :test ,test-fn))
+              :documentation ,documentation :reinitialize reinitialize))
 
 (defun all-config-info (&key (name-matches ".*") (package-matches ".*"))
   "List all of the available customizable settings matching the given criteria."
--- a/config-system.lisp
+++ b/config-system.lisp
@@ -1,0 +1,214 @@
+(in-package #:config-system)
+
+(defparameter *config-vars* (make-hash-table)
+  "Map that holds all of the configuration variables for the system")
+
+(defstruct config-info
+  (name t :type symbol :read-only t)
+  (validator #'identity :type (function (*) boolean))
+  (default nil :read-only t)
+  (value nil)
+  (doc "" :type string :read-only t))
+
+(defun %full-symbol-string (symb)
+  "Get the full name of the symbol complete with the package name"
+  (declare (type symbol symb))
+  (let* ((pkg (symbol-package symb))
+         (pkg-name (package-name pkg))
+         (symb-name (symbol-name symb)))
+    (concatenate 'string pkg-name
+                 (multiple-value-bind (symb status) (find-symbol symb-name pkg-name)
+                   (declare (ignore symb))
+                   (if (eql status :internal)
+                       "::"
+                       ":"))
+                 symb-name)))
+
+(defun describe-config-info (info &optional (stream *standard-output*))
+  (declare (type config-info info))
+  (format stream "Setting Name: ~A~%  Documentation:~%    ~A~%  Default value: ~S~%Current value: ~S~%  Validator function: ~A~%"
+          (%full-symbol-string (config-info-name info))
+          (config-info-doc info)
+          (config-info-default info)
+          (config-info-value info)
+          (config-info-validator info)))
+
+(defun %make-match-readtable (str)
+  "Try to make the given string have the correct case for a symbol"
+  (declare (type string str))
+  (ccase (readtable-case *readtable*)
+    ;; we are missing :invert, but I don't feel like implementing that.
+    (:upcase (string-upcase str))
+    (:downcase (string-downcase str))
+    (:preserve string)))
+
+(defun %map-config-matching (name-matches package-matches function)
+  (let ((name-scanner (ppcre:create-scanner (%make-match-readtable name-matches)))
+        (pkg-scanner (ppcre:create-scanner (%make-match-readtable package-matches))))
+    (maphash (lambda (key value)
+               (declare (ignorable key))
+               (let ((pkg-name (package-name (symbol-package (config-info-name value))))
+                     (symb-name (symbol-name (config-info-name value))))
+                 (when (and (funcall name-scanner symb-name 0 (length symb-name))
+                            (funcall pkg-scanner pkg-name 0 (length pkg-name)))
+                   (funcall function value))))
+             *config-vars*)))
+
+(defun describe-all-config-info (&key (stream *standard-output*) (name-matches ".*") (package-matches ".*"))
+  (%map-config-matching name-matches package-matches
+                        (lambda (info)
+                          (describe-config-info info stream)
+                          (format stream "~%"))))
+
+(define-condition config-error (error) ())
+
+(define-condition invalid-datum-error (config-error)
+  ((place-symbol :initarg :place :initform nil
+		 :accessor invalid-datum-error-place
+                 :type symbol)
+   (value :initarg :value :initform nil
+	  :accessor invalid-datum-error-value))
+  (:report
+   (lambda (c s)
+     (with-slots (place-symbol value) c
+       (format s "The value ~S is invalid for variable ~S." value place-symbol)))))
+
+(define-condition config-not-found-error (config-error)
+  ((place-symbol :initarg :place :initform nil
+                 :accessor config-not-found-error-place
+                 :type symbol)
+   (alternatives :initarg :alternatives :initform nil
+                 :accessor config-not-found-alternatives
+                 :type list))
+  (:report
+   (lambda (c s)
+     (with-slots (place-symbol alternatives) c
+       (if alternatives
+        (format s  "Setting ~A not found. Did you mean one of these? ~A"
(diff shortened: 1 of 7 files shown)


config-system.lisp
```diff
@@ -0,0 +1,157 @@
...
+  ;; if we are given a lambda, we need to evaluate it in order to call it
+  (%generate-config-var-code 'defparameter name default validator documentation))
+
+(defmacro define-config-var (name default validator documentation)
+  "Create and register a configurable variable with the given default value,
+validator function, and documentation using DEFVAR"
+  ;; if we are given a lambda, we need to evaluate it in order to call it
+  (%generate-config-var-code 'defvar name default validator documentation))
+
+(defmacro define-config-enum (name default values documentation)
```
sdilts, line 79: There are no tests for this macro.

config-system.lisp
```diff
@@ -0,0 +1,157 @@
...
+  "Same as DEFINE-CONFIG-PARAMETER, except the validation function
+is built by creating a function that checks if the set  value is in the VALUES list."
+  ;; if we wanted to get fancy, we could define a special type for this enumeration at this point too.
+  `(define-config-parameter ,name ,default (lambda (x) (member x ,values :test #'equal))
+                            ,documentation))
+
+(defun list-all-configurations ()
```
sdilts, line 86: I'm not a big fan of this function name. Maybe `get-all-config-info`?

config-system.lisp
```diff
@@ -0,0 +1,157 @@
...
+    ;; for some reason, the package isn't always included with the symbol name even
+    ;; if the top level isn't in the symbol's package, so we need to manually get the symbol name.
+    (mapcar #'%full-symbol-string
+            (remove-if-not (lambda (x) (string-equal (symbol-name x) name))
+                           (mapcar #'config-info-name (hash-table-values table))))))
+
+(defun get-configuration-info (setting-name)
+  "Find the info for the config variable stored in the symbol SETTING-NAME."
+  (declare (type symbol setting-name))
+  (gethash setting-name *config-vars*))
```
sdilts, line 104: The functions have the hash table hard-coded in, which will probably lead to problems in the future.

config-system.lisp
```diff
@@ -0,0 +1,157 @@
...
+       (let* ((,default-value ,name)
+              (,func ,validator)
+              (,is-valid (funcall ,func ,default-value)))
+         (if ,is-valid
+             (setf (gethash (quote ,name) *config-vars*)
+                   (make-config-info :name (quote ,name) :default ,default-value
+                                     :validator ,func :doc ,documentation))
+             (error 'invalid-datum-error :place (quote ,name) :value ,default-value))))))
+
+(defmacro define-config-parameter (name default validator documentation)
```
sdilts, line 67: Using keyword parameters here might be nice to allow for future expansion.

sdilts (commented):
Here are some potential improvements / problems with the provided implementation.

config-system.lisp
```diff
@@ -0,0 +1,157 @@
...
+    (mapcar #'%full-symbol-string
+            (remove-if-not (lambda (x) (string-equal (symbol-name x) name))
+                           (mapcar #'config-info-name (hash-table-values table))))))
+
+(defun get-configuration-info (setting-name)
+  "Find the info for the config variable stored in the symbol SETTING-NAME."
+  (declare (type symbol setting-name))
+  (gethash setting-name *config-vars*))
+
+(defmacro %set-configuration (setting-name value)
```
szos, line 106: I think it would be a good idea to store/dispatch based not only on the symbol but on the symbols package as well. This macro and the underlying hash table don't appear to currently support this. I solved this by making the hash table use `#'equalp` and using `(package . symbol)` for the hash key. What do you think of this? alternatively one could use the full name of the symbol as the key. This work is already partially done within `%full-symbol-string`.

config-system.lisp
szos: I agree with using keywords - I think it would be useful for things like changing the default database(s) or providing a different test function for `define-config-enum`, for example.

config-system.lisp
szos: What has been recommended to me (i think by puercopop) was to avoid using `get-` in getters. I think a name like `all-config-info` would be better. As mentioned in another response, this function could also benefit from a database optional or key variable, if we go that route.

szos (commented):
Ive added some feedback regarding features and how things are stored. Hope its useful :smile:

config-system.lisp
sdilts: What you are talking about is something that [`destructring-bind`](http://clhs.lisp.se/Body/m_destru.htm) provides. In this case though, I think an optional parameter with the default of `*config-vars*` would be fine.

config-system.lisp
sdilts: Symbols from different packages are distinct items, so there is no need to do this: they will never compare to be equal. For example, while in the `cl-user` package, the form `(eql `sb-ext::%boundp '%boundp)` returns `nil`. So as long as a package isn't using a symbol from another package, symbols with the same name from different packages will be distinct keys in the hash table.

szos: Ah! thanks for explaining this! looks like I have solved a nonexistent problem :rofl:

sdilts:
I've added some functions to print information on the config variables to a stream, added some filtering options to `all-config-info`, and normalized all of the the names to use `config` rather than have some use `configure`. With those changes, I'm happy with the PR and think that the  code is ready to be merged. That being said, we should decide if we want to be able to specify different config tables instead of the default before moving forward.

szos:
I think for this system, since its only (currently) being used by stump, the ability to specify a database isnt absolutely neccessary.

config-system.lisp
PuercoPop: Yeah, in particular I won't have separate macros for parameter/var, instead use keyword argument `:reinitialize` to use either defvar or defparameter.

I would also move documentation to a keyword argument and use the ['unsplice'](https://github.com/ruricolist/serapeum/blob/master/REFERENCE.md#unsplice-form) idiom ej. `,@(when documentation (list documentation))`. Or do we want to enforce a docstring for every configuration option?

> `define-config-enum`

I think that the valid values are better left as a keyword argument not a new kind of config parameter. ej. `(defconfig theme :type '(member 'dark 'light))`

PuercoPop:
One thing I'm not clear is, what is the purpose of all the introspection machinery at this stage? If at a later date the intention is to provide a GUI for the settings the introspection can be add in the PR that does that. As long as we are using the define-config-* macro that can be introduced w/o changing the call-sites.

config-system.lisp
szos: I lean towards having documentation be optional, as by requiring it the parameters/vars that wouldnt have them would most likely just put an empty string, which feels wrong. This is purely personal opinion, not necessarily informed by logic.

sdilts:
> what is the purpose of all the introspection machinery at this stage?

I was thinking that a repl-level interface for searching and getting info for settings would be nice, and help fulfill the "user friendly" and "self documenting" requirements that were laid out in the proposal. I think having some type of interactive interface, even if it through a repl in Slime would be much better than what we currently have. Like you said however, it's not a requirement to get the basics working.

config-system.lisp
sdilts: > I wouldn't have separate macros for parameter/var, instead use keyword argument :reinitialize to use either defvar or defparameter.

Good idea!

> Or do we want to enforce a docstring for every configuration option?

This was my intention, but it looks like I'm outvoted. Considering these are settings, there should be some type of blurb to go along with it. I can see the merits of not requiring it, and like the interface of using a keyword better anyways. I'll take it out.

> I think that the valid values are better left as a keyword argument not a new kind of config parameter

That does feel more natural. So instead of specifying a validation function, you would specify a [type specifier](http://www.lispworks.com/documentation/HyperSpec/Body/04_bc.htm) that would be used to make a new type? A validation method is required though, so I'm hesitant to put that in a keyword argument, although it is pretty easy to signal an error when one isn't included.

szos: In [defconfig](https://github.com/szos/defconfig) I've been using just a quoted list of items (or unquoted if its a function to generate such a list) which gets spliced into a call to member for the `:valid-values` key argument. I like the idea of using type specifiers, but id like to have the ability to just pass in a list of acceptable values alongside it. I see this being useful for things like `*input-window-gravity*` and the like which are just a list of keywords.

side note - I would love any and all feedback on defconfig, as its my first real system with a real purpose, and there are definitely design flaws.

sdilts: The[`member`](http://www.lispworks.com/documentation/HyperSpec/Body/t_member.htm)type specifier takes care of that, as long as the values can be compared with `eql`. You also have the `satisfies` specifier, which can take any predicate function.

szos: Ah! this thread has been quite educating for me! love it :smile:

PuercoPop: > This was my intention, but it looks like I'm outvoted.

I was just asking questions :grimacing: . I may have brought up the pattern out of reflex/custom. But upon thinking about it, forcing a docstring is a valid choice to make sure we don't slack with the documentation. I say we keep it as obligatory. 

> So instead of specifying a validation function, you would specify a type specifier that would be used to make a new type?

StumpWM already has its own deftype macro, define-stumpwm-type. One way to go about it would be to extend it to support validators. But that would be on a separate PR.

PuercoPop:
> Like you said however, it's not a requirement to get the basics working.

From what we have discussed above there are a couple of design decisions that merit further discussion (and hopefully input from more people). In the interest of not loosing steam and to help keep the discussion focused I would prefer to have this merged in smaller increments. What do you about the following plan to move forward:

1. Have on PR where we only replace the calls to `defvar` and `defparameter` with the new macro w/o any change in semantics. This would be straightforward to review and merge.
2. A second PR where we introduce an API for validation to the new macro (and if it makes sense to tie it with `define-stumpwm-type`.
3. Finally a PR to introduce the introspection capabilities.


Btw I used `defconfig` because I'm a lazy typer :grimacing: (and favor that naming variant), but the initial `define-*` naming is more aligned with the current codebase. Feel free to keep the previous name. If we end up going with the shorter name I think `defcustom` would be a better choice. As StumpWM users are likely to be Emacs users and emacs uses defcustom.

Happy new year y'all :partying_face: :tada:

dmb2:
Have we decided on this vs @szos example? I'm happy to back this as long as my concerns from the original proposal are addressed.  Specifically, I think a hashtable in memory of the original values is OK, but a database on disk seems like overkill (unless I'm overlooking something obvious).

szos:
I think that, at least for the moment, we should use @sdilts's implementation or a variation of it. The [system I'm working on](github.com/szos/defconfig) is still being tweaked, and is a bit more heavyweight than strictly necessary for StumpWM. I'm also wanting to try using it in a more complicated way than the tests suite ive written, in order to see what design decisions make the most sense - thusly the API may be subject to change (though I feel fairly settled on most of it). 

I'm a fan of @PuercoPop's proposal to start small and work our way up by initially only replacing the defvar/parameter calls with the new macro(s), then adding validation, and finally introspective capabilities. To this end I think using @sdilts's implementation will be easier/more malleable.

dmb2:
fix the conflicts and I'll merge this.

szos:
Its been a little under a month with no updates here, Id like to put forward the idea that this could be done as a contrib module. I've started working on such a thing [here](https://github.com/szos/stumpwm-settings), and while it would mean writing settings names twice, it might have less impact than integrating with stumpwm proper. thoughts?

dmb2:
lets do this as a module

PR #844, opened 2021-01-16
Allow forking from within the ~/.stumpwmrc

ralt:
This PR is offering 2 commits:

- The first one exposes a new hook that runs before any threads have been spawned, as SBCL requires an application to have no threads for sb-posix:fork to work
- The second one exposes a new function to close all the resources in a child process; I had a lot of weird sessions until I realized I should close the X display in the child :)

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -42,6 +42,7 @@
           *destroy-window-hook*
           *focus-window-hook*
           *place-window-hook*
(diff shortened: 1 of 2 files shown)


dmb2:
Not sure why you want this but the PR looks good to me :)

ralt:
I want to do some cgroups-related things, which means forking all over the place :)

PR #847, opened 2021-01-27
Fix the path to the conf-rc file.

ralt:
This is currently pointing to a directory, so LOAD on it is going to
be... complicated.


ralt:
Nevermind, probe-file is making it a file, which makes it work.

Issue #811, opened 2020-09-17
How to implement dynamic tiling like dwm?


spavlovich001:
How to implement dynamic tiling in stumpwm like dwm?

equwal:
StumpWM doesn't have that yet, so your options are:
- Hack the CL yourself to add it
- Install dwm

dmb2:
@equwal neither of your options are constructive.  Please consider how someone new to our community might react to your response before you write something.  This response add nothing of material to the question.

@spavlovich001 While we don't have dynamic tiling built in yet, it is something I would like to see either as an module or as an extension to stumpwm proper.

The file `dynamic-window.lisp` was a step in this direction.  Essentially you write a driver function to arrange the windows when a new one is added to the current group.  Then you run your command with the `*new-window-hook*` and it re-arranges all the windows when is created or restored.

If you do go down this route, please share code, even if its just a gist from your .stumpwmrc.  Also, you may need to move/add where the new-window-hook is run or make a new hook specifically for calling a dynamic tiling function.  I tried this a while ago and it got a bit squirrelly and I couldn't figure out why.

szos:
While it isnt dwm, Ive implemented a set of [i3-esq autotiling functions](https://gist.github.com/szos/d05d3d11845196d62e8c66548701ec40) . I dont think its exactly the same as i3, but im really not sure, as i dont use i3.

jcguu95:
I suggest to close this issue as we have #851, which is merged into the repo.

jcguu95:
I suggest to close this issue as we have #851, which is merged into the repo.

I'm also working on a floating version of dynamic tiling - 
![Demo](https://raw.githubusercontent.com/jcguu95/stumpwm--dynamic-floating-group/master/img/demo.gif)

Issue #760, opened 2020-03-29
error when compiled with sbcl 2.0.3


emacsomancer:
When I try to launch a StumpWM compiled with sbcl 2.0.3, I get an error:

`don't know how to REQUIRE sb-clt12`

PuercoPop:
This seems to be more an error with your SBCL installation not finding contribs that with StumpWM. What happens if you try running `(require "sb-cltl2")` in a bare REPL?

emacsomancer:
```
CL-USER> (require "sb-cltl2")
NIL
```

I tried starting with a empty `init.lisp`, and then evaluating my original `init.lisp`.  It chokes at the line `(set-prefix-key (kbd "s-f"))`, with the output:

```
There is no applicable method for the generic function
  #<STANDARD-GENERIC-FUNCTION STUMPWM::SCREEN-CURRENT-GROUP (1)>
when called with arguments
  (NIL).
   [Condition of type SB-PCL::NO-APPLICABLE-METHOD-ERROR]
See also:
  Common Lisp Hyperspec, 7.6.6 [:section]

Restarts:
 0: [RETRY] Retry calling the generic function.
 1: [RETRY] Retry SLIME interactive evaluation request.
 2: [*ABORT] Return to SLIME's top level.
 3: [ABORT] abort thread (#<THREAD "worker" RUNNING {100544CBB3}>)

Backtrace:
  0: ((:METHOD NO-APPLICABLE-METHOD (T)) #<STANDARD-GENERIC-FUNCTION STUMPWM::SCREEN-CURRENT-GROUP (1)> NIL) [fast-method]
  1: (SB-PCL::CALL-NO-APPLICABLE-METHOD #<STANDARD-GENERIC-FUNCTION STUMPWM::SCREEN-CURRENT-GROUP (1)> (NIL))
  2: (SCREEN-CURRENT-WINDOW NIL)
  3: (SYNC-KEYS)
  4: (SET-PREFIX-KEY #S(KEY :KEYSYM 102 :SHIFT NIL :CONTROL NIL :META NIL :ALT NIL :HYPER NIL ...))
  5: (SB-INT:SIMPLE-EVAL-IN-LEXENV (SET-PREFIX-KEY (KBD "s-f")) #<NULL-LEXENV>)
  6: (EVAL (SET-PREFIX-KEY (KBD "s-f")))
 --more--

```

(Currently I've downgraded back down to 2.0.2.)

dmb2:
[...]

PR #851, opened 2021-02-08
Add Dynamic Window Management

szos:
This pull requests begins implementing dynamic window management that vaguely resembles DWM style window management, as requested in #811. It implements a new group type called `dynamic-group` alongside associated functions and methods for managing adding and deleting windows, and moving around within the group. This new group type departs from the `tile-group` style of window management in that not only must every window have a frame, but every window must be the top level window of its frame. It is also stricter than DWM, allowing only one master window, instead of having a master-area with multiple windows. 

I would appreciate a review of these changes, as while I am running this without any crashes or hangs, I am concerned that there are corner cases I haven't considered yet. There are also inefficiencies to consider. 

I've included a video that demonstrates how the group handles new windows, and the overflow functionality of placing extra windows into a hidden group, window cycling/rotating, swapping a stack window with the master window and selecting a window by number. While the code refers to a „stack“ it is better thought of as a circular list.

https://user-images.githubusercontent.com/39150985/107167284-c7af8580-69b8-11eb-834b-f7293b241c90.mp4

I think that for the most part, this has a solid base of functionality. I would like to change how overflow is handled, but I am unsure what the best way to handle that is. Currently, the user must manually move the window back to the dynamic group. There are also places where things get inefficient, such as a lot of pulling happening in the function `dyn-rotate-windows`. I'm not super familiar with which functions are efficient and which are inefficient in StumpWM. It may also work fine on stronger computers - my computer is rather low powered (intel celeron from 2011). 

Update: a quick recap for those new to the thread. The current issues I'm wanting feedback on are: 
- should we modify the group class hierarchy? My vote is for yes, as this would solve the issues with command validity and make groups more extensible, and I think it can be done in a way that preserves backwards compatibility (though I'd like others to give that a thought and challenge me on it as I'm certain I've forgotten something or another). 
- how to solve the problem with `toggle-always-show`? I cant figure out what is going wrong when `ŧoggle-always-show` is called, but it crashes dynamic groups.


dmb2:
I don't have time for a proper code review, but I'm excited to get this merged!

szos:
I think this should be safe to merge, especially considering all the behavior is specific to the new group type, but I want to be certain; I want to avoid breaking anyone's setup. I think, given that StumpWM has been without dynamic window management for so long, waiting for a code review isn't terribly inconvenient.

szos:
Here is a list of things that I think merit discussing: 
1. The changes to the `top-maps` function. I added a check to prevent loading the `tile-group` maps when in a dynamic group, because it prevents a hard hang when using `ẁhich-key-mode`. I dont fully understand why this problem was occuring, but I believe its to do with `dynamic-group` inheriting from `tile-group`. I chose to inherit in order to piggyback on all the methods used for tile groups. 
2. Frame splitting. I rewrote some frame splitting functions to A) take a frame to split (as opposed to using the current frame) and B) error when a frame cant be split instead of returning `nil`. This involves some code duplication and I'm wondering about the extracting the body of `split-frame` into a new function that takes a frame to split and having `split-frame` use that function and pass in `(frame-head …)`. This would let us get rid of `dyn-split-frame`.

jcguu95:
Some questions and comments.

1. There are assumptions made in the group `dynamic-group`, stronger than those of the current version. For example
  > This new group type departs from the tile-group style of window management in that not only must every window have a frame, but every window must be the top level window of its frame.

How does `dynamic-group` enforce these? 

2. Even with commands for other groups disabled, the users can always run any thing from the repl. I'd suggest `dynamic-group` to take care of the edge cases by being able to recover from an undesirable state (after `expose` is executed, for example).

3. Instead of creating a super-group that contains `manual` and `dynamic`, I'd suggest for the sake of backward compatibility that we keep this simple, at this for the moment.

szos:
The enforcement of every window being at the 'top-level' is done through `group-add-window` and `group-delete-window`. While I havent had issues with this being enforced (outside of `toggle-always-show`, but that breaks things in a different way I think) my usage is fairly simple. I think it would be useful to have some sort of enforcer function, but I'm unsure how that should be triggered. In earlier versions I had an emergency retile function, but after I solved the problems with `group-[add|delete]-window`, it was unneeded. 

I hadn't considered the repl, thanks for bringing that up. I think checking the health of the group should be done at the beginning of `group-add-window` and the end of `group-delete-window`, but are there other places it should be done? Perhaps in the post command hook? Recovering from undesirable state is, i think, best done by wiping the group clean and retiling every window. 

As far as the super-group goes, I think backwards compatibility could be preserved by retaining the current class name and having the parent be called something like `common-tile-group`, such that the hierarchy is: 
```
(defclass group () …)
(defclass common-tile-group (group) …)
(defclass tile-group (common-tile-group) …)
(defclass dynamic-group (common-tile-group) …)
```
I had chosen the name `manual-group` because it fit better (imo) with the name `dynamic-group`, and allowed all of the methods already defined upon tile-group to be unchanged. One of the reasons I think changing the group hierarchy is worthwhile is it allows us to avoid the hack in events.lisp where we handle dynamic groups specially. Without that hack or a redefined hierarchy there isn't a way to limit behavior to only manually tiling groups. I'm also trying to keep in mind future group types. If someone prefers a different style of dynamic tiling, then a new group type could be implemented which would (assumedly) want to piggyback of the predefined behavior. 

If its not too much trouble, would you mind testing out these changes? I'm certain that my usage is not universal and im probably missing some things.

jcguu95:
Somehow it doesn't load with my current config. So I have to test it without my setting.. I still couldn't find a way to load slynk, so what I had now is limited.

Under `dyn-switch-prompt-*`, choosing the window/frame to switch which happens to be at master returns a complaint, that the selected window is not in the stack. I think it should do nothing instead.

After `dyn-switch-prompt-*`, it isn't obvious to me which window/frame is to be focused. I suggest make it so that it focuses at the same "location". For example, if it has been focusing at master, then after switching it should still focus at master. If it has been focusing at the top of the stack, then after switching it should still focus at the top of the stack.

szos:
Thanks for taking time to test it out! As far as slynk goes, I have had to change the asd file to depend upon sb-cltl2 in order to load it ever since the SBCL_HOME changes came into effect. If you change the asd and run make again slynk should load (assuming that this is whats happening for you). 

as far as the ideas go - I can see how that would be odd/off. I'll change the way location is handled. I may have implemented this in the other branch, and forgotten to bring it over into this branch, but regardless it should be relatively easy to implement.

jcguu95:
Yw! 

sb-cltl2 -- I cannot spot where you added it. I've searched the directory with `rg`.. no luck. Also, the [diff function](https://github.com/stumpwm/stumpwm/pull/851/files) in this page doesn't seem to suggest any instance of `sb-cltl2`. Did I miss something?

szos:
I havent changed the asd for this PR - iirc there was a discussion about it and people were undecided. If you add the line `#:sb-cltl2` to the asd `:depends` list, and then rebuild stump, the neccessary stuff for slynk will be included in the image.

jcguu95:
Sweet! After adding `sb-cltl2` and `sb-rotate-byte` in the dependence, my configuration works. 

Some more ideas

1. When a frame/window is killed, the selected "location" also jumps.
2. It seems that by your design, when the master window is killed, the window in the bottom of the stack should come to replace. This is reasonable. However, if `exchange-direction` is used to swap windows, when the master window is killed, this behavior is messed up. 
3. Implement `dyn-next-frame` to focus on the next frame, in the clockwise direction?
4. The symbols `cl` and `ccl` could be misleading.

szos:
Wonderful! 

1. I can see how the location jumping would be frustrating. Perhaps instead of always selecting the master window, it selects the closest stack window if it was a stack window, and stays on master if it was the master window that was killed?
2. This is my bad, I changed the behavior in the other branch and never brought it over to this branch. I'll implement it on this branch as well and change `exchange-direction` to use it. 
3. Ill throw together a `dyn-rotate-focus` command and next/previous commands that use it. 
4. I'm terrible at naming things, so thanks for the suggestion! One of the reasons I used window numbers as well as frames is that there exists no function to repack frame numbers, and when a frame is removed it may leave a „hole”, so to speak, in the frame numbering list. 
5. I'm curious what I should use instead of `cl` and `ccl`? Clockwise and counterclockwise are also accepted, and my gut feeling was that rotation symbols would be clearer than directional symbols or next/prev.

jcguu95:
> 1. I can see how the location jumping would be frustrating. Perhaps instead of always selecting the master window, it selects the closest stack window if it was a stack window, and stays on master if it was the master window that was killed?

```
;; 0 for master, _n_ for "n is selected"

0 1 2 _3_ 4 5 6

kill 2 => 0 1 _2_ 3 4 5 ;; focused window remains focused, but level ups to 2
kill 3 => 0 1 2 _3_ 4 5 ;; focused window is killed, the old 4 becomes the new 3 and got focused
kill 4 => 0 1 2 _3_ 4 5

_0_ 1 2 3 4 5 6
kill 0 => _0_ 1 2 3 4 5
kill 1 => _0_ 1 2 3 4 5
kill 5 => _0_ 1 2 3 4 5
```

> 4. I'm terrible at naming things, so thanks for the suggestion! One of the reasons I used window numbers as well as frames is that there exists no function to repack frame numbers, and when a frame is removed it may leave a „hole”, so to speak, in the frame numbering list.

Hmm.. by `select-frame` I mean a function that switch focus to the selected frame, but not move the selected frame to master.

> 5. I'm curious what I should use instead of `cl` and `ccl`? Clockwise and counterclockwise are also accepted, and my gut feeling was that rotation symbols would be clearer than directional symbols or next/prev.

Perhaps just leave `clockwise` and `counterclockwise`?

szos:
Your diagram is a little confusing to me, would you mind elaborating? 

Ah that makes more sense. Side note, `fnext` and `fprev` work just fine in dynamic groups, so I think those should be used to move around instead of implementing `dyn-next-frame` & co. 

I think leaving clockwise and counterclockwise works fine; i had mostly added cl & ccl to save on typing :smile:

jcguu95:
> Your diagram is a little confusing to me, would you mind elaborating?

My bad.. sure! 

```
init = 0 1 2 _3_ 4 5 6
kill 2 => 0 1 _2_ 3 4 5 ;; focused window remains focused, but level ups to 2
```
I mean that we start with 7 frames, from the 0th to the 6th. The 0th is the master, the 1st is at the top of the stack, and the 6th is at the bottom of the stack. Suppose that the 3rd frame is currently focused.

Now kill the 2nd. The 3rd one, originally focused, remains focused. But since there's no 2 anymore, 3 got elevated to 2. thus at the end, the 2nd is focused -- but this is the same window.

> Ah that makes more sense. Side note, `fnext` and `fprev` work just fine in dynamic groups, so I think those should be used to move around instead of implementing `dyn-next-frame` & co.

I see! If the frame numbering is health-checked regularly, this would be fine!

> I think leaving clockwise and counterclockwise works fine; i had mostly added cl & ccl to save on typing

On my end, I see a list that I can autocomplete by pressing tab. So `cl <tab>` completes it to `'clockwise` and `co <tab>` completes it to `'counter-clockwise`. Also, could we make the options visible upon `dyn-cycle` is called? It isn't a problem for me now since I know that the directions are `clockwise` and `counter-clockwise`. But it would be easier and friendlier to new users.

szos:
Thanks for elaborating! I think that should be relatively easy to implement - ill take a look now. 

Currently frame numbers arent checked or repacked in any way shape or form. I suppose that would be nice to do at the end of `group-delete-window` (when its not a superfluous window or floating window). Ill look into how those numbers could be repacked according to our stack/master setup. 

I'm unsure how to make those options visible as they're command arguments. I suppose we could make the command take no arguments and use a menu instead? but that would prevent it from being used as a function or binding specific keys to cl/ccl rotation. How about just changing the prompt string to `"Clockwise or Counterclockwise:  "?

jcguu95:
> Currently frame numbers arent checked or repacked in any way shape or form. I suppose that would be nice to do at the end of `group-delete-window` (when its not a superfluous window or floating window). Ill look into how those numbers could be repacked according to our stack/master setup.

Thank you! This seems to the hardest.. `stumpwm` provides us (too) much freedom, and this somehow makes it harder to deal with ;)

> I'm unsure how to make those options visible as they're command arguments. I suppose we could make the command take no arguments and use a menu instead? but that would prevent it from being used as a function or binding specific keys to cl/ccl rotation. How about just changing the prompt string to `"Clockwise or Counterclockwise: "?

I'm not sure how to do it, but `gselect` is an example that comes in my mind. Upon `gselect` is fired, user gets a prompt that also shows a menu hinting which options are available.

szos:
Repacking frame numbers was actually way easier than I thought. They're setf-able so I implemented a generic function with a method to make the master window frame 0 and then loop through the stack giving all frames new numbers. 

I took a peek at `gselect` and it uses an optional argument with no prompt. then if the argument is provided and valid it switches to that group, and if not it calls `grouplist`, which is a full-fledged menu one can cycle through with movement keybindings. I'm still hesitant to do this for `dyn-cycle`, as I have been viewing it as more akin to `exchange-directions` than a selection command, and the prompts will never change, as opposed to `grouplist` where they will change depending on the group names.

dmb2:
I'm ready to merge this when you guys are

szos:
I think its soon ready for merging. I want to get to the bottom of the issues with windows always being shown, as at the moment it hard crashes. Ive been rather busy the last days, so i havent had many opportunities to investigate and find a fix.

7. mar. 2021 00:55:09 David <notifications@github.com>:

szos:
A couple things I'm wondering about: 
1. I'm still curious about reorganizing the group class hierarchy. I have yet to think of a way it wouldn't be backwards compatible, but I could certainly be missing something. I'd love to hear the thoughts of the maintainers on this given its such a large change. It would remove the need for odd hacks to handle the difference between tile/dynamic groups as inheritance doesn't quite work with the current hierarchy. (on that note, i'll try rewriting everything to descend from `group` instead of `tile-group`, though that would mean a fair bit of code duplication). 
2. Ive taken a look at the way dynamic groups interact with always-on-top windows and its left me with a bad taste in my mouth. The window becomes unmanageable when always-show is toggled off in a tiling group because it gets changed to a dynamic window upon being added to the dynamic group, but never changed back. The crashing comes from the window never being removed from the group properly - it remains tracked by the group and then trying to focus it causes a hard hang.  I'm not sure that this is necessarily a problem with dynamic groups specifically, as i encounter issues with tiling and floating groups as well.

jcguu95:
> I think its soon ready for merging. I want to get to the bottom of the issues with windows always being shown, as at the moment it hard crashes.

@szos no worries! Thanks for the work!

Could you describe the problem with details, and how to reproduce? I'm happy to take a look too.

jcguu95:
I took the courage to fully move to  the `dynamic` setting, and made some tweak on my keybindings. Now it works almost fine.. but it has crashed two times in the first 30 minutes. I'm still not sure why. Here's a backtrace if it could provide any clue.

```
The value
  NIL
is not of type
  STUMPWM::FRAME
   [Condition of type TYPE-ERROR]

Restarts:
 0: [QUIT-STUMPWM] Quit Stumpwm

Backtrace:
0: (STUMPWM::DYNAMIC-GROUP-PLACE-WINDOW #<STUMPWM::DYNAMIC-GROUP {1009D81A73}> #S(DYNAMIC-WINDOW "Lickorish–Wallace theorem - Wikipedia - qutebrowser" #x260001D))
1: (STUMPWM::DYNAMIC-GROUP-ADD-WINDOW #<STUMPWM::DYNAMIC-GROUP {1009D81A73}> #S(DYNAMIC-WINDOW "Lickorish–Wallace theorem - Wikipedia - qutebrowser" #x260001D))
2: ((:METHOD STUMPWM:GROUP-ADD-WINDOW (STUMPWM::DYNAMIC-GROUP T)) #<STUMPWM::DYNAMIC-GROUP {1009D81A73}> #S(DYNAMIC-WINDOW "Lickorish–Wallace theorem - Wikipedia - qutebrowser" #x260001D)) [fast-method]
3: (STUMPWM::PROCESS-MAPPED-WINDOW #S<screen #<XLIB:SCREEN :0.0 1366x768x24 TRUE-COLOR>> #<XLIB:WINDOW :0 260001D>)
4: (STUMPWM::PROCESS-EXISTING-WINDOWS #S<screen #<XLIB:SCREEN :0.0 1366x768x24 TRUE-COLOR>>)
5: (STUMPWM::STUMPWM-INTERNAL ":0")
6: (STUMPWM:STUMPWM ":0")
7: ((LAMBDA NIL :IN "/home/jin/data/capture/stumpwm/make-image.lisp"))
8: ((FLET SB-UNIX::BODY :IN SB-IMPL::START-LISP))
9: ((FLET "WITHOUT-INTERRUPTS-BODY-1" :IN SB-IMPL::START-LISP))
10: (SB-IMPL::START-LISP)
```

I'll let you know if more evidence are found. Thank you again!

(In summary, this is not stable so far.)

szos:
Thanks for the extended testing! I can reproduce (with an error being kicked up about no window parent) but It doesnt happen when there already exists a window in the groups. 
What works: open two dynamic groups, and launch a window in each of them. In one of the groups, `fullscreen` the window, and then run `gnext-with-window`. It should work fine (at least, it does for me). This points to the issue being with the logic when no windows are present. Ill do some more digging and see if i cant hack together a fix. 

Regarding `toggle-always-show`, I'm not a fan of the way that functionality is implemented. the command isnt symmetrical in how it adds windows to groups (it calls `group-add-window` but never `group-delete-window`), it clobbers window numbers, and while trying to figure out how to make it play nice with dynamic groups, i got it to crash stump when interacting with only tile/float groups. I'm unsure of how to reproduce, but I think the issue stems from calls to `change-class`. And regarding dynamic groups, the issue is that the removal logic never gets called because `group-delete-window` isnt used to remove the window from the group. I'm not going to try and deal with the issues of `toggle-always-show` right now, and will come back to them later.

jcguu95:
Oh indeed it returns errors as you said for other groups. Though it should be fixed, it doesn't crash my `stumpwm` (weird enough). In our case, it crashes my `stumpwm` and I have to `pkill -9 stumpwm` to resolve.

Now the issue seems to be clear. Would you want to resolve it? Or I can take a look at it.

szos:
I think I'm misunderstanding you? In my testing, while using the modified `geometry-hints`, I encounter no errors or crash/hang when moving a fullscreen window to a dynamic group as per your instructions to reproduce. Could you elaborate on what happens before the crash and/or provide instructions to reproduce (if theyre different from the ones you provided above)?

also, if you dont want to kill the x session, you can also sighup stumpwm. That should be the equivalent of the `restart-hard`command.

jcguu95:
Currently I've been using for your last commit and a manual patch from PR #862, and the old problem indeed goes away. Thank you :-)

Here's a new bug, however:

1. Open two dynamic groups `g1`, `g2`. 
2. Open two windows `w11` `w12` under `g1`, and a window `w2` under `g2`.
3. Move the smaller window `w12` from `g1` to `g2`. (Error!)
4. Switch to `g2`, and run `fullscreen`. (Crash!)

szos:
Thanks for the reports! this new bug isnt triggering on my machine. Also, in step 3, i think im confused. in step 2 I open `w11`, then `w12`, pushing `w11` onto the stack. but then in step three its asserted that w12 is the smaller window. 

regardless, this bug isnt triggering for me regardless of whether its `w11` or `w12` that gets moved from `g1` to `g2`. Whats the error gets messaged in step 3?

jcguu95:
Thanks for you response!

First of, it's my bad.. I meant to say `w12` the **larger** window. Also, interestingly, though moving the window to `g2` as described creates an error, if I don't switch to `g2`, then `stumpwm` doesn't crash.

Second, to make sure that we are on the same page, I will tell you the exact version I'm using, and then tell you the errors  after that.

## Version

On top of `2b0c...`, which you commited on March 17, I added some changes from #862 and a little patch we discussed above. Here's the diff.

```
diff --git a/stumpwm.asd b/stumpwm.asd
index b2afae6..ed11283 100644
--- a/stumpwm.asd
+++ b/stumpwm.asd
@@ -17,6 +17,8 @@
                #:clx
                #:sb-posix
                #:sb-introspect
+	       #:sb-cltl2
+	       #:sb-rotate-byte
                #:cffi)
   :components ((:file "package")
                (:file "debug")
diff --git a/tile-window.lisp b/tile-window.lisp
index 270d4c0..290df9c 100644
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -120,9 +120,10 @@ than the root window's width and height."
                y (frame-y head)
                width (frame-width head)
                height (frame-height head)
-               (xlib:window-priority (window-parent win)
-                                     (window-parent (group-raised-window win-group))) :above
-               (group-raised-window (window-group win)) win))
+               (group-raised-window (window-group win)) win)
+         (when (group-raised-window win-group)
+           (setf (xlib:window-priority (window-parent win)
+                                       (window-parent (group-raised-window win-group))) :above)))
        (return-from geometry-hints (values x y 0 0 width height 0 t)))
       ;; Adjust the defaults if the window is a transient_for window.
       ((find (window-type win) '(:transient :dialog))
```

## Errors

I tested it again for a few more times, and every time the same error happens. What's trickier is that the error doesn't pass through to my `sly` under another tty `tty2`.. `stumpwm`. Instead, the error shows up in `tty1` only, with one exception. I will manually type the other things out below.

### an error that shows up in `sly` back-trace

This shows up when I move `w12` to `g2`.

```
The value
  NIL
is not of type
  STUMPWM::FRAME
   [Condition of type TYPE-ERROR]

Restarts:
 0: [QUIT-STUMPWM] Quit Stumpwm

Backtrace:
0: (STUMPWM::FRAME-HEAD #<STUMPWM::DYNAMIC-GROUP {1003C26C63}> NIL)
1: ((LAMBDA (STUMPWM::W) :IN STUMPWM::HEAD-WINDOWS) #S(DYNAMIC-WINDOW "Alacritty" #x2E00009))
2: (REMOVE-IF-NOT #<CLOSURE (LAMBDA (STUMPWM::W) :IN STUMPWM::HEAD-WINDOWS) {100283149B}> (#S(DYNAMIC-WINDOW "Alacritty" #x2E00009) #S(DYNAMIC-WINDOW "Alacritty" #x2400009)))
3: (STUMPWM::FMT-HEAD-WINDOW-LIST #S(STUMPWM:MODE-LINE :SCREEN #S<screen #<XLIB:SCREEN #1=:0.0 1366x768x24 TRUE-COLOR>> :HEAD #S(frame 0 NIL 0 0 1366 768) :WINDOW #<XLIB:WINDOW #1#:0 A0000B> :FORMAT STUM..
4: (STUMPWM:FORMAT-EXPAND ((#\d STUMPWM::FMT-MODELINE-TIME) (#\v STUMPWM::FMT-HEAD-WINDOW-LIST-HIDDEN-WINDOWS) (#\W STUMPWM::FMT-HEAD-WINDOW-LIST) (#\n STUMPWM::FMT-GROUP) (#\h STUMPWM::FMT-HEAD) (#\g ST..
5: ((:METHOD STUMPWM::MODE-LINE-FORMAT-ELT (LIST)) ("%W" "^>" "^[^5*%d^]" " ^[^2*%n^]")) [fast-method]
6: (STUMPWM::REDRAW-MODE-LINE #S(STUMPWM:MODE-LINE :SCREEN #S<screen #<XLIB:SCREEN #1=:0.0 1366x768x24 TRUE-COLOR>> :HEAD #S(frame 0 NIL 0 0 1366 768) :WINDOW #<XLIB:WINDOW #1#:0 A0000B> :FORMAT STUMPWM:..
7: (STUMPWM::UPDATE-ALL-MODE-LINES)
8: (STUMPWM::PROCESS-MAPPED-WINDOW #S<screen #<XLIB:SCREEN :0.0 1366x768x24 TRUE-COLOR>> #<XLIB:WINDOW :0 160001D>)
9: (STUMPWM::PROCESS-EXISTING-WINDOWS #S<screen #<XLIB:SCREEN :0.0 1366x768x24 TRUE-COLOR>>)
10: (STUMPWM::STUMPWM-INTERNAL ":0")
11: (STUMPWM:STUMPWM ":0")
12: ((LAMBDA NIL :IN "/home/jin/data/capture/stumpwm-testing-dynamic-window-management/make-image.lisp"))
13: ((FLET SB-UNIX::BODY :IN SB-IMPL::START-LISP))
14: ((FLET "WITHOUT-INTERRUPTS-BODY-1" :IN SB-IMPL::START-LISP))
15: (SB-IMPL::START-LISP)
```

### An error that shows up only in the screen under `tty1`

(manually typed)

```
Error In Command 'gmove 2': The value
                                           NIL
                                         is not of type
					   STUMPWM::FRAME
```

### An error that shows up after

(manually typed)

```
StumpWM Crashed With An Unhandled Error!
Copy the error to the clipboard with the 'copy-unhandled-error'
The value
    NIL
  is not of type
    STUMPWM::FRAME
Backtrace for: #<SB-THREAD:TRHEAD "main thread" RUNNING {100..
[..]
[the following seems to be similar to what I got in sly.]
```

szos:
Thanks for the info! I was able to reproduce by manually moving the window via `move-window-to-group`, wheras using `gnext-with-window` wasnt triggering this. By doing this through the repl it causes a crash exactly as you describe. Ill do some digging and edit this comment with my findings.

szos:
The fix for this was to not assume that splits are occurring in the current group. Now we pass in the group as an argument to `dyn-*split-frame`. This solves the issue, but the frame border isnt propperly drawn - I'm unsure how to fix that. In addition, theres an issue with this when theres already two+ windows in the group. The master window is set, but not shown in the frame. Ill investigate this next.

jcguu95:
Oh dear, tested it.. it's so smooth! I will keep using this setting and see if I can find some un-handled edge cases. Thanks so much for the hard work!

szos:
Thank **you** for all the testing! This would have been a much slower and difficult process without you!

@dmb2 I think, unless we want to rework the group class hierarchy (to avoid the hacks in command.lisp and events.lisp), that this is ready for merging. So far it seems like only @jcguu95 and I have been testing this, so there may be other bugs out there, but weve covered everything I can think of.

jcguu95:
@szos  thank you too! Some comments:

1. To merge, or at least to get more people involved, it might be a good idea to put the patch I gave above, as following 

 > On top of 3348..., which you commited on March 17, I added the change from #862 and a little patch we...

2. Just curious and not being pushy.. are you happy with how `dyn-group` handles `full-screen` now? IIRC, in `dwm`, while in full screen mode, all screens (here frames) are supposed to be full, and the user should be able to switch between them. However, that's not the case here.. which could be somehow confusing.

Take the following as example. Open and switch to `dyn-group` `g1`. Open three windows `w1, w2, w3`. Full screen `w3`. Switch focus to `w2`. Now the layout is confusing already. Full screen `w2`.. which makes it more confusing.

szos:
regarding 1, That patch is currently open as a PR as it affects more than just dynamic groups. I suspect that it will be merged alongside or before this is merged, but I'm unsure. I dont think its a terribly large issue. 

Regarding 2, I dont really have an opinion. One reason you found so many bugs regarding fullscreen windows is that I dont really use them, and as such they were woefully undertested by me. After running your example, I can absolutely see that being confusing. But as I dont really use fullscreen windows all that much I'm unsure what the best option would be. What do you think would be best here? My first impression is that it should function the same as in tiling groups. If you open up a tiling group with two frames, fullscreen one frames window, then call `fnext`, youll find the same behavior. I think handling fullscreen windows differently should be done in a discussion/PR regarding fullscreen specifically, but am open to other ideas.

jcguu95:
1. Got it.
2. How hard is it to advice `full-screen` such that when it's called in a `dyn-group`, all frames in that group is also full-screened? And vice versa?

szos:
There isnt a default advice system in CL. I've written one, but I dont think it makes sense to add yet another dependency to StumpWM for that. 

Fullscreen currently works by telling the X server to make the window full screen by changing an X11 property for the window, so I suppose we could map `update-fullscreen` over all windows when were in a dynamic group. But is this what is wanted? I think it would be better to instead introduce a new command (call it `fullscreen-all-windows-in-group` for now) that does this, as when a user fullscreens a window, I think its fair to assume they only want that window to be made fullscreen.

jcguu95:
You're right! That should be easy enough for the users! I'd say besides potential flaws that might come up within the next three days, I'm very happy with the setting now! Thanks :-)

jcguu95:
I have a good news and a bad news.

The good news is that the experience so far has been smooth for the past three days, with one exception.

The bad news is.. that exception would be hard to tackle. It happens when `zoom` tries to open a meeting room. I don't know what the internal is, but it seems that many windows of mixed types are called simultaneously. At the end, in `dynamic group` it crashes stumpwm.. It does not happen in the tiling group.

This only suggests that there are still some edge cases. But this time, it's hard to catch as we don't understand the internal of `zoom`.

szos:
Thanks for catching that! I dont use zoom - its not in my distros repos and I dont use it for work/school - so im unable to debug or diagnose this problem. One workaround would be to hard code a rule sending all zoom windows to a tiling/floating group, but i dont know what window classes (or other info) to dispatch on for that. If you could capture the zoom window classes ill push a fix in a couple days.

24. mar. 2021 19:03:40 jcguu95 ***@***.***>:

jcguu95:
Thanks for offering the help! Right now I just have to remember to open them in the traditional tiling (otherwise the whole x session should be killed and restarted!). So not a big problem for me.

But, just curious, how does one get the window classes for zoom?

tychoish:
I use zoom a bunch (tiling only,) and while it doesn't instantly crash Stump, it doesn't always play well with stump (sometimes zoom fails to open, or has some weird window-opens-follows-pointer behavior that I don't really get), though the latest version seems to be a little better.

jcguu95:
@tychoish I had that issue. Now before I launch zoom, I always `pkill -9 zoom` before-hand. With this trick it never  failed me in tiling groups.

In short, and what's related is that, `zoom` is troublesome..

jcguu95:
Another week of testing without any crashing! Well done!

---

Not sure if it's of your interest. But I imagine it wouldn't be too hard for you. 

1. For now the main frame is always on the left. Is it possible to let user choose if it should be on the right, or top/bottom?
2. Can we let the user to customize the width of the master frame? Sometimes I need it larger, sometimes smaller.

szos:
Thanks for the continued testing!

Regarding 1, im sure its possible. Right now the frame we use is hard coded to be on the left, but it should be able to be used on the right/top/bottom as well. Im not in a position where i have a lot of time - i just moved and am still getting over jetlag, among other things, but ill take a look when i have time and see what has to happen to make this customizable.

Regarding 2, this is already implemented. One can use iresize to modify the size of the master frame (though for stack frames any changes will get overrided when we balance them), and can control the size of the initial split with `*dynamic-group-master-split-ratio*`.

30. mar. 2021 06:00:41 jcguu95 ***@***.***>:

> Another week of testing without any crashing! Well done!
> 
> ----------------------------------------
> 
> Not sure if it's of your interest. But I imagine it wouldn't be too hard for you.
> 
- * > For now the main frame is always on the left. Is it possible to let user choose if it should be on the right, or top/bottom?
- * > Can we let the user to customize the width of the master frame? Sometimes I need it larger, sometimes smaller.

szos:
An update regarding 1: this is a bit more work than i thought it would be. The assumption that frame zero is always the master runs deep in this implementation, which dictates the left or top frame being the master. It will take a little time to remove this assumption.

30. mar. 2021 06:00:41 jcguu95 ***@***.***>:

> Another week of testing without any crashing! Well done!
> 
> ----------------------------------------
> 
> Not sure if it's of your interest. But I imagine it wouldn't be too hard for you.
> 
- * > For now the main frame is always on the left. Is it possible to let user choose if it should be on the right, or top/bottom?
- * > Can we let the user to customize the width of the master frame? Sometimes I need it larger, sometimes smaller.

szos:
Changing the layout has been implemented! it is tracked through a class allocated slot, with a normal slot for overriding the default behavior.

dmb2:
@szos if you'd like commit access to keep refining this let me know.  The general rule is: for small changes/bugfixes your fine to merge but for PRs like this one, I'd like to review/give my opinion.

jcguu95:
OH GREAT :-) Now we need a little documentation for this wonderful branch. <3

szos:
@dmb2 Wow thanks for the show of trust! While I would love commit access I'm a little hesitant - I still dont know git super well, and dont want to mess anything up. That being said I think it would be useful to avoid a new PR for every bugfix that comes along. I suppose yes, I would appreciate commit access to deal specifically with dynamic group bugfixes and implementation details. If you have any best practices regarding git I would love to hear them. 

@jcguu95 There is some minor documentation in `stumpwm.texi.in`, but it most definitely needs expanding. I'm unsure where in the manual this information should be placed (personally I would organize the manual differently) so If you have any ideas let me know.

jcguu95:
Oh yeah, it seems that the work is almost done! I'm happy to write the doc when I know what's going on surely. Before that, some comments as usual ;-)

1. With an empty config, it works well. But as I mentioned above, for my config to work, I need to put `:sb-cltl2` and `:sb-rotate-byte` in `stumpwm.asd`. 

  + This might need to be addressed in the documentation.
  + Is there anyway to fix by adding some lines in the user config instead, rather in the source?

2. In the freshest commits, you introduced the function `change-layout`. IMO it should be called `dyn-change-layout` just to align with the other dynamic-group functions.

3. `*dynamic-group-master-split-ratio*` works nicely so far. But currently it doesn't change the layout on live. In fact, the user must remove frames in the group and add them back to see effect. If such effect can be taken on live, the user can easily add a function that increases/decreases the value by 5%, and see it changes on live. Hopefully this is not too much more work though.. I'm aware you are busy - but if you can lay any hint on this others might find it easier to implement :)

szos:
Thanks for offering to write the docs. Ive got a fair bit added to the manual already, with most functions/commands/variables being documented.

Regarding 1, this is unrelated to dynamic groups. :sb-cltl2 is required for slynk/swank, though im unsure what :sb-rotate-byte is used for in you config. While this could be addressed in documentation, i think that is a separate issue. Im unsure how it can be done purely in the user config, but it most likely has to do with the SBCL_HOME environment variable.

Regarding 2, i agree and will implement the change asap.

Regarding 3, we cant watch variables to trigger functions when they are changed. So this functionality needs to be implemented as a function which calls retile after setting the variable a la dynamic-group-master-location-dwim.

2. apr. 2021 04:48:57 jcguu95 ***@***.***>:

> Oh yeah, it seems that the work is almost done! I'm happy to write the doc when I know what's going on surely. Before that, some comments as usual ;-)
> 
- * > With an empty config, it works well. But as I mentioned above, for my config to work, I need to put :sb-cltl2 and :sb-rotate-byte in stumpwm.asd.
* > This might need to be addressed in the documentation.
* > Is there anyway to fix by adding some lines in the user config instead, rather in the source?
- * > 
> In the freshest commits, you introduced the function change-layout. IMO it should be called dyn-change-layout just to align with the other dynamic-group functions.
> 
- * > *dynamic-group-master-split-ratio* works nicely so far. But currently it doesn't change the layout on live. In fact, the user must remove frames in the group and add them back to see effect. If such effect can be taken on live, the user can easily add a function that increases/decreases the value by 5%, and see it changes on live. Hopefully this is not too much more work though.. I'm aware you are busy - but if you can lay any hint on this others might find it easier to implement :)

szos:
@jcguu95 check out issue #760 regarding SBCL_HOME. The solution is to add `export SBCL_HOME=/path/to/sbcl/files` to your xinitrc (assuming you use startx to start stumpwm. For me this becomes `export SBCL_HOME=/usr/lib/sbcl` this should allow you to remove #:sb-cltl2 and #:sb-rotate-byte as dependencies.

2. apr. 2021 05:05:26 Nathan Shostek ***@***.***>:

> Thanks for offering to write the docs. Ive got a fair bit added to the manual already, with most functions/commands/variables being documented.
> 
> Regarding 1, this is unrelated to dynamic groups. :sb-cltl2 is required for slynk/swank, though im unsure what :sb-rotate-byte is used for in you config. While this could be addressed in documentation, i think that is a separate issue. Im unsure how it can be done purely in the user config, but it most likely has to do with the SBCL_HOME environment variable.
> 
> Regarding 2, i agree and will implement the change asap.
> 
> Regarding 3, we cant watch variables to trigger functions when they are changed. So this functionality needs to be implemented as a function which calls retile after setting the variable a la dynamic-group-master-location-dwim.
> 
> 2. apr. 2021 04:48:57 jcguu95 ***@***.***>:
> 
>> Oh yeah, it seems that the work is almost done! I'm happy to write the doc when I know what's going on surely. Before that, some comments as usual ;-)
>> 
- * >> With an empty config, it works well. But as I mentioned above, for my config to work, I need to put :sb-cltl2 and :sb-rotate-byte in stumpwm.asd.
* >> This might need to be addressed in the documentation.
* >> Is there anyway to fix by adding some lines in the user config instead, rather in the source?
- * >> 
>> In the freshest commits, you introduced the function change-layout. IMO it should be called dyn-change-layout just to align with the other dynamic-group functions.
>> 
- * >> *dynamic-group-master-split-ratio* works nicely so far. But currently it doesn't change the layout on live. In fact, the user must remove frames in the group and add them back to see effect. If such effect can be taken on live, the user can easily add a function that increases/decreases the value by 5%, and see it changes on live. Hopefully this is not too much more work though.. I'm aware you are busy - but if you can lay any hint on this others might find it easier to implement :)

jcguu95:
Thanks for pointing our #760! I will take a look.

A bad news.. after switching to `commit a5ab063398cbea2dbfe711cacd5623b0c5779bf4`, dynamic group gets fragile again.. It has #862 and #851 merged.. To reproduce

1. Switch to a dynamic group.
2. Open a window, and close it (so no more window for that group).
3. Unhandled error! Message below.

It doesn't crash stumpwm. But it comes out by itself every now and then, which is so annoying that I'd just kill `stumpwm`.

Here's the unhandled error message:

```
StumpWM Crashed With An Unhandled Error!

Attempt to use closed display #<XLIB:DISPLAY :0 (The X.Org Foundation R12010000)>
Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {1002138173}>
0: (STUMPWM::BACKTRACE-STRING)
1: ((:METHOD STUMPWM::HANDLE-TOP-LEVEL-CONDITION (SERIOUS-CONDITION)) #<XLIB:CLOSED-DISPLAY {100847F243}>) [fast-method]
2: (SB-KERNEL::%SIGNAL #<XLIB:CLOSED-DISPLAY {100847F243}>)
3: (ERROR XLIB:CLOSED-DISPLAY :DISPLAY #<XLIB:DISPLAY :0 (The X.Org Foundation R12010000)>)
4: (XLIB::X-ERROR XLIB:CLOSED-DISPLAY :DISPLAY #<XLIB:DISPLAY :0 (The X.Org Foundation R12010000)>)
5: (XLIB:TEXT-WIDTH #<XLIB:FONT fixed :0 8388609> " 3:51" :START 0 :END NIL :TRANSLATE #<FUNCTION STUMPWM::TRANSLATE-ID>)
6: (STUMPWM::RENDERED-STRING-SIZE "^>^[^5* 3:51^] ^[^2*1^]" #S(STUMPWM::CCONTEXT :SCREEN #S<screen #<XLIB:SCREEN #1=:0.0 1366x768x24 TRUE-COLOR>> :WIN #<XLIB:WINDOW #1#:0 80000B> :PX #<XLIB:PIXMAP #1#:0 80000D> :GC #<XLIB:GCONTEXT #1#:0 8388620> :DEFAULT-FG 8355711 :DEFAULT-BRIGHT 12566463 :DEFAULT-BG 3355443 :FG 5 :BG :RESET :BRIGHTP NIL :REVERSEP NIL :COLOR-STACK ((5 :RESET NIL NIL #2=#<XLIB:FONT fixed #1#:0 8388609>) (5 :RESET NIL NIL #2#) (NIL NIL NIL NIL #2#)) ...) NIL)
7: (STUMPWM::RENDERED-SIZE ("^>^[^5* 3:51^] ^[^2*1^]") #S(STUMPWM::CCONTEXT :SCREEN #S<screen #<XLIB:SCREEN #1=:0.0 1366x768x24 TRUE-COLOR>> :WIN #<XLIB:WINDOW #1#:0 80000B> :PX #<XLIB:PIXMAP #1#:0 80000D> :GC #<XLIB:GCONTEXT #1#:0 8388620> :DEFAULT-FG 8355711 :DEFAULT-BRIGHT 12566463 :DEFAULT-BG 3355443 :FG 5 :BG :RESET :BRIGHTP NIL :REVERSEP NIL :COLOR-STACK ((5 :RESET NIL NIL #2=#<XLIB:FONT fixed #1#:0 8388609>) (5 :RESET NIL NIL #2#) (NIL NIL NIL NIL #2#)) ...))
8: (STUMPWM::RESIZE-MODE-LINE #S(MODE-LINE :SCREEN #1=#S<screen #<XLIB:SCREEN #2=:0.0 1366x768x24 TRUE-COLOR>> :HEAD #S(frame 0 NIL 0 0 1366 768) :WINDOW #3=#<XLIB:WINDOW #2#:0 80000B> :FORMAT *SCREEN-MODE-LINE-FORMAT* :POSITION :BOTTOM :CONTENTS "^>^[^5* 3:51^] ^[^2*1^]" :CC #S(STUMPWM::CCONTEXT :SCREEN #1# :WIN #3# :PX #<XLIB:PIXMAP #2#:0 80000D> :GC #<XLIB:GCONTEXT #2#:0 8388620> :DEFAULT-FG 8355711 :DEFAULT-BRIGHT 12566463 :DEFAULT-BG 3355443 :FG 5 :BG :RESET :BRIGHTP NIL :REVERSEP NIL :COLOR-STACK ((5 :RESET NIL NIL #4=#<XLIB:FONT fixed #2#:0 8388609>) (5 :RESET NIL NIL #4#) (NIL NIL NIL NIL #4#)) ...) :HEIGHT 17 :FACTOR 751/768 :MODE :STUMP))
9: (STUMPWM::REDRAW-MODE-LINE #S(MODE-LINE :SCREEN #1=#S<screen #<XLIB:SCREEN #2=:0.0 1366x768x24 TRUE-COLOR>> :HEAD #S(frame 0 NIL 0 0 1366 768) :WINDOW #3=#<XLIB:WINDOW #2#:0 80000B> :FORMAT *SCREEN-MODE-LINE-FORMAT* :POSITION :BOTTOM :CONTENTS "^>^[^5* 3:51^] ^[^2*1^]" :CC #S(STUMPWM::CCONTEXT :SCREEN #1# :WIN #3# :PX #<XLIB:PIXMAP #2#:0 80000D> :GC #<XLIB:GCONTEXT #2#:0 8388620> :DEFAULT-FG 8355711 :DEFAULT-BRIGHT 12566463 :DEFAULT-BG 3355443 :FG 5 :BG :RESET :BRIGHTP NIL :REVERSEP NIL :COLOR-STACK ((5 :RESET NIL NIL #4=#<XLIB:FONT fixed #2#:0 8388609>) (5 :RESET NIL NIL #4#) (NIL NIL NIL NIL #4#)) ...) :HEIGHT 17 :FACTOR 751/768 :MODE :STUMP) NIL)
10: (STUMPWM::UPDATE-ALL-MODE-LINES)
11: (SB-KERNEL:%MAP-FOR-EFFECT-ARITY-1 #<FUNCTION STUMPWM::EXECUTE-TIMER> (#S(STUMPWM::TIMER :TIME 145090064 :REPEAT 10 :FUNCTION STUMPWM::UPDATE-ALL-MODE-LINES :ARGS NIL)))
12: ((:METHOD IO-LOOP (STUMPWM::SBCL-IO-LOOP)) #<STUMPWM::SBCL-IO-LOOP {10084539D3}> :DESCRIPTION "StumpWM") [fast-method]
13: (STUMPWM::STUMPWM-INTERNAL-LOOP)
14: (STUMPWM::STUMPWM-INTERNAL ":0")
15: (STUMPWM ":0")
16: ((LAMBDA NIL :IN "/home/jin/data/capture/stumpwm/make-image.lisp"))
17: ((FLET SB-UNIX::BODY :IN SB-IMPL::START-LISP))
18: ((FLET "WITHOUT-INTERRUPTS-BODY-1" :IN SB-IMPL::START-LISP))
19: (SB-IMPL::START-LISP)
```

jcguu95:
Floating support is given in #885.

Issue #606, opened 2019-04-29
Timers don't run while entering a key sequence


goose121:
During the time when a key sequence is being entered, `handle-keymap` [blocks on `read-key-no-modifiers`](https://github.com/stumpwm/stumpwm/blob/master/events.lisp#L202), which prevents any active timers from running until the key sequence is completed, and may have other effects seeing as it's blocking up the I/O loop. One easily visible example of this is the timer for updating the mode line; while a key sequence is being entered, the mode line will not update

equwal:
The solution to this is to rewrite the restarts to use [SBCL's threaded version](http://www.sbcl.org/manual/#Timers).

sdilts:
I fixed this in Mahogany by not blocking when a key sequence is being entered. You can find the code for it in [kmap.lisp](https://github.com/stumpwm/mahogany/blob/master/lisp/keyboard/kmap.lisp) and [input.lisp](https://github.com/stumpwm/mahogany/blob/master/lisp/input.lisp). The idea is that if a keybinding is active, we store the state of the key sequence until a command is found or the user types an invalid sequence. The code is almost verbatim from Stumpwm's key handling code, just broken out so it doesn't block the main  thread.

The main blocker for the code to be directly imported is that it builds the list of active maps once when the current state is being initialized to avoid consing a bunch of stuff on every keypress. Stumpwm builds the list of maps on every keypress [because the active keymaps rely on at least one dynamic variable that users can set at any time](https://github.com/stumpwm/stumpwm/blob/baf5df7467553c52ec925a3b44ab24186f08cd9b/events.lisp#L218-L251). Mahogany gets around this by guarding the keymap list with a function, which allows it to re-initialize the keystate object whenever the list gets updated.

PR #853, opened 2021-02-17
Add an idle delay to which-key-mode

landakram:
This changeset makes which-key-mode more similar to emacs. Instead of popping up
immediately upon a keypress, `*which-key-idle-delay*`can be specified to delay
presentation of the which-key window. In addition, `*which-key-secondary-idle-delay*`
can be used to specify the delay for additional presentations of the which-key window
when it is already showing e.g. when navigating into another keymap.

This changeset also adds some light customization of the keybindings presentation. The
`*display-bindings-formatter*` parameter can be set to change presentation of
keybindings in the help window. By default, the keybindings are left unsorted to
maintain backwards compatible behavior. Two other formatting functions,
`alphabetical-command-bindings-formatter` and `alphabetical-key-bindings-formatter`
can be used to sort keybindings by command name or key sequence respectively. An
additional parameter, `*which-key-bindings-formatter*` can be used to adjust the
which-key keybindings format independently of the C-h format.

trivial-timers and bordeaux-threads have been added for portable versions of the timer 
functionality in sb-ext and threading in sb-thread. Threading was needed for the same
reason as https://github.com/stumpwm/stumpwm/issues/606, which is that key-sequences block the main thread and so stumpwm's 
timers don't run. Instead we use threaded timers and a mutex around displaying messages 
to achieve thread safety.

I've been running this for a few days now without errors, but it's all been manual testing.
I also understand if we don't want to bring in more dependencies (though timers and 
threads seems sensible) -- let me know if you think this belongs in stumpwm-contrib rather
than stumpwm.

FWIW, I've been running with this config:

```lisp
(setf *which-key-idle-delay* 0.5)
(setf *which-key-secondary-idle-delay* 0.0)
(setf *which-key-bindings-formatter* #'alphabetical-command-bindings-formatter)
(setf *display-bindings-formatter* #'alphabetical-command-bindings-formatter)

(which-key-mode)
```

### DIFF
--- a/help.lisp
+++ b/help.lisp
@@ -54,11 +54,44 @@
                                   (:right (format nil "~a~a~a" (if (= c 0) "" padstr) len s)))))))
     (apply 'mapcar 'concat (or cols '(nil)))))
 
+(defun sort-flattened-bindings (keymaps sort-pred)
+  "Returns all bindings in KEYMAPS in a flattened list, globally sorted with SORT-PRED."
+  (let ((all-bindings (flatten (mapcar #'kmap-bindings keymaps))))
+    (sort all-bindings sort-pred)))
+
+(defun sort-by-alphabetical-command-p (b1 b2)
+  "Sort predicate that sorts bindings alphabetically by command name."
+  (not (null (string-lessp (binding-command b1) (binding-command b2)))))
+
+(defun sort-by-alphabetical-key-p (b1 b2)
+  "Sort predicate that sorts bindings alphabetically by key."
+  (not (null (string-lessp (print-key (binding-key b1)) (print-key (binding-key b2))))))
+
+(defun format-binding (binding)
+  (format nil "^5*~5a^n ~a" (print-key (binding-key binding)) (binding-command binding)))
+
+(defun unsorted-bindings-formatter (keymaps)
+  "Returns all bindings in KEYMAPS in a flattened, unsorted list."
+  (let ((bindings (flatten (mapcar #'kmap-bindings keymaps))))
+    (mapcar #'format-binding bindings)))
+
+(defun alphabetical-command-bindings-formatter (keymaps)
+  "Returns all bindings in KEYMAPS in a flattened list, globally sorted by command name."
+  (let ((bindings (sort-flattened-bindings keymaps #'sort-by-alphabetical-command-p)))
+    (mapcar #'format-binding bindings)))
+
+(defun alphabetical-key-bindings-formatter (keymaps)
+  "Returns all bindings in KEYMAPS in a flattened list, globally sorted by keys."
+  (let ((bindings (sort-flattened-bindings keymaps #'sort-by-alphabetical-key-p)))
+    (mapcar #'format-binding bindings)))
+
+(defvar *display-bindings-formatter* #'unsorted-bindings-formatter
+  "A function taking a list of keymaps and returning a list of strings suitable for
+presentation as columnized data in the keybindings help menu.")
+
 (defun display-bindings-for-keymaps (key-seq &rest keymaps)
   (let* ((screen (current-screen))
-         (data (mapcan (lambda (map)
-                         (mapcar (lambda (b) (format nil "^5*~5a^n ~a" (print-key (binding-key b)) (binding-command b))) (kmap-bindings map)))
-                       keymaps))
+         (data (funcall *display-bindings-formatter* keymaps))
          (cols (ceiling (1+ (length data))
                         (truncate (- (head-height (current-head)) (* 2 (screen-msg-border-width screen)))
                                   (font-height (screen-font screen))))))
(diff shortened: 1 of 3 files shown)


message-window.lisp
```diff
@@ -252,42 +252,45 @@ When NEW-ON-BOTTOM-P is non-nil, new messages are queued at the bottom."
...
-          ;; any left over timers need to be canceled.
-          (when (timer-p *message-window-timer*)
-            (cancel-timer *message-window-timer*)
-            (setf *message-window-timer* nil))
-          (reset-message-window-timer)))
-    (push-last-message screen strings highlights)
-    (xlib:display-finish-output *display*)
-    (dformat 5 "Outputting a message:~%~{        ~a~%~}" strings)
-    (apply 'run-hook-with-args *message-hook* strings)))
+  (bordeaux-threads:with-lock-held (*message-lock*)
```
landakram, line 260: The only change here is wrapping everything in `bordeaux-threads:with-lock-held`

szos:
While I haven't looked at this enough to have an opinion on the PR itself, I dont think introducing new dependencies for the sake of portability is worthwhile, as StumpWM only supports SBCL and (as far as i know) has no intention to support other implementations.

tychoish:
To be honest, I'm kind of surprised that we weren't already using bordeux-threads, and while I think I'm generally conservative on the "reduce dependency footprint," I don't find myself particularly troubled by these dependencies specifically (and would greatly prefer that we used bt for threading things, even if we end up being pretty sbcl-focused in practice.)

PuercoPop:
> StumpWM only supports SBCL and (as far as i know) has no intention to support other implementations.

That is correct. Another thing is that we are already using the [sb-thread API](https://github.com/stumpwm/stumpwm/pull/853) so I would prefer to stay consistent and keep using sb-thread instead of having some parts of the codebase use sb-thread and another use bordeux-thread. Similarly we have a timer API (timers.rb) so I'm against using trival-timers.

help.lisp
```diff
@@ -54,11 +54,44 @@
                                   (:right (format nil "~a~a~a" (if (= c 0) "" padstr) len s)))))))
     (apply 'mapcar 'concat (or cols '(nil)))))
 
+(defun sort-flattened-bindings (keymaps sort-pred)
+  "Returns all bindings in KEYMAPS in a flattened list, globally sorted with SORT-PRED."
+  (let ((all-bindings (flatten (mapcar #'kmap-bindings keymaps))))
```
PuercoPop, line 59: Would `alexandria:mappend` (akka flatmap) be an appropiate replacement to (flatten (mapcar ...))?

help.lisp
```diff
@@ -54,11 +54,44 @@
...
+    (sort all-bindings sort-pred)))
```
PuercoPop, line 60: `sort` is a destructive operation, lets copy the list to before sorting to avoid mutating the arguments. 

```suggestion
    (sort (copy-list all-bindings) sort-pred)))
```

help.lisp
```diff
@@ -54,11 +54,44 @@
...
+
+(defun sort-by-alphabetical-command-p (b1 b2)
+  "Sort predicate that sorts bindings alphabetically by command name."
+  (not (null (string-lessp (binding-command b1) (binding-command b2)))))
```
PuercoPop, line 64:
```suggestion
  (string-lessp (binding-command b1) (binding-command b2)))
```

`string-lessp` already returns a [generalized boolean](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_g.htm#generalized_boolean) so it is not necessary to wrap it with `(not (null ...))`

landakram:
Thanks for the feedback! I'll look over the comments and make some revisions. I'm a CL newbie so it's helpful to be pointed to useful functions in alexandria, etc. that I probably missed

To address some questions:

@szos said:

> would you mind explaining why the message lock is used within echo-string-list instead of at a higher level? What happens if there is a message with no timeout being displayed, and then a new message is displayed?

The lock exists to prevent crashing I experienced in practice. I couldn't determine the exact issue, but it had something to do with concurrent use of xlib. I haven't experienced any crashing since adding the lock. I don't have a particular reason for locking in echo-string-list, other than it was the lowest level stump code that wraps the xlib calls and which solved the issue. It also covers all "message-like" output e.g. `message`, `message-no-timeout`, `echo-string` and `echo-string-list`, which are all used directly in various parts of the stumpwm codebase.

If a message with no timeout is displayed and then a new message is displayed, the message with no timeout is cleared (or appended, depending on `*queue-messages-p*`). This is the same behavior as before this PR, as neither `message` nor `message-no-timeout` doesn't block, but rather, `message` uses a stumpwm timer to clear the message later.

@tychoish said:

> I don't find myself particularly troubled by these dependencies specifically (and would greatly prefer that we used bt for threading things, even if we end up being pretty sbcl-focused in practice.)

and @PuercoPop said:

> we are already using the sb-thread API so I would prefer to stay consistent and keep using sb-thread instead of having some parts of the codebase use sb-thread and another use bordeux-thread. 

If we're already using sb-thread elsewhere, I'm fine to remove the portable dependencies and continue using SBCL extensions. 

> Similarly we have a timer API (timers.rb) so I'm against using trival-timers.

I linked a related issue for this in the PR description, but stumpwm timers run in the main thread and are therefore blocked from executing during partial key sequences. So they couldn't work for the purposes of which-key, which displays mid key-sequence. I'm happy to use `sb-ext` timers directly rather than trivial-timers, but unfortunately I don't think stumpwm timers are going to work here without making them threaded, which I'm pretty afraid would break a lot of things.

dmb2:
Use sb-thread and address @PuercoPop's code review, then this is good to merge.

dmb2:
fix merge conflicts and I'll merge.

landakram:
Sorry, I started a new job and haven't had time to come back to this yet. Will @ you when I fix it up and think it's ready to go.

dmb2:
I tried to fix merge-conflicts and it broke more than I wanted to deal with.  I'm going to close this for now, but if you can use sbcl primitives I'm happy to merge.

teutonicjoe:
This would be so nice to have... will see if I can get it to work whenever I have more free time, and dive into StumpWM

PR #857, opened 2021-02-23
Allow refining input completion candidates

landakram:
Hello again!

Completion candidates during completing-read can now be refined in a customizable
way. The `*input-refine-candidates-fn*` can be set to customize candidate
refinement. This parameter can be used to filter or sort candidates in novel ways.

By default, completion candidates are refined by prefix in unsorted order to maintain
backwards compatibility. Another refinement function, `input-refine-regexp`, is
implemented, which provides similar behavior to select-from-menu. That is, it refines
candidates by splitting the input string by space and treating each word as a regexp
to be matched.

I attached a screencast below of what the `input-refine-regexp` function looks like when
run with `colon`. It  looks a lot like `select-from-menu`, but without the drawbacks
of losing the command history. To be honest, I'd love to slowly converge the featuresets
of single selection menus and completing read, as they are quite similar. 

https://user-images.githubusercontent.com/700616/109921066-14147b00-7c70-11eb-9b81-8d3c35a6627f.mp4

I also took the opportunity to refactor the completions display logic to ensure it uses
the same set of candidates that are used to actually do the completion itself. The separate
implementations had diverged a little (in insignificant ways, as far as I can tell).

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -122,6 +122,23 @@ and complete the input by mutating it."))
   (make-instance 'input-completion-style-unambiguous
                  :display-limit display-limit))
 
+(defun input-refine-prefix (str candidates)
+  (remove-if-not (lambda (elt)
+                   (when (listp elt)
+                     (setf elt (car elt)))
+                   (and (<= (length str) (length elt))
+                        (string= str elt
+                                 :end1 (length str)
+                                 :end2 (length str))))
+                 candidates))
+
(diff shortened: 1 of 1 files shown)


szos:
Would you mind either explaining whats happening in the video or re-recording with a larger font size? I cant make out any of the text.

landakram:
Sorry about that, I just uploaded a zoomed-in screencast.

PR #862, opened 2021-03-16
Check for nil value before calling `window-parent` within `geometry-hints`

szos:
While testing #851 we found a bug where if a fullscreen window is moved to a group that has never had a window in it, StumpWM tries to call `window-parent` on `nil`. This is due to the way the function `geometry-hints` handles fullscreen windows. This pull request moves the offending code into a `when` statement dependent upon `group-raised-window` being truthy.

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -120,9 +120,10 @@ than the root window's width and height."
                y (frame-y head)
                width (frame-width head)
                height (frame-height head)
-               (xlib:window-priority (window-parent win)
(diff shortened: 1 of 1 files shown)


jcguu95:
To quote how exactly to reproduce as in #851 

> [..] to reproduce: open two new tiling groups. open a window in one of them. Fullscreen that window. run gnext-with-window to move it to the second tiling group. encounter error. [..]

jcguu95:
Great! Thanks for the work :)

PR #869, opened 2021-04-11
remove unnecessary uses of multiple-value-list

defaultxr:
`cl:nth-value` can be used to access an nth return value. The `(first (multiple-value-list ...))` in `timers.lisp` is not needed since functions like `/` ignore additional return values by default.

### DIFF
--- a/color.lisp
+++ b/color.lisp
@@ -195,7 +195,7 @@ If COLOR isn't a colorcode a list containing COLOR is returned."
                     (screen-color-map-bright screen)
                     (screen-color-map-normal screen))
                 color))
-        (t (first (multiple-value-list (alloc-color screen color))))))
+        (t (nth-value 0 (alloc-color screen color)))))
 
 (defun find-font (cc specified-font &aux (font (or specified-font 0)))
   (if (integerp font)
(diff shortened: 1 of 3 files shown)


color.lisp
```diff
@@ -195,7 +195,7 @@ If COLOR isn't a colorcode a list containing COLOR is returned."
                     (screen-color-map-bright screen)
                     (screen-color-map-normal screen))
                 color))
-        (t (first (multiple-value-list (alloc-color screen color))))))
+        (t (nth-value 0 (alloc-color screen color)))))
```
PuercoPop, line 198:
```suggestion
        (t (alloc-color screen color))))
```
Wouldn't this work just the same? Or we want to avoid returning multiple values from the function?

PuercoPop (approved):
Left one question before. LGTM

color.lisp
PuercoPop: > I assumed that wanting to avoid returning multiple values was intentional. 

It was probably intentional, but that intention is likely to be misguided. Unless calling m-v-b or m-v-l the caller won't see the additional values. But you made the right call by preserving the semantics when rewriting code that doesn't have tests.

Issue #872, opened 2021-04-13
focus-color used on unfocused window


svenssonaxel:
To reproduce:
```
gnew A
gnew B
gselect A
exec x-terminal-emulator
exec x-terminal-emulator
gmove B
gmove B
gselect B
vsplit
```
Expected: Two windows side by side, one with a focus-color border, the other with an unfocus-color border.
Actual: Two windows side by side, both with focus-color borders.

svenssonaxel:
Link PR #873

PR #873, opened 2021-04-14
Do not put a border around the only window

### LINKED ISSUE #866: Unfocus Color used when only one window
The documentation for `set-unfocus-color` states that it's only used when there is more than one frame. However, it is used when there is only one frame. Either,
1) the documentation should be updated (suggestion below), or
2) preferrably, the behaviour should change to not use a border when there is only one window.

```diff
diff --git a/screen.lisp b/screen.lisp
index cae928a..1a90c04 100644
--- a/screen.lisp
+++ b/screen.lisp
@@ -280,8 +280,8 @@ there is more than one frame."
   (set-any-color screen-focus-color color))

 (defun set-unfocus-color (color)
-  "Set the border color for windows without focus. This is only used when
-there is more than one frame."
+  "Set the border color for windows without focus, and when there is only
+one frame."
   (set-any-color screen-unfocus-color color))

 (defun set-float-focus-color (color)
```

svenssonaxel:
Fixes #866

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -16,8 +16,7 @@ like xterm and emacs.")
   ;; give it a colored border but only if there are more than 1 frames.
   (let* ((group (window-group window))
          (screen (group-screen group)))
-    (let ((c (if (and (> (length (group-frames group)) 1)
-                      (eq (group-current-window group) window))
+    (let ((c (if (eq (group-current-window group) window)
                  (screen-focus-color screen)
                  (screen-unfocus-color screen))))
       (setf (xlib:window-border (window-parent window)) c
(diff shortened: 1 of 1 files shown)


svenssonaxel:
Just discovered that this will allow windows to become slightly overlapped. It looks like the window border is still not updated after `gmove`, maybe due to the same bug as in issue #872. Ideally, that issue should be fixed before merging.

svenssonaxel:
OK, I think both #866 and #872 should be fixed now. Not sure that it's fixed in the "right way" though, please review.

svenssonaxel:
Link issue #872

PR #878, opened 2021-04-18
Typo fix: match-properties-or-function -> match-properties-and-function

kevwargo:
The `match-properties-or-function` value was incorrectly used.


PuercoPop:
Thanks for submitting the fix! :tada:

PR #895, opened 2021-06-08
GitHub actions CI

tpine:
Build and run tests inside a dockerfile with every push and pull request.

This will help to ensure that the build doesn't break with changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+FROM ubuntu:20.04
+
+ENV HOME /root/
+
+RUN apt-get update \
+    && apt-get install -y sbcl curl build-essential autoconf git \
+    && curl -O https://beta.quicklisp.org/quicklisp.lisp \
+    && sbcl --load quicklisp.lisp --eval "(quicklisp-quickstart:install)" \
+    && sbcl --load "/root/quicklisp/setup.lisp"  --eval "(progn (setf ql-util::*do-not-prompt* t)(ql:add-to-init-file))" \
+    && sbcl --eval "(progn (ql:quickload \"clx\")(ql:quickload \"cl-ppcre\")(ql:quickload \"alexandria\")(ql:quickload \"fiasco\"))"
+
+COPY . .
+
+RUN ./autogen.sh \
+    && ./configure \
+    && make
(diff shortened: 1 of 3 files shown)


sdilts:
LGTM, except why use a separate docker container when the job should already be running in one? The build environment should already be clean, so there shouldn't be a need to use an additional container.

tpine:
Because you can then also reuse the same environment as the CI locally to run test and builds with if you wish to

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:20.04
+
+ENV HOME /root/
+
+RUN apt-get update \
+    && apt-get install -y sbcl curl build-essential autoconf git \
+    && curl -O https://beta.quicklisp.org/quicklisp.lisp \
+    && sbcl --load quicklisp.lisp --eval "(quicklisp-quickstart:install)" \
+    && sbcl --load "/root/quicklisp/setup.lisp"  --eval "(progn (setf ql-util::*do-not-prompt* t)(ql:add-to-init-file))" \
+    && sbcl --eval "(progn (ql:quickload \"clx\")(ql:quickload \"cl-ppcre\")(ql:quickload \"alexandria\")(ql:quickload \"fiasco\"))"
```
PuercoPop, line 10: `quickload` takes a list and we can take advantage that ASDF accepts string designators  as well to avoid escaping

```suggestion
    && sbcl --eval "(progn (ql:quickload '(clx cl-ppcre alexandria fiasco)))"
```

PuercoPop (approved):
Do we need to do something to enable the workflow as a status check on PRs?

tpine:
It should be doing that but it appears that github uses the branch you are merging into for its CI Actions

https://github.com/stumpwm/stumpwm/pull/895/files#diff-faff1af3d8ff408964a57b2e475f69a6b7c7b71c9978cccc8f471798caac2c88R3

tpine:
Here is a example of it running on push in my own repo.
https://github.com/tpine/stumpwm/runs/2780168978

tpine:
Ive tested by opening a pull request into this branch on my repo and it appears to work as long as the branch you are merging into has the CI Actions.
https://github.com/tpine/stumpwm/pull/1

PuercoPop:
> as long as the branch you are merging into has the CI Actions.

@tpine Yeah, I think they did that to avoid abusing CI for crypto or to avoid exploits that expose repository secrets

PR #896, opened 2021-06-21
Removed call to update-configuration from maximize-window.

### LINKED ISSUE #859: Reaper slowing down GUI
This is more an observation I want to record, rather than a precise bug report.

Reaper ( the commercial DAW, https://reaper.fm ) is slowing down progressively as it runs, causing the GUI of other running programs to become more sluggish as well, to a point where it is near impossible to open any menu, enter key or just switch windows. 

This seems to be some strange interaction with stumpwm, it does not happen with Gnome Shell (Mutter).  I have not tried other window managers. In the Reaper forums there are a few reports of GUI slowdowns, but without mentioning the window manager used. The slowdown seems to progress quicker, the larger the reaper window is. It is also most noticeable at first in the time it takes to open menus.

As soon as I terminate Reaper, the speed returns to normal.

Reaper uses libSwell, which seems to be their own binding over X11.

ldd reaper-6.24/libSwell.so 
        linux-vdso.so.1 (0x00007ffdeaf50000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fe6478cd000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007fe6478c8000)
        libX11.so.6 => /usr/lib64/libX11.so.6 (0x00007fe647784000)
        libXi.so.6 => /usr/lib64/libXi.so.6 (0x00007fe647772000)
        libfontconfig.so.1 => /usr/lib64/libfontconfig.so.1 (0x00007fe64772a000)
        libstdc++.so.6 => /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/libstdc++.so.6 (0x00007fe6474ac000)
        libm.so.6 => /lib64/libm.so.6 (0x00007fe647370000)
        libgcc_s.so.1 => /usr/lib/gcc/x86_64-pc-linux-gnu/9.3.0/libgcc_s.so.1 (0x00007fe647356000)
        libc.so.6 => /lib64/libc.so.6 (0x00007fe647196000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fe647bd4000)
        libxcb.so.1 => /usr/lib64/libxcb.so.1 (0x00007fe64716a000)
        libXext.so.6 => /usr/lib64/libXext.so.6 (0x00007fe647155000)
        libfreetype.so.6 => /usr/lib64/libfreetype.so.6 (0x00007fe64708f000)
        libexpat.so.1 => /usr/lib64/libexpat.so.1 (0x00007fe64705e000)
        libuuid.so.1 => /lib64/libuuid.so.1 (0x00007fe647054000)
        libXau.so.6 => /usr/lib64/libXau.so.6 (0x00007fe64704f000)
        libXdmcp.so.6 => /usr/lib64/libXdmcp.so.6 (0x00007fe647047000)
        libbz2.so.1 => /lib64/libbz2.so.1 (0x00007fe647032000)
        libpng16.so.16 => /usr/lib64/libpng16.so.16 (0x00007fe646ff8000)
        libz.so.1 => /lib64/libz.so.1 (0x00007fe646fdc000)
        libbsd.so.0 => /usr/lib64/libbsd.so.0 (0x00007fe646fc2000)

anwyn:
Calling update-configuration on maximize triggers an event loop for
all Reaper windows [https://reaper.fm].

Fixes #859

I have stumpwm running with this patch for a week now and did not notice any strange window behavior.

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -215,8 +215,7 @@ than the root window's width and height."
                                   wy
                                   (- (xlib:drawable-height (window-parent win)) height wy))
                             :cardinal 32))
-    (update-decoration win)
-    (update-configuration win)))
+    (update-decoration win)))
 
 ;;;
 


PuercoPop:
I'm not sure why this fixes the issue you are running into. The `update-configuration` call was added to fix a bug. Notifying the clients about the new dimensions of the window seems like the right thing to do. 

https://github.com/stumpwm/stumpwm/pull/132
https://github.com/stumpwm/stumpwm/commit/5d26584080720301765895c2b0ce1e9be65b242a

Skimming the bug report (#859) it seems the call results in an 'infinite ping pong' of property notify messages.

anwyn:
Now that you mention it, I tried and can reproduce the menu/mouse position problem with a Java Swing application. You are right, it is an infinite property notify loop with reaper. Perhaps it can be broken in a better place, perhaps not sending a notification if nothing changed? Unfortunately I'm lacking the insight into stumpwm's code base to pinpoint the right spot to do this.

PR #897, opened 2021-06-23
be more stringent about string inputs to "y or n" prompts

wdouglass:
I kept trying to call commands with arguments like `t`, which `:y-or-n` evaluates to `nil`, which confused me. So i made `:y-or-n` throw an error if the argument isn't `"y"` or `"n"`

alternatively, we could create a list of arguments that evaluates to `t`, i would propose `t`, `"t"`, and `"Y"`

let me know what you think!

### DIFF
--- a/command.lisp
+++ b/command.lisp
@@ -340,7 +340,11 @@ then describes the symbol."
 (define-stumpwm-type :y-or-n (input prompt)
   (let ((s (or (argument-pop input)
                (read-one-line (current-screen) (concat prompt "(y/n): ")))))
-    (equal s "y")))
+    (if (equal s "y")
+        t
+        (if (equal s "n")
+            nil
+            (throw 'error (format nil "Invalid (y/n) input: ~A" s))))))
 
 (defun lookup-symbol (string)
   ;; FIXME: should we really use string-upcase?
--- a/command.lisp
+++ b/command.lisp
@@ -338,18 +338,10 @@ then describes the symbol."
       ,@body)))
 
 (define-stumpwm-type :y-or-n (input prompt)
-  (let* ((valid '(("y" . t)
-                  ("Y" . t)
-                  (t . t)
-                  ("n" . nil)
-                  ("N" . nil)
-                  (nil . nil)))
+  (let* ((valid '("y" "Y" t))
          (s (or (argument-pop input)
                 (read-one-line (current-screen) (concat prompt "(y/n): "))))
-         (out (assoc s valid :test 'equal)))
-    (if out
-        (cdr out)
-        (throw 'error (format nil "Invalid (y/n) input: ~A" s)))))
+         (member s valid :test #'equal))))
 
 (defun lookup-symbol (string)
   ;; FIXME: should we really use string-upcase?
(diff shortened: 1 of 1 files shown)


PuercoPop:
> alternatively, we could create a list of arguments that evaluates to t, i would propose t, "t", and "Y"

IMHO that is the way to go. Accept 'y'/'Y' and t.

wdouglass:
Cool, i'll revise the commit

sdilts:
Is it going to be a problem that this function can now throw an error? It adds the potential for a non-local exit everywhere that it is called, although I would hope that anything important enough to be affected would already be wrapped in `unwind-protect` or `handler-case`.

PuercoPop:
> Is it going to be a problem that this function can now throw an error? 

You are right, thanks for bringing it up. Yeah we shouldn't throw on 'unexpected' input

command.lisp
```diff
@@ -338,9 +338,18 @@ then describes the symbol."
...
-    (equal s "y")))
+  (let* ((valid '(("y" . t)
+                  ("Y" . t)
+                  (t . t)
+                  ("n" . nil)
+                  ("N" . nil)
+                  (nil . nil)))
+         (s (or (argument-pop input)
+                (read-one-line (current-screen) (concat prompt "(y/n): "))))
+         (out (assoc s valid :test 'equal)))
```
PuercoPop, line 349: There is no need for an assoc list. Just

```(let ((s (or (argument-pop input)
               (read-one-line (current-screen) (concat prompt "(y/n): ")))))
    (member s '("y" t) :test #'equal)))
```

In Lisp the expection for predicates is nil => false, anything else => true.

wdouglass: > Is it going to be a problem that this function can now throw an error? It adds the potential for a non-local exit everywhere that it is called, although I would hope that anything important enough to be affected would already be wrapped in `unwind-protect` or `handler-case`.

In my testing, it's manifested itself as a message displayed when i try certain commands (with invalid arguments), although that's probably coming from a `handler-case` somewhere

maybe expanding the "affirmative" values is enough and we don't need to be strict about the negative ones -- and thanks for the advice about `member`, i'm still learning

command.lisp
```diff
@@ -338,9 +338,10 @@ then describes the symbol."
       ,@body)))
 
 (define-stumpwm-type :y-or-n (input prompt)
-  (let ((s (or (argument-pop input)
-               (read-one-line (current-screen) (concat prompt "(y/n): ")))))
-    (equal s "y")))
+  (let* ((valid '("y" "Y" t))
+         (s (or (argument-pop input)
+                (read-one-line (current-screen) (concat prompt "(y/n): "))))
+         (member s valid :test #'equal))))
```
PuercoPop, line 344: - valid is not correct, "n'"  is not invalid input, it is a negative response. 
-  we can use equalp to do case insensitive comparisons.

```suggestion
  (let* ((positive-responses '("y" t))
         (s (or (argument-pop input)
                (read-one-line (current-screen) (concat prompt "(y/n): "))))
         (member s positive-responses :test #'equalp))))
```

PuercoPop (approved):
Sorry for not being more concrete with my suggestion. In the interest of avoiding the back and forth I'm auto-accepting my suggestion and merging. Thanks for the PR :tada:

PR #899, opened 2021-06-26
Fix compile error: fix let binding body position

vhallac:
* command.lisp (:y-or-n):
The body was placed in let varlist by mistake. I moved it out.


PuercoPop:
Thats what I get for editing code in the GitHub UI. Thanks for catching it!

PR #900, opened 2021-06-26
Some floating group fixes for gmove

vhallac:
This pull request addresses two issues when moving a window to a float group:

1. The group was unconditionally selected after a gmove - as if gmove-and-select was used
2. Full screen windows got back to normal size after moving.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -138,14 +138,15 @@
 
 (defmethod group-startup ((group float-group)))
 
-(flet ((add-float-window (group window)
+(flet ((add-float-window (group window raise)
          (change-class window 'float-window)
          (float-window-align window)
-         (group-focus-window group window)))
-  (defmethod group-add-window ((group float-group) window &key &allow-other-keys)
-    (add-float-window group window))
-  (defmethod group-add-window (group (window float-window) &key &allow-other-keys)
-    (add-float-window group window)))
+         (when raise
+           (group-focus-window group window))))
+  (defmethod group-add-window ((group float-group) window &key raise &allow-other-keys)
+    (add-float-window group window raise))
+  (defmethod group-add-window (group (window float-window) &key raise &allow-other-keys)
+    (add-float-window group window raise)))
 
 (defun %float-focus-next (group)
   (if (group-windows group)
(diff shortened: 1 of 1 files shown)


PuercoPop:
> 2. Full screen windows got back to normal size after moving.

This behaviour makes sense to me, or at least seems like the reasonable default. Not sure if we want to change it. Although I don't use fullscreen much tbh

vhallac:
> This behaviour makes sense to me, or at least seems like the reasonable default. Not sure if we want to change it. Although I don't use fullscreen much tbh

I used to use tiling mode. Now, I am in a little experimental mode after realizing most of my desktops had one frame only. I am using float groups, where my common windows are switched to full screen on create; which turned out to be surprisingly suitable for my workflow. 

Anyhow, It is OK to keep old behavior for fullscreens. But the first commit in the merge request is a bug fix, I think. Should I close this and open a new pull request with just the one commit?

PuercoPop:
> Anyhow, It is OK to keep old behavior for fullscreens. But the first commit in the merge request is a bug fix, I think. Should I close this and open a new pull request with just the one commit?

Yes, in general each pull request should address a single issue. Let's open a separate pull request to address the issue in the first commit. I would keep this one open with only the second commit in case someone else want's to chime in and support the new behaviour though.

<hr />

The first commit looks like the right solution at first glance however If we check who calls group-add-window (M-? if you are an emacs user) we it is called by the following functions:

- PROCESS-MAPPED-WINDOW
- COPY-WINDOW-TO-GROUP
- MOVE-WINDOW-TO-GROUP
- RESTORE-WINDOW

However only process-mapped-window calls the function why any keyword arguments (`placement-data`) and it checks for the raise keyword argument afterwards. But let's discuss that in the separate PR.

vhallac:
> Yes, in general each pull request should address a single issue. Let's open a separate pull request to address the issue in the first commit. 
Done: https://github.com/stumpwm/stumpwm/pull/904

> I would keep this one open with only the second commit in case someone else want's to chime in and support the new behaviour though.
Let's see if anyone has an opinion. It can be considered correct either way, and I doubt there are a lot of people who use floating-groups. :)


> But let's discuss that in the separate PR.
OK

vhallac:
The PR can be tracked in PR #904. I am closing this one.

PR #904, opened 2021-06-28
Call group-focus-window only when requested

vhallac:
For float groups, when moving window to another group, the target group was selected automatically.

This patch fixes this by adding the raise parameter - similar to tiling and dynamic groups.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -138,14 +138,15 @@
 
 (defmethod group-startup ((group float-group)))
 
-(flet ((add-float-window (group window)
+(flet ((add-float-window (group window raise)
          (change-class window 'float-window)
          (float-window-align window)
-         (group-focus-window group window)))
-  (defmethod group-add-window ((group float-group) window &key &allow-other-keys)
-    (add-float-window group window))
-  (defmethod group-add-window (group (window float-window) &key &allow-other-keys)
-    (add-float-window group window)))
+         (when raise
+           (group-focus-window group window))))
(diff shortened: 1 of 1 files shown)


vhallac:
Continuing discussion on https://github.com/stumpwm/stumpwm/pull/900... 

> The first commit looks like the right solution at first glance however If we check who calls group-add-window (M-? if you are an emacs user) we it is called by the following functions:
> 
>     * PROCESS-MAPPED-WINDOW
> 
>     * COPY-WINDOW-TO-GROUP
> 
>     * MOVE-WINDOW-TO-GROUP
> 
>     * RESTORE-WINDOW
> 
> 
> However only process-mapped-window calls the function why any keyword arguments (`placement-data`) and it checks for the raise keyword argument afterwards. But let's discuss that in the separate PR.

I've looked at the same locations. All three groups now accept the `raise` parameter, and currently tiling and (after this fix) floating groups would switch group when it is set.

Since tiling groups have additional logic on frame and raise, I would suggest adding raise support to dynamic group, and removing calls to `(switch-to-group (window-group window))`. We should also update `gmove-and-follow` to make use of the raise parameter. The caveat here is that I don't understand how the `group-focus-window` selects window's group.

A simpler alternative is to remove raise parameters from floating and dynamic groups, and not worry about tiling group (maybe experiment with removing the call to `group-focus-window` in there. I am not too fond of this, mainly because it leaves switching to target group on the caller side, when it can be handled at a more central location.

PR #906, opened 2021-07-15
Add a type declamation for *TOP-LEVEL-ERROR-ACTION*

CatherineTower:
The compiler will now ensure that *TOP-LEVEL-ERROR-ACTION* will only be one of the prescribed values. This is useful because toplevel errors are handled in an ECASE expression, so if *TOP-LEVEL-ERROR-ACTION* contains an unexpected value at the time of a toplevel error, the expression will cause another, more cryptic error to be signaled.


PuercoPop:
Thanks for the PR!

PR #910, opened 2021-07-30
Rewrite INPUT-INSERT-CHAR

CatherineTower:
This fixes a bug that came up on one of my machines where any time
this function was called, an error would be signalled saying that
REPLACE couldn't use -1 as an index. This implementation is
considerably simpler, and doesn't seem to cause any problems in
usage.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -606,10 +606,7 @@ second and neither excedes the bounds of the input string."
   "Insert @var{char} into the input at the current
 position. @var{input} must be of type @var{input-line}. Input
 functions are passed this structure as their first argument."
-  (vector-push-extend #\_ (input-line-string input))
-  (replace (input-line-string input) (input-line-string input)
-           :start2 (input-line-position input) :start1 (1+ (input-line-position input)))
-  (setf (char (input-line-string input) (input-line-position input)) char)
+  (vector-push-extend char (input-line-string input))
   (incf (input-line-position input)))
 
 (defun input-substring (input start end)


CatherineTower:
I checked out all of the edits to `input.lisp` that touched `input-insert-char`, and I couldn't find any specific reason it was written the way it was; the implementation was that way from the get-go.

landakram:
I just ran into this same issue. FWIW, the error seems to signal in the latest version of sbcl (2.7.1). I worked around it by downgrading to 2.1.1, the previous version I was using in arch.

CatherineTower:
Oh weird. I spent some time looking at the internal implementation of `replace` until my brain started oozing out of my ears and tried messing with Stump instead.

I've got StumpWM running on two different machines, one Arch Linux and one FreeBSD, both using very up-to-date SBCLs built from the repos. The BSD machine hasn't had the issue at all, it's only shown up under Arch.

Interestingly, the Arch machine was running Fedora until last week when I decided to switch, but I don't think I upgraded SBCL when I did my distro-hop, and the problem only manifested under Arch. So basically I have no idea where the problem actually resides, but this patch fixes it. And -- I think -- cleans up the code a bit.

CatherineTower:
I can confirm that the error doesn't happen under a freshly-built SBCL 2.1.1.

CatherineTower:
I've tracked it down, it's commit `1db840a65` in SBCL that introduced the bug. I'll open an issue on their issue tracker as well. I still think this pull request holds up, since I'm not really sure why the `replace` code was in there in the first place. It's just a cleanup rather than a bug fix.

Unless of course it's in there for a reason that I couldn't find, in which case I'd definitely be interested to hear about why.

jorams:
Thank you for figuring out the error!
The reason the code uses `replace` and `setf` is that you can move the cursor. With this PR new characters are always inserted at the end.

CatherineTower:
Ah yes, I can see how that would be an issue. Thanks for letting me know!
SBCL has fixed their big on the development branch, so it's sounding like there'll never be an issue for the majority of users.

Issue #912, opened 2021-08-02
The value -1 is not of type SB-INT:INDEX when binding SB-VM::SRC-START


emacsomancer:
Running stumpwm HEAD with sbcl-2.1.7, I'm now getting the following errors with [prefix]-`semicolon` and [prefix]-`colon`, as soon as I try to enter any other keystrokes:
```
Error In Command 'colon': The value -1 is not of type SB-INT:INDEX when binding SB-VM::SRC-START

Error In Command 'eval': The value -1 is not of type SB-INT:INDEX when binding SB-VM::SRC-START

```

emacsomancer:
Ah, I suppose this is the same issue as in https://github.com/stumpwm/stumpwm/pull/910 .

PR #913, opened 2021-08-02
 Fix problem input-insert-char on SBCL 2.1.7

epanji:
In **input-insert-char** function, we use replace with same sequence in arguments. It signal an error "The value -1 is not of type SB-INT:INDEX when binding SB-VM::SRC-START". It is appear when SBCL updated to version 2.1.7 on my machine. Since replace destructively modifies SEQUENCE1, we should not used same sequence for SEQUENCE2.

It is probably the same issues:
https://github.com/stumpwm/stumpwm/issues/912
https://github.com/stumpwm/stumpwm/pull/910

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -607,8 +607,9 @@ second and neither excedes the bounds of the input string."
 position. @var{input} must be of type @var{input-line}. Input
 functions are passed this structure as their first argument."
   (vector-push-extend #\_ (input-line-string input))
-  (replace (input-line-string input) (input-line-string input)
-           :start2 (input-line-position input) :start1 (1+ (input-line-position input)))
+  (replace (input-line-string input) (copy-seq (input-line-string input))
+           :start1 (1+ (input-line-position input))
+           :start2 (input-line-position input))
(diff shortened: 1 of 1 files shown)


jorams:
The error is a bug in SBCL, fixed in https://github.com/sbcl/sbcl/commit/2bb2c6dbfac0926a62cfbb8bb412ceaf84450555. It's fine for both sequences passed to `replace` to be the same object.

The [spec][1] specifies how the sequence is modified and even deals with the source and destination being overlapping regions of the same object. It would only be problematic if they were *not* the same object, but the actual data is the same:

> However, if sequence-1 and sequence-2 are not the same, but the region being modified overlaps the region being copied from (perhaps because of shared list structure or displaced arrays), then after the replace operation the subsequence of sequence-1 being modified will have unpredictable contents.

[1]: http://www.lispworks.com/documentation/HyperSpec/Body/f_replac.htm

epanji:
Thank you for clarifying this matter.

Feel free if you need to close this PR.

PR #927, opened 2021-10-02
Fix window head selection on full screen

vhallac:
When a window is maximized it can move to the wrong head, depending on whether left/top is matched or right/bottom is matched first in the list of heads. This patch makes it choose the same head independent of the head list order.


dmb2:
In the future, no white space changes please :)

PR #933, opened 2021-11-30
Fix improper item-object in menu filter predicate

xFA25E:
I found a bug in `select-from-menu` function. It tried to take a second element of menu element when my menu elements were conses.

I think that we should not force the menu elements to be lists of length 2. Instead, they should be conses where car element is a string displayed in menu and cdr is item-object.

### DIFF
--- a/menu-definitions.lisp
+++ b/menu-definitions.lisp
@@ -230,7 +230,7 @@ a re-computation of the match."
           (labels ((match-p (table-item)
                      (funcall (single-menu-filter-pred menu)
                               (car table-item)
-                              (second table-item)
+                              (cdr table-item)
                               (single-menu-current-input menu))))
             (setf (menu-table menu) (remove-if-not #'match-p (single-menu-unfiltered-table menu))
                   (menu-selected menu) 0)


dmb2:
I'm concerned with merging this.  cdr returns the rest of a list, but second returns the second element.  I don't see any changes to the rest of how data is handled in these functions to know for certain that this won't break with your change.  Could you give some example code of what you were trying when you discovered the bug?

xFA25E:
I was trying to put cons pairs (alist) in the `select-from-menu` function. I am using stumpwm with this change for quite some time now.

If this is an intended feature, than I will close this PR. But, in the manual section about `select-from-menu`, it does not say that ITEM-OBJECT  is a second element of the sublist, so I assumed that it accepted a cons.

sdilts:
Can we see the code you are using to build the menu data structure?

I agree that a cons cell would make more sense here, but it would probably be a (minor) breaking change. I'll look at it closer when I have a computer I can actually test it on.

xFA25E:
```common-lisp

;; parse contents of a file into an alist

(defun parse-entry-contents (contents)
  (let ((fields nil))
    (re:do-register-groups (pass) ("^(.*)\\n" contents)
      (push (cons "password" pass) fields))
    (re:do-register-groups (key value) ("\\n([a-zA-Z_]+): (.*)" contents)
      (push (cons key value) fields))
    (sort (delete-if #'null fields) #'string< :key #'car)))

(let ((menu (acons "autotype" :autotype (parse-entry-contents contents))))
  (swm:select-from-menu (swm:current-screen) menu "Pass type field: "))

;; use it to build a menu for interactive commands

(swm:defcommand pass-menu () ()
  (let ((menu '(("type" . "pass-type")
                ("edit" . "pass-edit")
                ("insert" . "pass-insert")
                ("otp" . "pass-otp"))))
    (match (swm:select-from-menu (swm:current-screen) menu "Pass menu: ")
      ((cons _ command)
       (swm::eval-command command t)))))

```

xFA25E:
Also, I looked at every call of `select-from-menu` in github. Nobody is using `:filter-pred` argument. It is very likely that this change is not gonna break any code. However, this is, indeed, a (minor) breaking change and I understand this. 

I can totally rewrite my config to conform to existing API. Originally, I thought that it was a bug, because, as I said earlier, nobody was using `:filter-pred` on github. I assumed that, simply, nobody noticed it.

xFA25E:
I also use code like this. Putting _cdr_ into a _second_ element of a list woud be ugly
```common-lisp
;;; MAIN MENU

(defvar *main-menu*
  (list (list "screenshot"     :command "screenshot")
        (list "suspend"        :shell (list "systemctl" "suspend"))
        (list "mount"          :shell (list "rmount"))
        (list "unmount"        :shell (list "rumount"))
        (list "clipboard-type" :command "clipboard-type")))

(swm:defcommand main-menu () ()
  (match (swm:select-from-menu (swm:current-screen) *main-menu* "Main menu: ")
    ((list _ :command command) (swm::eval-command command t))
    ((list _ :shell shell) (launch-program shell))))

```

dmb2:
I don't see a compelling reason accept this breaking change.

PR #967, opened 2022-03-04
Downcase ungrammatical capitals in prompts

uti5:


szos:
While I cant speak for others, personally I think having each word capitalized (as in a title or similar) is best for prompts.

uti5:
I didn't expect any argument here, and I wasn't expressing my personal taste. There is no tradition for capitalizing prompts and there's no reason to do it. In modern non-poetic English, capitals are only used in sentence initials, proper names, and titles, and the propriety of using of them in titles is disputed.

It may also be observed that the prompt in #'describe-function in Emacs, on which #'describe-function in Stump is presumably modelled, is "Describe function: ".

szos:
My apologies, I intend for my comment to come across as argumentative; I was just expressing what style of prompt I like best.

uti5:
Hi, you're welcome!  (Thanks for maintaining Stump!)  As I said above,
this has nothing to do with personal taste, I'm simply pointing out
what is correct English.  I'll leave it to you to decide whether
correct language is "an advantage" or "increases usability".  The
reason one _cannot_---grammatically---use "title case'' for prompts
is that prompts are not titles.  (What is a title?---A name.  That's
why it has capitals.---What is a prompt?---A question.---Is a question
a name?---No.)

PR #968, opened 2022-03-27
Additional keysym names

Phundrak:
Add additional keysym names. This gives better readability to several
characters that can be accessed directly from the French bépo layout,
as well as the quoteleft and quoteright characters. That way, the
character itself is displayed in which-key-mode and other similar
features instead of the character’s internal name which the user
probably don’t care about that much.


bovid-19:
I think these should be called [chevrons](https://en.wikipedia.org/wiki/Bracket#Angle_brackets) and [guillemets](https://en.wikipedia.org/wiki/Guillemet):

> (define-keysym-name "<" "quoteleft")
> (define-keysym-name ">" "quoteright")

> (define-keysym-name "«" "guillemotleft")
> (define-keysym-name "»" "guillemotright")

And thank you for the learning opportunity: I wasn't aware chevrons had the same name [in french](https://fr.wikipedia.org/wiki/Chevron_(typographie)).

Phundrak:
> I realized the names you used were the current ones in stumpwm.
>
> Since quote(left|right) are deprecated and guillemot actually
> denotes a bird, adding the names I suggested might make sense,
> though.

You’re right, I didn’t notice `quoteleft' and `quoteright' are deprecated,
I’ll push a new commit to replace their name with `grave' and `apostrophe'
respectively.

However, as you noted in your edit, `guillemot' is the internal name of
the `«' and `»' characters as defined in Xorg, so this cannot change. If
you have the Xorg headers on your machine, you can check it in the
`/usr/include/X11/keysymdef.h' file on your system, or on the Xorg
repository here:
<https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/master/include/X11/keysymdef.h>

> And thank you for the learning opportunity: I wasn't aware chevrons
> had the same name in french.

The English word “chevron” actually comes from French ;)

--
Lucien “Phundrak” Cartier-Tilet
<https://phundrak.com> (Français)
<https://phundrak.com/en> (English)
Sent from GNU/Emacs

PR #969, opened 2022-03-28
Export symbol only-one-frame-p

tobyWorland:


dmb2:
Please explain why this is needed.  If you are developing a module and require the symbol explain the context.

tobyWorland:
Apologies.
I wanted to add an option to the gaps contrib that would only add gaps if there is more than one frame. I planned to use this symbol to do so.

dmb2:
Please add it to the manual and I'll merge this.

tobyWorland:
It seems someone had already added it to the manual prior to me exporting it.
https://github.com/tobyWorland/stumpwm/blob/db34981389c412a493ccc9fbdce0e81369d57c7b/stumpwm.texi.in#L1878

PR #970, opened 2022-04-24
allow `unmaximize` command to accept a window as its argument

defaultxr:
`unmaximize` is a command that is useful from Lisp, not just interactively. Because it assumes the current window, it is not possible to use it from functions in `*new-window-hook*` since they are called prior to the new window being made the current one. This fixes that by allowing you to specify a window when calling it from Lisp. Interactively, it still defaults to the current one.

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -456,11 +456,10 @@ frame. Possible values are:
                                                 (window-height-inc window)))))
       (maximize-window window)))))
 
-(defcommand (unmaximize tile-group) () ()
+(defcommand (unmaximize tile-group) (&optional (window (current-window))) (:rest)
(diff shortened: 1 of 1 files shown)


tile-window.lisp
```diff
@@ -456,11 +456,10 @@ frame. Possible values are:
                                                 (window-height-inc window)))))
       (maximize-window window)))))
 
-(defcommand (unmaximize tile-group) () ()
+(defcommand (unmaximize tile-group) (&optional (window (current-window))) (:rest)
```
PuercoPop, line 459: Does this change the interactive call behaviour? Will it now ask for a window?

defaultxr: No. Interactively it still just uses the current window; it doesn't ask for any input.

Issue #971, opened 2022-04-26
[Feature Request] Fullscreen contained in a frame


joshcho:
In other window managers (e.g. xmonad), going fullscreen on a video in browser can be configured to be contained in the frame.

![full-screen-2](https://user-images.githubusercontent.com/6498524/165205880-c754fd39-4664-48a4-9964-056adba13f77.png)
This is xmonad, with Brave on the left frame with youtube video as fullscreen.

If I wanted to implement something like this, where should I start looking?

Reference links:
https://www.reddit.com/r/stumpwm/comments/u6tfpr/can_fullscreen_eg_youtube_be_contained_in_the/
https://www.reddit.com/r/stumpwm/comments/awnfwf/is_it_possible_to_contain_full_screen_to_a_frame/

szos:
This feature was merged in #974, can this issue be closed?

PR #974, opened 2022-05-19
Add in frame fullscreen functionality

szos:
This PR allows for windows to be made fullscreen within their frame instead of their head. By default all windows will fill the head when being made fullscreen, it is only by defining a rule that they are made fullscreen within the window. The rule add/remove functionality was placed in `primitives.lisp` along with the rule defining macro, as this is where `define-frame-preference` is defined, but if there is a better location they can be moved there. This would address issue #971.

This PR doesnt change the default behavior, and windows fullscreened within their frame can coexist with normal fullscreen windows (tested with firefox and vlc). The usage of keys in an alist allow for temporarily overriding a rule, or adding rules that run only once; users could define a macro such as: 

```lisp
(defmacro fullscreen-in-frame-only-once ((win &key keep-until-matched) &body body)
  (with-gensyms (g l)
    `(define-fullscreen-in-frame-rule ,g (,win)
       (let ((,l (progn ,@body)))
         ,(if keep-until-matched
              `(when ,l (remove-fullscreen-in-frame-rule ',g))
              `(remove-fullscreen-in-frame-rule ',g))
         ,l))))
```

Here is a screenshot of multiple fullscreen-in-frame firefox instances: 

2022-05-19-145336_1366x768_scrot

And here is the rule used to fullscreen these windows specifically (helper functions not included): 

```lisp
(define-fullscreen-in-frame-rule firefox-rule (win)
  (when (classed-p win "Firefox")
    (youtube-url-p (get-firefox-url win))))
```

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -1179,6 +1179,44 @@ Must be an unquoted symbol to be looked up at runtime.
   "Clear all window placement rules."
   (setf *window-placement-rules* nil))
 
+(defvar *fullscreen-in-frame-p-window-functions* nil
+  "A alist of predicate functions for determining if a window should be
+fullscreen in frame.")
+
+(defun fullscreen-in-frame-p (win)
+  (some (lambda (r)
+          (let ((res (funcall (cdr r) win)))
+            (when res
+              (dformat 3 "Fullscreen in frame selector ~A matches window ~A"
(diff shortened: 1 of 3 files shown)


joshcho:
Thank you! This was sorely needed.

PR #977, opened 2022-05-23
Implement clickable mode line

szos:
This PR implements a clickable mode line. Im opening this as a draft PR because, while this is functional, I'm unsure if this is the approach that should be taken to implement a clickable mode line. The goals with this implementation were to try and do things the way that other formatters are done, and maintain backwards compatibility. No mode line formatters were changed so no default mode line formatters will allow for clickable items. If this is an acceptable approach to implementing clicking, Ill go through and add (or modify, if its desired) the appropriate mode line formatters to allow clicking, eg click to focus window, click to focus group, etc. 

For a brief overview, checkout this branch and make the documentation (stumpwm.info), and look at the new section under the mode line chapter. 

Here is a video of the clicking behavior: 

https://user-images.githubusercontent.com/39150985/169815876-3ae4f9e8-2e42-47d0-9092-0e2852bb836b.mp4

### The way it works

This works by implementing two more color formatters, `:on-click` and `:on-click-end`, which track the x and y bounds of whatever is enclosed by them, as well as an ID and some arbitrary arguments associated with that specific instance of the formatter. These bounds are updated every time the modeline is redrawn, as part of the rendering process. They are stored in two slots that are added to the mode line structure. 

The actual click->funcall process is done using the `*mode-line-click-hook*`, and checks the mode line for the first bounds that the x y coordinates of the click falls within. If none is found, then nothing is done, but if one is found then its ID is looked up in a global alist of IDs->functions, and the function is called.


dmb2:
A few thoughts:
1. THIS IS AMAZING!
2. I'm happy to do it this way just because something working is better than nothing.
3. FWIW: I always envisioned a re-write of the modeline code that could stand-alone from stumpwm.  The custom clickable modeline would be a generic taskbar that would happen to be the default for stumpwm.
4. We have some outstanding bugs related to when and how the modeline refreshes itself.  If it refreshes two often it bogs down the system badly.  I'm a bit worried about performance implications here.

szos:
I'm glad you like it! It sounds like this is a good way of doing it for now, so I'll make the final changes and make this a true pull request. I think a new mode line would be best - if it could communicate with stumpwm through an ipc mechanism and run entirely in its own thread without touching shared global state, I think that would be best. I've tried several times to get such a thing working with CLIM, but I either dont understand CLIM well enough or it doesnt expose low enough level facilities to do what is needed (for example, last attempt stalled when I couldnt get a consistent lower window bound, so I eventually gave up). 

Regarding performance, I dont think this should impact performance. While the mode line gets rendered multiple times, this uses a temporary slot to store the bounds, and then updates the main clickable slot. Granted, I havent tested with more than 8 clickable areas active at a time, but I saw no performance drop (on a 2011 celeron). Ultimately this is just logging the bounds that are already calculated when rendering, so I think the most expensive things will be consing up n lists of 6 elements every render, where n is the number of clickable areas. That being said, Ill try generating many clickable areas (perhaps 20-30?) and seeing how it performs; perhaps you could do the same?

With regards to making final changes, would it be best to leave all existing formatters as they are, update some of them to use clickable areas, or implement new versions that do the same thing but with clickable areas?

montaropdf:
Very promizing.

Roland Everaert
---
Use the F.O.S.S., Luke

Envoyé avec la messagerie sécurisée [ProtonMail](https://protonmail.com/).

------- Original Message -------

TeMPOraL:
> I'm unsure if this is the approach that should be taken to implement a clickable mode line. (...)
> This works by implementing two more color formatters, `:on-click` and `:on-click-end` (...).

I sort of like this implementation. In a way, "colored" strings in StumpWM are already sort of like propertized strings in Emacs, and this is exactly how you do clickable text in Emacs - by a text property.

szos:
> I sort of like this implementation. In a way, "colored" strings in StumpWM are already sort of like propertized strings in Emacs, and this is exactly how you do clickable text in Emacs - by a text property.

One limitation of this approach is that clickable text areas cant be nested - is it the same in emacs?

TeMPOraL:
> One limitation of this approach is that clickable text areas cant be nested - is it the same in emacs?

I'm not sure what do you mean by nesting clickable text. Do you mean something like:
  (:clickable :target 'one "foo"
     (:clickable :target 'two "bar"
     "baz"))

Which would render as "foobarbaz", where clicking on "bar" runs action `'two`, and clicking anywhere else in the text runs action `'one`?

Either way, I'm not sure. I think the underlying data structure is flat, and given the Emacs text property API is imperative, I guess it depends on the order in which you set text properties.

In fact, I did the following test just now:

```elisp
(let ((put-2-then-1 (get-buffer-create "click-put-2-then-1"))
      (put-1-then-2 (get-buffer-create "click-put-1-then-2"))
      (add-2-then-1 (get-buffer-create "click-add-2-then-1"))
      (add-1-then-2 (get-buffer-create "click-add-1-then-2"))
      (map-1 (make-sparse-keymap))
      (map-2 (make-sparse-keymap)))

  (define-key map-1 [mouse-1] (lambda () (interactive) (message "Action in map 1")))
  (define-key map-2 [mouse-1] (lambda () (interactive) (message "Action in map 2")))

  (with-current-buffer put-2-then-1
    (insert "foobarbaz\n")
    (put-text-property 4 7 'keymap map-2)
    (put-text-property 1 9 'keymap map-1))

  (with-current-buffer put-1-then-2
    (insert "foobarbaz\n")
    (put-text-property 1 9 'keymap map-1)
    (put-text-property 4 7 'keymap map-2))

  (with-current-buffer add-2-then-1
    (insert "foobarbaz\n")
    (add-text-properties 4 7 `(keymap ,map-2))
    (add-text-properties 1 9 `(keymap ,map-1)))

  (with-current-buffer add-1-then-2
    (insert "foobarbaz\n")
    (add-text-properties 1 9 `(keymap ,map-1))
    (add-text-properties 4 7 `(keymap ,map-2))))
```

In case of both `put-text-property` and `add-text-properties`, the second call overrides the first call. Thus, the `2-then-1` cases have the entire string execute "action in map 1" on click, while in `1-then-2` cases, clicking on "bar" will execute "action in map 2".

szos:
> Thus, the 2-then-1 cases have the entire string execute "action in map 1" on click, while in 1-then-2 cases, clicking on "bar" will execute "action in map 2", while rest of the text will execute "action in map 1".

This is different than the StumpWM implementation, where a :on-click formatter after an on click formatter will override the previous one, and :on-click-end will terminate the previous :on-click formatter.

TeMPOraL:
Wouldn't nesting be achievable if you made `current-on-click` into a stack, with `:on-click` pushing on it, and `:on-click-end` popping from it? Height of the stack could be used as "z-order", to ensure the overlap happens in the correct order.

dmb2:
Good to go?

szos:
Id like to implement the suggestion from @TeMPOraL, ill bang it out this weekend and @ you when its ready to be merged

montaropdf:
By the way, does your implementation allows to click on other elements then the application items?

I think about the module that allows to display the date & time this could allow the developer of said module, or someone else, to show an interactive calendar, for example.

szos:
> By the way, does your implementation allows to click on other elements then the application items?

Yes it does; the way this works allows arbitrary functions to be called when a text area is clicked. You could register a function under the id `:ml-on-click-open-calendar`, and then wrap the date/time text using `(format-with-on-click-id text :ml-on-click-open-calendar)`. Then when that text is clicked, the function you registered would be run.

montaropdf:
Great.

I hope their will be an extensive documentation about this feature, I usually find the documentation a bit terse, at least the last time I read it.

szos:
If you have suggestions on the documentation id like to hear them, especially if the examples added are overly terse.

szos:
@dmb2 This should be ready for merging; I've implemented nested clickable areas as per what @TeMPOraL described.

szos:
When merging, it might be best to squash these commits, as some of them change very little.

dmb2:
I'm not a fan of squashing... I don't think our codebase is big enough that git is slowing down development.

Issue #936, opened 2021-12-01
Modifier AltGr ignored in function `read-one-line` when using bépo layout


Phundrak:
When entering a string in the input line, any keypress to AltGr (right alt) is ignored if the current layout is the bépo layout.

Step to reproduce:
- `setxkbmap fr bepo_afnor` or `setxkbmap fr bepo` (keep a `setxkbmap us` or equivalent somewhere just in case)
- Call the command `eval-line` interactively (`C-t :` by default)
- Press the AltGr / right alt key
- The input line will blink as it does in case of an incorrect input

I tested just in case if the issue is present with the qwerty layout and it seems it is not the case.

This breaks some of the StumpWM usability as the AltGr key is quite important with this layout (`&`, `|`, and `'` are accessed with it for instance). I did some digging and I think the issue comes from the function `all-modifier-codes` which, for some reason, does not collect the AltGr modifier code used by the layout.

It seems the key code passed by the layout through AltGr is `(108 . 16)`, I obtained this code by modifying the `key-loop` of `read-one-line` like so:

```lisp
(key-loop ()
  (with-focus (screen-input-window screen)
    (loop for key = (read-key-or-selection)
       do
          (progn
            (message "~A" key)
            (cond
               ((stringp key)
                ;; handle selection
                (input-insert-string input key)
                (draw-input-bucket screen prompt input))
                  ;; skip modifiers
               ((is-modifier (car key)))
               ((process-input screen prompt input (car key) (cdr key))
                (if (or (not require-match)
                        (match-input))
                    (return (input-line-string input))
                    (draw-input-bucket screen prompt input "[No match]" t))))))))

```

So `all-modifier-codes` misses the value `108`, it returns `(50 62 66 37 66 105 64 205 77 133 134 206 207 92 203)`. ***However***, if I run `all-modifier-codes` with the `us` layout enabled, it returns `(50 62 66 37 66 105 64 108 205 77 133 134 206 207 92 203)`.

I have to admit I am very hesitant forcing it into the result of `all-modifier-codes` somewhere, and it seems to be more of an issue from the `xlib` package than StumpWM itself.

Any opinion on this?

Phundrak:
I can confirm this is indeed an issue from the `xlib` package. Here is the output of `(xlib:modifier-mapping *display*)` with the `us` layout:
```lisp
(50 62)
(66)
(37 66 105)
(64 108 205)
(77)
NIL
(133 134 206 207)
(92 203)
```

[...]

PR #978, opened 2022-05-26
Add preliminary altgr support

szos:
This PR implements preliminary support for altgr key bindings. This wont allow all keys to be input in the input bar, but it will allow binding to keys that can only be pressed with altgr. For example, the `[` key lies on altgr+8 when using the no layout, and this allows entering that key in the input bar, as well as binding commands and keymaps to it. 

While this is 100% a problem that should be solved in CLX instead of here, there has been little progress on it and it seems unlikely to be implemented soon, so this is presented as a workaround to address issues such as #936. 

Certain characters fail to be looked up properly, but I'm unsure if this is an issue specific to my own setup or not. For example, while `lstroke` gets registered properly for binding keys, the keysym that it gets parsed into when entering into the input bar is 435, which doesnt return a character when passed to `xlib:keysym->character`. This may be related to many characters not being present in `xlib::*keysym->character-map*`. 

If this is best maintained as a set of personal patches I'll keep them in a diff to apply myself. 

EDIT: To be very clear, I have not tested this on multiple keyboard layouts outside of `no` and `us`, and I'm unsure how it will function on multiple keyboard layouts that make use of altgr. It will likely fail if the altgr offset is not correct, though this will just put the user back in the position of not having altgr bindings (or modifying the altgr offset when they change layouts).

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -225,6 +225,7 @@ Available completion styles include
       (= (xlib:keycode->keysym *display* keycode 0) 0)))
 
 (defun register-altgr-as-modifier ()
+  "Register the keysym(s) for ISO_Level3_Shift as modifiers."
   (setf *all-modifiers*
         (append (multiple-value-list
                  (xlib:keysym->keycodes *display*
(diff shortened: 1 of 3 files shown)


szos:
Upon exiting StumpWM and rebuilding, I found that this causes stumpw to fail to build. Im closing this for now while I work on better supporting altgr.

szos:
As it turns out the error was far simpler to address than I thought, so I'm reopening this, along with additional commits to add explanations to the manual.

PR #979, opened 2022-05-27
Support dead keys in the input bar

szos:
This adds support for inputting dead keys in the input bar. This works by calculating all the dead keys (via `define-dead-keysym`) when defining keysyms, and when reading input, if one of these keysyms is encountered it is checked against the most recently encountered dead key. If they are the same, the dead keys non-dead version is inserted. Otherwise it is stored in the input line as the most recently encountered dead key. When a non-dead key is encountered, then it is combined with the dead key. If this doesnt yield a valid character, then the encountered key is input as per usual, otherwise the combined character is entered. 

This restricts the user to only one dead key at a time, but I'm unsure if this is an issue as from a brief perusal it doesnt seem that StumpWM defines multi accent keys (eg I couldnt find ḉ or similar).

Here is an example of it in action, where every accent is input with a dead key:

https://user-images.githubusercontent.com/39150985/170715612-2ea751f8-b926-4806-a41c-a8aaa5be7070.mp4


montaropdf:
At last. Thanks for this patch. Does it also allow to use quote, double-quote, backquote, when they are dead keys?

szos:
> At last. Thanks for this patch. Does it also allow to use quote, double-quote, backquote, when they are dead keys?

It depends on how you press them. If they can be entered with only regular or shifted key presses then yes. If they require Alt Graph to input then you'll need the patch proposed in #978 as well.

montaropdf:
> > At last. Thanks for this patch. Does it also allow to use quote, double-quote, backquote, when they are dead keys?
> 
> It depends on how you press them. If they can be entered with only regular or shifted key presses then yes. If they require Alt Graph to input then you'll need the patch proposed in #978 as well.
> 
> Edit: I misunderstood the question. Yes you can use every dead key as a normal key by pressing it twice.
> demonstrate-dead-key-alone.mp4

Excellent.

Thanks for all your work.

PR #981, opened 2022-06-01
Make y-or-n-p and yes-or-no-p follow the standard

szos:
The standard dictates that `y-or-n-p` and `yes-or-no-p` take an optional format control string and arguments for that control string. The current implementations of `stumpwm::[y|yes]-or-[n|no]-p` take a single mandatory argument. This PR changes the argument list to match the arguments specified in the standard. This is backwards compatible in all cases except when the string passed has a `~`, which will now be interpreted as a format directive.

### DIFF
--- a/input.lisp
+++ b/input.lisp
@@ -939,20 +939,25 @@ input (pressing Return), nil otherwise."
 ;; (defun cook-keycode (code state)
 ;;   (values (xlib:keycode->keysym *display* code 0) (x11mod->stumpmod state)))
 
-(defun y-or-n-p (message)
+(defun y-or-n-p (&optional message &rest arguments)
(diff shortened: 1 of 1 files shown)


PuercoPop:
`stumpwm:y-or-n-p` and `lisp:y-or-n-p` are two different functions. Is there a reason/use case why we want to change the function to harmonize with the standardized implementation? I do worry about the backwards incompatible change. ~ is used in file paths, so I think it is likely it will affect someone.

szos:
> `stumpwm:y-or-n-p` and `lisp:y-or-n-p` are two different functions.

Ah, I had assumed the intent behind those functions was to imitate the standard functions but use StumpWM IO. Ill close this PR as theres no compelling reason to change.

PR #983, opened 2022-06-04
Implement Minor Modes for StumpWM

szos:
This PR implements minor modes for StumpWM which function similarly to Emacs minor modes, as requested in the StumpWM readme. 

Here is an example (also in the manual) of an evil minor mode for stumpwm:

```lisp
(define-minor-mode swm-evil-mode () ()
  (:interactive t)
  (:top-map ("i" . "swm-evil-mode")
            ("h" . "move-focus left")
            ("j" . "move-focus down")
            ("k" . "move-focus up")
            ("l" . "move-focus right")
            ("g" . *groups-map*)
            ("H n" . "evil-next-head"))
  (:expose-keymaps t)
  (:lighter "CMD")
  (:lighter-make-clickable nil)
  (:scope :global)
  (:make-hooks nil))

(define-swm-evil-mode-command evil-next-head () ()
  (let* ((gh (group-heads (current-group)))
         (curhead-list (member (current-head) gh)))
    (if (cdr curhead-list)
        (focus-frame (current-group)
                     (car (flatten (head-frames (current-group)
                                                (cadr curhead-list)))))
        (focus-frame (current-group)
                     (car (flatten (head-frames (current-group) (car gh))))))))

(define-key *top-map* (kbd "s-ESC") "swm-evil-mode")
```

I think this implements a robust mechanism for extending StumpWM, and could be used to implement contribs easily. Additionally I think this has a good deal of flexibility, not the least of which is the ability to scope minor modes to a window, frame, head, group, screen, or globally. For example, a minor mode could be defined that is only active on one head. While I'm unsure what the exact usage of such scopes would be, I was unable to determine the StumpWM concept that was analogous to Emacs buffers so I implemented a whole set of scopes - that way the user can have the choice. 

I've tested this with all scopes, but have only one head, and only one screen, so while I dont have any reason to thing there would be issues with those scopes, I cant say for certain. 

For clarity I've added an example and explanation of defining minor modes to the manual, but can add it here if desired. 

It should be noted that this was developed with the assumption that the clickable mode line PR will be merged.

Editing to add another example usage: traditional workspaces (ie one workspace per head) can be implemented using this minor mode mechanism. While I havent tested with multiple heads due to lack of monitors, I think this should work with multiple heads as well: https://gist.github.com/szos/77e844ef482a4a846a5380bfd96a11ce. Additionally I think dynamic tiling could have been implemented as a minor mode, but I'm not interested in rewriting it again should this get merged (although maybe at some point in the future).


szos:
The initial implementation of minor modes I wrote implemented minor modes as mixins, and used `dynamic-mixins` to add and remove them from group objects. While it would afford a greater versatility I ultimately decided against that implementation mainly because it would introduce another dependency. If introducing another dependency would be acceptable I would like to develop that implementation further and replace this PR with it. It affords versatility by allowing the minor mode to override and extend methods. The scopes contained in this PR could be implemented as well by changing the frame and head structures to classes. I didnt feel that the mixin implementation was appropriate as a contrib as it requires modifying several StumpWM internals, including changing structs to classes which violates the spec.

minor-mode.lisp
```diff
@@ -0,0 +1,545 @@
...
+          disable-minor-mode
+          minor-mode-on-enable
+          minor-mode-on-disable
+          minor-mode-keymap
+          minor-mode-lighter
+          generate-keymap))
+
+(defvar *minor-mode*)
+
+(defvar *active-global-minor-modes* nil
```
PuercoPop, line 16: minor style suggestion, write nil as () here as it is used as a list instead of a boolean
```suggestion
(defvar *active-global-minor-modes* ()
```

minor-mode.lisp
```diff
@@ -0,0 +1,545 @@
...
+
+(defun global-minor-modes (&optional ignore)
+  (declare (ignore ignore))
+  *active-global-minor-modes*)
+
+(defun set-global-minor-modes (ignore value)
+  (declare (ignore ignore))
+  (setf *active-global-minor-modes* value))
+
+(defsetf global-minor-modes set-global-minor-modes)
```
PuercoPop, line 27: For simple cases one can define a setf function directly. Although unless the plan is to add hooks for extension or something similar, wouldn't using the global variable directly be good enough?

```suggestion
(defun (setf global-minor-modes) (value ignore)
  (declare (ignore ignore))
  (setf *active-global-minor-modes* value))
```

szos: Accessing the global minor mode list through setter/getter functions was done to ease code generation in macros, see `define-enable-methods`. While the variable can be used directly, it does mean adding complexity to the enable/disable method generation.

szos:
Thank you for the review PuercoPop! Im curious what your thoughts are on the dynamic mixin approach to minor modes i described above? The more I think on it the more it seems like that is the best approach.

szos:
After more coding and looking at how this can be put together, I think the approach of dynamic mixins is the better approach, and could be implemented as a contrib module to avoid introducing a dependency. Alternatively the code from dynamic mixins could be incorporated in StumpWM as its licensed under the BSD 2 clause. 

In my branch [mixin-modes](https://github.com/szos/stumpwm/tree/mixin-modes) I added plumbing which would allow minor modes to hook into the `top-maps` function and the `command-active-p` functions, and converted frames/heads to classes to allow dynamically mixing in classes to them. 

In my branch [mixin-modes-in-main-trunk](https://github.com/szos/stumpwm/tree/mixin-modes-in-main-trunk) I added the mixin modes to stumpwm proper, but they could also be implemented as a contrib module given the plumbing in the mixin-modes branch. 

I'm curious in what the maintainers think of this approach; if it is considered a good approach then I'll close this PR and open a new one containing either just the plumbing or the full minor mode implementation.

minor-mode.lisp
```diff
@@ -0,0 +1,543 @@
...
+(defun find-minor-mode (mode group)
+  (car (minor-mode-enabled-p mode :group group)))
+
+(defun minor-mode-top-maps (group)
+  "Return a list of keymaps for all active minor modes."
+  (check-type group group)
+  (let ((minor-modes (list-minor-modes :group group)))
+    (flatten 
+     (loop for mode in minor-modes
+           collect (minor-mode-keymap mode)))))
```
PuercoPop, line 151: iiuc this is just a flatten + map, which is alexandria:mappend
```suggestion
    (alexandria:mappend (lambda (mode) (minor-mode-keymap mode)) minor-modes) 
```

PuercoPop:
> I've tested this with all scopes, but have only one head, and only one screen, so while I dont have any reason to thing there would be issues with those scopes, I cant say for certain.

protip: One can use Xephyr to test multiple heads. See for more details https://github.com/sulami/blog/blob/master/content/using-xephyr-to-simulate-multiple-monitors.md

I think vendoring the dynamic-mixings code would be best approach. If you decide to use CLOS inheritance as the minor mode registry.

IIUC this PR implements keymaps local to window, frame or head right? Ideally this would be achived by having the concept of a parent keymap or keymap mix-ins? This is definitely out of scope of this PR but I think it would let us do away with the top-maps concept.

One thing that I saw in a branch is moving toward classes and away from structs for frame. I think that is something we should do regardless. It can be part of a separate PR is it easier for you.

Sorry I don't have more substantive comments/recommendations, just wanted to say keep up to good work!

szos:
Thanks for the tip on Xephyr, Ill give it a look. 

> IIUC this PR implements keymaps local to window, frame or head right? Ideally this would be achived by having the concept of a parent keymap or keymap mix-ins? This is definitely out of scope of this PR but I think it would let us do away with the top-maps concept.

While thats one thing this PR implements, the goal is to set up a framework for easily extending StumpWM behavior. To that end I think this PR isn't the right direction to go, and the dynamic mixin approach coupled with a generification of StumpWM functions would be the best way of implementing such a framework. I'm going to close this PR and open a separate one with the vendorized dependency and a mixin based approach.  

> just wanted to say keep up to good work!

Thanks!

PR #985, opened 2022-06-16
Implement Minor Modes as Mixins

szos:
This PR implements minor modes as mixins to support overriding/augmenting the default behavior of StumpWM. This PR will be followed by one or more PRs to generify functions to allow minor modes to override/augment their behavior. 

I'm opening this as a draft PR because this PR is rather large in scope, and I would like feedback on how it functions. In particular, the implementation and design of the scope system feels shaky.

Minor modes implement a standard way of extending StumpWMs default behavior through methods, allowing modification without needing to hack on the source itself. The features include:

- top and root keymaps specific to the minor mode, which are active only when the minor mode is currently active. For example, a keymap that is only active when a window is focused. 
- hooks to be run when autoenabling, autodisabling, and explicitly enabling. 
- lighters per mode (a function that returns a string to print in the mode line, a la lighters in emacs minor modes).
- commands which are only active when the minor mode is currently active. 
- autogenerating a command which toggles the minor mode 
- above all else, the ability to augment generic functions which dispatch upon the object its mixed into. 

Some of the terms used when discussing minor modes: 
- autoenable/autodisable: Since minor modes can be global or local, enabling is differentiated from autoenabling. When a minor mode is enabled for an object automatically (eg when a minor mode is global) then only the autoenable function is called. The same goes for autodisable
- scope: a minor mode is scoped to a specific object type. scopes are arbitrary and can be as general as "all windows" or as specific as "only windows with the name X and resource Y". They can be defined by the user, but a set of them are already implemented and described in the manual. 

Minor modes are implemented as mixins and use `dynamic-mixins` to handle the actual mixing in. To avoid adding another quicklisp dependency this system is vendored and has some minor modifications to make it play nice with StumpWM. 

If we assume that all functions are generic and can have methods defined for them, we could define a topbar minor mode (the implementation of drawing the topbar is not specified):

```lisp
(defclass frame-topbar ()
  ((frame-topbar-height :initform 10 :accessor frame-topbar-height)))

(defmethod frame-display-height :around (group (frame frame-topbar))
  (let ((height (call-next-method)))
    (- height (frame-topbar-height frame))))

(defmethod frame-display-y :around (group (frame frame-topbar))
  (let ((y (call-next-method)))
    (+ y (frame-topbar-height frame))))

(defmethod (setf frame-window) :after (new (frame frame-topbar))
  (update-topbar frame))

(define-minor-mode frame-bar (frame-topbar minor-mode) ()
  (:global t)
  (:scope :frame)
  (:lighter "T-BAR")
  (:interactive frame-topbar-mode))

(defmethod autoenable-minor-mode :after ((mode (eql 'frame-bar)) (obj frame))
  (when (frame-window obj)
    (let* ((group (window-group (frame-window obj)))
           (windows (frame-windows group obj)))
      (mapc #'maximize-window windows))))

(defmethod autodisable-minor-mode :after ((mode (eql 'frame-bar)) (obj frame-bar))
  (when (frame-window obj)
    (let* ((group (window-group (frame-window obj)))
           (windows (frame-windows group obj)))
      (mapc #'maximize-window windows))))
```


szos:
> this seams a great inclusion to stumpwm.

Thanks!

> do you plan to provide a command to display the list of enable minor-mode?

I hadnt thought of such, but it would certainly be useful! and it can be easily implemented given the existing function `list-minor-modes`. Ill implement it and commit.

szos:
I have thought of a potential issue with this implementation; the core of enabling minor modes is the function `change-class`, and the hyperspec states "[…] a programmer must not use [change-class](http://www.lispworks.com/documentation/HyperSpec/Body/f_chg_cl.htm#change-class) inside a [method](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_m.htm#method) if any [methods](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_m.htm#method) for that [generic function](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_g.htm#generic_function) [access](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_a.htm#access) any [slots](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_s.htm#slot), or the results are undefined."

I think there is only one case where we risk violating the spec, but i'm unsure. Take the following example: 

We have a group `g`, and a minor mode `m` scoped to groups. When we enable `m` in `g` we define a new class `mg` which descends from both `m` and `g`, and as such has the same slots. This would imply that slots may be accessed safely since they are not removed. 

The place where we may encounter issues is in the generic function `autodisable-minor-mode` - specifically in after methods which reference a slot present in the minor mode. Because after methods will be called after changing the class of the object if any of them reference a slot that is no longer in the object that will be an issue. Hopefully someone more knowledgeable than I will see this and comment, and I'll also ask around for help understanding the specific behavior of SBCL in this instance.

sdilts:
I really like this! One of the things I think that Stumpwm is really missing is the ability to experiment with customization, and this should really help with that!

> I have thought of a potential issue with this implementation

Does this just affect this last bullet point as described below?

> above all else, the ability to augment generic functions which dispatch upon the object its mixed into

I know this is a big part of the proposal, but if it might not be viable, could it be merged in later? Just having all of the other listed features are worth a lot by themselves. It might help with the code review process too, but I haven't even looked at what you have yet.

szos:
> > I have thought of a potential issue with this implementation

> Does this just affect this last bullet point as described below?

Yes this only affects the ability to modify method dispatch via minor modes. Im unsure if this even is behavior that violates the hyperspec - the only instance I can think of that this would be an error would be the aforementioned after methods, but the hyperspec doesnt mention this situation outside of the note on the change class page. In my testing with this it hasnt raised any errors or presented any problems (on SBCL 2.2.4) so I'm inclined to say this isnt a violation, but I dont understand the issue enough to say for certain. I' m currently looking into ways of changing the enabling and disabling of minor modes to not use generic functions, in which case there would be no issue whatsoever.

szos:
The main issue is that the way I had envisioned users to initialize the minor modes was by adding after methods to `enable-minor-mode` and `autoenable-minor-mode`, which would expect to allow access to slots in the object without violating the spec. However one option would be to relegate setup to the enable hook, which while requiring a small change to how autoenabling is done would allow the setup to touch the slots. However as far as the code generated by `define-minor-mode` goes it doesnt access any slots in any objects.

szos:
Here is a fully fleshed out minor mode for a frame pseudo modeline: https://gist.github.com/szos/b5d678d2ad057b631df0e187797978fd

dmb2:
This is truly incredible! I like it a lot!  

My gut feeling is to get it merged and let people kick the tires.  I just tagged a release, so we've got a decent amount of time before the next release (hopefully November) to get any kinks worked out.

I'm not too worried about the scoping issues.  Maybe @PuercoPop could give a review of the code itself, but it looks pretty good to me.

Really, this is AWESOME!

szos:
> This is truly incredible! I like it a lot!

Thank you! 

> My gut feeling is to get it merged and let people kick the tires. I just tagged a release, so we've got a decent amount of time before the next release (hopefully November) to get any kinks worked out.

I'm glad weve got some time as theres some kinks to work out. I'm very interested in getting people using and abusing minor modes to try and get an idea of the weak points and poor design points. I think I've addressed the scopes issues I was having. I want to try and reimplement several contrib modules with minor modes to stress test it.

minor-modes.lisp
```diff
@@ -0,0 +1,1083 @@
...
+
+

+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (defun make-special-variable-name (mode name)
+    (intern (format nil "*~A-~A*" mode name)))
+  
+  (defun parse-minor-mode-options (options)
+    (let ((valid-options
+            '((:interactive            1)
```
PuercoPop, line 572:
```list
            '((:interactive      .     1)
...
```

A alist is a list of conses, not a list of lists. That is, each element is a single cons, with the car being the key and the cdr the value. You are using assoc down below, which is meant to operate on alists, so I think you intended to define an alist

minor-modes.lisp
```diff
@@ -0,0 +1,1083 @@
...
+              (:root-map               1)
+              (:top-map                1)
+              (:enable-when            t)
+              (:make-hooks             1)
+              (:default-initargs       t)
+              (:define-command-definer 1)))
+          (all-vals '())
+          (other-opts '()))
+      (flet ((collect-values (option)
+               (let ((argcount (cadr (assoc (car option) valid-options))))
```
PuercoPop, line 587: I think using destructuring-bind here would be more clear and avoid the need to use all the cons manipulation operations.

mode-line-formatters.lisp
```diff
@@ -117,6 +117,19 @@ fmt-highlight. Any non-visible windows are colored the
...
+  (let* ((screen (mode-line-screen ml))
+         (group (mode-line-current-group ml))
+         (head (mode-line-head ml))
+         (window (group-current-window group)))
+    (format nil "~{~A~^ ~}"
+            (loop for mode in (list-current-mode-objects
+                               :screen (mode-line-screen ml))
+                  for text = (minor-mode-lighter mode)
+                  unless (string= text "")
+                    collect text))))
```
PuercoPop, line 131: :thought_balloon: consider using with-output-to-string to avoid the double iteration when constructing the string

primitives.lisp
```diff
@@ -561,18 +562,131 @@ Use the window's resource class.
...
+  ((name
+    :initform ""
+    :accessor head-name
+    :initarg :name)))
+
+;; duplicate frame accessors for heads.
+(macrolet ((define-head-accessor (name)
+             (let ((pkg (find-package :stumpwm)))
+               `(progn
+                  (defgeneric ,(intern (format nil "HEAD-~A" name) pkg) (head)
```
PuercoPop, line 658:
```suggestion
                  (defgeneric ,(sb-int:package-symbolicate pkg "HEAD-" name)
```

consider using package-symbolicate, although idk how the SBCL people would feel abou using their internal utility package ourselves :thinking:

primitives.lisp
```diff
@@ -561,18 +562,131 @@ Use the window's resource class.
...
+    :accessor swm-class-new-objects
+    :allocation :class
+    :documentation
+"Track all newly created objects in order to mix in the appropriate minor modes
+when they are touched")))
+
+(defmethod initialize-instance :after ((obj swm-class) &key &allow-other-keys)
+  ;; Register all newly created objects so that they can have the relevant minor
+  ;; modes autoenabled.
+  (pushnew obj (swm-class-new-objects obj) :test #'eq))
```
PuercoPop, line 577: Does it make sense to track them in a list? Can't it be an issue when too many objects are tracked? Have you considered using a hash-table instead? You can use a hash-table as a set by using the element you care about as the key and something like t as the value.

PuercoPop (commented):
Left a couple of comments, the no-applicable-method one being the most important

minor-modes.lisp
```diff
@@ -0,0 +1,1083 @@
...
+  nil)
+
+(defgeneric minor-mode-hook (minor-mode-symbol)
+  (:documentation
+   "Returns the minor mode hook for a given minor mode symbol. This hook is run
+whenever the minor mode is explicitly enabled."))
+
+(defmethod no-applicable-method ((f (eql #'minor-mode-hook)) &rest rest)
+  (declare (ignore f rest))
+  nil)
```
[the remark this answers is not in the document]
szos: The goal for many of these was to default to doing nothing. For this one specifically we want to run the hook returned by this function. But if a user calls this function with a non-minor mode symbol then nil will be returned and interpreted as a hook with no functions. Perhaps it would be better to error if a non minor mode symbol is passed to these hook functions?

minor-modes.lisp
```diff
@@ -0,0 +1,1083 @@
...
+             (cond ((typep object minor-mode)
+                    (restart-case 
+                        (error 'minor-mode-enable-error :mode minor-mode
+                                                        :object object
+                                                        :reason 'already-enabled)
+                      (continue () nil)))
+                   ((autoenable-minor-mode minor-mode object)
+                    (unless run-hook
+                      (setf run-hook object))))))
+      (mapc #'enable
```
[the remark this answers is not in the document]
szos: Yes that was the intent. Ill change it to `(map nil ...)`.

minor-modes.lisp
```diff
@@ -0,0 +1,1083 @@
...
+
+(defun sync-all-minor-modes ()
+  "Loop through all recently created objects and ensure that the appropriate minor
+modes are enabled in them, then nullify the list of objects."
+  ;; This functions is needed because calling autoenable-minor-mode from within
+  ;; a method that accesses slots is implied to be undefined behavior, so we
+  ;; cant do this from within initialize-instance.
+  (let ((objects (prog1 (swm-class-new-objects (current-screen))
+                   (setf (swm-class-new-objects (current-screen)) nil))))
+    (when (and objects *active-global-minor-modes*)
```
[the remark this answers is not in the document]
szos: I'm unsure what you mean by `*active-global-minor-modes*` being disabled - it is intended to hold a list of all enabled minor modes that are also global, not neccessarily be a minor mode which is enabled/disable. The goal here is to get the list of new objects and then loop through all the enabled minor modes and enable them in objects where appropriate. If we have no active global minor modes then we dont need to do the loop, same if we have no new objects. Regardless we want to nullify the list of new objects.

minor-modes.lisp
```diff
@@ -0,0 +1,1083 @@
...
+;; (defmethod change-class :after (object (new-class standard-class) &rest rest)
+;;   (declare (ignore rest))
+;;   (labels ((class-search (subclass)
+;;              (loop for class in (sb-mop:class-direct-superclasses subclass)
+;;                    when (or (eql (class-name class) 'swm-class)
+;;                             (class-search class))
+;;                      do (setf *my-tracker* object)
+;;                         (sync-minor-modes object)
+;;                         (return-from class-search nil))))
+;;     (class-search new-class)))
```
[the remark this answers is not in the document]
szos: Absolutely. This is still a work in progress, hence me opening it as a draft PR.

minor-modes.lisp
szos: While it currently is an alist whose values are lists of length 1, I wasnt sure when writing all of this if more values would be useful for the options (eg type information). Ill change this to a proper alist given that I dont see other information being particularly useful.

primitives.lisp
szos: Im not familiar with package-symbolicate, but I'll take a look. I'm hesitant to use internals given that it may change without warning and across versions, and some folks might use a newer/older version of SBCL.

primitives.lisp
szos: I think a list is appropriate here because whenever we access the list we walk the entire length to handle all objects before nullifying it. In practice the list seldom grows to more than a handful of elements. When objects are created via `make-swm-class-instance` they have the relevant global minor modes enabled in them, and because they are always towards/at the front of the list grabbing them is performant. If we were to change to hash tables we would be using maphash or similar.

minor-modes.lisp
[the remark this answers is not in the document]
szos: After thinking about this a little more, the proper course of action is likely to remove these methods and add appropriate restart bind/case and handler bind/case forms to the enabling/disabling/hook-running code. I will work out where that is and after adding that remove these methods.

minor-modes.lisp
szos: I'm not sure I quite follow - do you mean to use destructuring-bind on the options we receive? I'm not sure that would work, as the options are received in the form `((:o1 a b c) (:o2 d e f))`, and I dont think destructuring bind can handle nested lists in an arbitrary order like that. I may not be understanding you correctly.

minor-modes.lisp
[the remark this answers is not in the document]
PuercoPop: > I'm unsure what you mean by *active-global-minor-modes* being disabled
It was my bad, for some reason I was thinking it was a toggle instead of a list.

> Regardless we want to nullify the list of new objects.
Ok, then the code is correct as is. :+1:

minor-modes.lisp
PuercoPop: I meant instead of stuff like (cadr option), (car option) etc. Using d-bind we can bind the key and value of the current option at the start of the function, instead of the let.

Also iiuc we don't need to use assoc as the count would be the cdr of the current option being processed right?

szos: Ah I was misunderstanding you. Yes d-bind would work there and I'll add it in. I do think we need to use assoc as this is constructing a list to (eventually) pass to d-bind in the main `define-minor-mode` macro. I want to keep things open ended with regards to adding options to this macro, and if we ditched assoc we would have no way to discriminate between a future 2 argument option and an option with an arbitrary number of arguments. Since this is all happening at macroexpansion time it doesnt have all that much impact on performance.

PuercoPop: > I do think we need to use assoc as this is constructing a list to (eventually) pass to d-bind in the main define-minor-mode macro. I want to keep things open ended with regards to adding options to this macro, and if we ditched assoc we would have no way to discriminate between a future 2 argument option and an option with an arbitrary number of arguments.

I may be missing something, but isn't the open ended design achieved by the encoding the semantics in data (the alist)? That description is then processed one at the time. But if we change the valid-options alist we still need to recompile for it take effect. Wether we use the assoc or not.

szos: I think I understand what you're getting at (correct me if im wrong); Since we want to return the `cadr` of 1 arg options and the `cdr` of n arg options we can change options to be passed as `(:foo . bar)` for 1 arg options and `(:baz qux)` for n arg options, and then just take the cdr of all options instead of looking them up in an alist. Is that what you're talking about? 

I am somewhat attached to using an alist - this parsing was heavily inspired by the function `parse-define-application-frame-options` from the [mcclim codebase](https://github.com/McCLIM/McCLIM/blob/master/Core/clim-core/frames/define-application-frame.lisp#L138), and having the ability to easily add additional information without restructuring the core of how we parse options is something I'd like to hold on to. For example if in the future we wish to add the option `:define-hooks` which defines only specific hooks and conflicts with `:make-hooks`, we could add conflict checking by modifying the alist and `collect-values`.

events.lisp
```diff
@@ -220,19 +220,28 @@ The Caller is responsible for setting up the input focus."
...
-   ;; TODO: Minor Mode maps go here
+   ;; If a minor mode map element is a function or a symbol that does not denote
+   ;; a keymap but is fbound, then funcall it with the group. Otherwise it
+   ;; should be a kmap object or a symbol bound to a kmap object.
+   (loop for map in *minor-mode-maps*
+         if (or (functionp map)
+                (and (symbolp map)
+                     (not (and (boundp map)
+                               (kmap-p (symbol-value map))))
+                     (fboundp map)))
```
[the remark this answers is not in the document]
szos: I extracted this to a local function, since I couldnt see a use for it outside this specific scenario.

minor-modes.lisp
PuercoPop: >  and then just take the cdr of all options instead of looking them up in an alist. Is that what you're talking about?

yes, this is what I mean.

> I am somewhat attached to using an alist

I haven't suggested otherwise, I think using an alist is a great idea. Pushing semantics to a declarative description is great in my book. My issue is what you mentioned above, that we don't need to perform a lookup as we can get all the information we want by d-binding the description.

> this parsing was heavily inspired by the function parse-define-application-frame-options from the [mcclim codebase](https://github.com/McCLIM/McCLIM/blob/master/Core/clim-core/frames/define-application-frame.lisp?rgh-link-date=2022-06-23T08%3A05%3A44Z#L138),

interesting the linked code uses an alist of (almost) plists, which is why it uses getf to access information inside the description. In our case the value of the mapping is a single value so no need to.

What do you think?

szos: I'm sorry, I think I must be misunderstanding you, either in part or in full. I thought you were referring to dbinding the `option` variable within the local function `collect-values`. Do you think you could give me a small code example of what you intend with dbind? 

To my understanding this branch of the conversation came from you saying we dont need to use assoc, but if we dont use assoc then the alist will never get used, so any information contained there would be best put in a comment or docstring. But then you say that using the alist is a good idea, so I'm fairly certain I have misunderstood something along the way. 

Sorry this communication is proving so difficult; if its useful we could "reset" the thread and just start from the beginning again.

As a side note, I dont think taking options as described in my prior comment would be the smoothest experience for people defining minor modes, eg i prefer
```lisp
(define-minor-mode mymode ()
  (:scope :frame)
  (:enable-when (mode obj)
   (something obj)))
```
as opposed to 
```lisp
(define-minor-mode mymode () 
  (:scope . :frame)
  (:enable-when (mode obj)
   (something obj)))
```

minor-modes.lisp
```diff
@@ -0,0 +1,1113 @@
...
+                                       (= argcount 1))
+                                  (push (car option-arguments) all-vals)
+                                  (push option-arguments all-vals))
+                              (push option-name all-vals))
+                       (push option other-opts))))))
+        (mapc #'collect-values options)
+        (values all-vals other-opts))))
+
+  (defun define-command-macro (mode)
+    `(defmacro ,(intern (string-upcase (format nil "define-~A-command" mode)))
```
PuercoPop, line 600: We can upcase inside the format call
```suggestion
    `(defmacro ,(intern (format nil "~:@(define-~A-command~)" mode))
```

another alternative would be to use alexandria:symbolicate, `(alexandria:symbolicate "DEFINE-" mode "-COMMAND")`

minor-modes.lisp
```diff
@@ -0,0 +1,1113 @@
...
+         (name (&rest args) (&rest interactive-args) &body body)
+       (multiple-value-bind (bod decls docstring)
+           (parse-body body :documentation t)
+         `(defcommand (,name ,',mode) ,args ,interactive-args
+            ,@(when docstring
+                `(,docstring))
```
PuercoPop, line 606: stylenit: The way we write the idiom in StumpWM is (list ...). I've the the `(,...) way of 'spelling' it though and there is nothing wrong with that 

```suggestion
                (list docstring))
```

minor-modes.lisp
```diff
@@ -0,0 +1,1113 @@
...
+                                  (push option-arguments all-vals))
+                              (push option-name all-vals))
+                       (push option other-opts))))))
+        (mapc #'collect-values options)
+        (values all-vals other-opts))))
+
+  (defun define-command-macro (mode)
+    `(defmacro ,(intern (string-upcase (format nil "define-~A-command" mode)))
+         (name (&rest args) (&rest interactive-args) &body body)
+       (multiple-value-bind (bod decls docstring)
```
PuercoPop, line 602: what do you think about just shadowing body instead of using bod because body is already in scope?
```suggestion
       (multiple-value-bind (body decls docstring)
```

minor-modes.lisp
```diff
@@ -0,0 +1,1113 @@
...
+              (:root-map               . 1)
+              (:top-map                . 1)
+              (:enable-when            . t)
+              (:make-hooks             . 1)
+              (:default-initargs       . t)
+              (:define-command-definer . 1)))
+          (all-vals '())
+          (other-opts '()))
+      (flet ((collect-values (option)
+               (destructuring-bind (option-name &rest option-arguments) option
```
PuercoPop, line 587:
```suggestion
               (destructuring-bind (option-name . option-arguments) option
```
We can use the cons literal notation here, no need &rest

minor-modes.lisp
```diff
@@ -0,0 +1,1113 @@
...
+                 (let ((argcount (cdr (assoc option-name valid-options))))
+                   (if argcount
+                       (progn (if (and (numberp argcount)
+                                       (= argcount 1))
+                                  (push (car option-arguments) all-vals)
+                                  (push option-arguments all-vals))
+                              (push option-name all-vals))
+                       (push option other-opts))))))
```
PuercoPop, line 595: We can use the if-let idiom here

```lisp
 (alexandria:if-let ((argcount (cdr (assoc option-name valid-options))))
   (progn (if (and (numberp argcount)
                   (= argcount 1))
              (push (car option-arguments) all-vals)
              (push option-arguments all-vals))
          (push option-name all-vals))
   (push option other-opts))
```

szos: This has been implemented.

minor-modes.lisp
szos: done and done :smile:

minor-modes.lisp
szos: I think its a good move, and I implemented it.

szos:
I'm very sorry, but I wasnt quite ready for this to be merged yet. Specifically I was in the process of fixing a build issue with dynamic mixins. I will open a PR with emergency changes to ensure that StumpWM can still build.

PR #986, opened 2022-06-18
Fix fullscreen frame

ghost-of-freedom:
hey, i wrote a big wall of text here, but wonderful firefox browser crashed the tab with no survivors, so forgive me if this is not complete or it sounds bad

i tried out the code in #974, and it didn't really work well for me, it left space at the top when i fullscreened a frame. probably has to do with the fact that @szos has the modeline at the top, and i have it at the bottom. i modified the code so that there's no manual checking of the modeline height and stuff, since there are frame-display functions to handle all that. there was also a comment saying there is a check on whether we are fullscreening a frame or head, but there was no such check, so i added it. the code should probably be changed a bit or something, i don't know lisp so i don't know what is good style/performant.

also, do you think it would be possible to decouple the frame fullscreening functionality from the head fullscreening? for example, if you wanted all the frames to be fullscreened by default, the only way i can currently think of is creating a predicate that always returns true, but this makes it impossible to make a window fullscreened for the head

### DIFF
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -128,7 +128,7 @@ than the root window's width and height."
              (setf x (frame-x frame-to-fill)
                    y (frame-y frame-to-fill)
                    width (frame-width frame-to-fill)
-                   height (frame-height frame-to-fill)))
+                   height (frame-height frame-to-fill))))
        (return-from geometry-hints (values x y 0 0 width height 0 t)))
       ;; Adjust the defaults if the window is a transient_for window.
       ((find (window-type win) '(:transient :dialog))
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -116,20 +116,19 @@ than the root window's width and height."
        (let* ((win-group (window-group win))
               (fs-in-frame (fullscreen-in-frame-p win))
               (head (frame-head win-group f))
-              (frame-to-fill (if fs-in-frame f head))
-              (ml (head-mode-line head))
-              (adjust-by (if (and fs-in-frame
-                                  (or (eq :stump (mode-line-mode ml))
-                                      (eq :visible (mode-line-mode ml))))
-                             (mode-line-height ml)
-                             0)))
-         ;; determine if the window should be fullscreened in the frame or the
-         ;; head. When fullscreen-in-frame-p returns true, use the current frame
-         ;; and adjust for mode line size. 
-         (setf x (frame-x frame-to-fill)
-               y (+ (frame-y frame-to-fill) adjust-by)
-               width (frame-width frame-to-fill)
-               height (- (frame-height frame-to-fill) adjust-by)))
+              (frame-to-fill (if fs-in-frame f head)))
+         ;; Determine if the window should be fullscreened in the frame or the
+         ;; head. If fullscreening a frame, use the frame-display functions on
+         ;; y axis to account for the modeline.
+         (if fs-in-frame
+             (setf x (frame-x frame-to-fill)
+                   y (frame-display-y win-group frame-to-fill)
+                   width (frame-width frame-to-fill)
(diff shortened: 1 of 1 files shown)


ghost-of-freedom:
also maybe the variable names could be changed, the 1-2 character identifiers make it a pain to read the code

szos:
> also, do you think it would be possible to decouple the frame fullscreening functionality from the head fullscreening? for example, if you wanted all the frames to be fullscreened by default, the only way i can currently think of is creating a predicate that always returns true, but this makes it impossible to make a window fullscreened for the head

I'm not sure I understand what youre wanting here; frames cant be fullscreened, only windows can. If you could explain in more detail that would be helpful. Assuming you mean that all windows should be fullscreened in frame by default, and then override that, you could do that in any number of ways, for example using window marks: 

```lisp 
(define-fullscreen-in-frame-rule fs-in-head-when-marked (w)
  (window-marked w))
```

ghost-of-freedom:
> I'm not sure I understand what youre wanting here; frames cant be fullscreened, only windows can.

yeah i guess i used misleading terminology, sorry

> If you could explain in more detail that would be helpful. Assuming you mean that all windows should be fullscreened in frame by default, and then override that, you could do that in any number of ways, for example using window marks:
> 
> ```lisp
> (define-fullscreen-in-frame-rule fs-in-head-when-marked (w)
>   (not (window-marked w)))
> ```

this would work with a fullscreen function on `*new-window-hook*`, but i don't know if it wouldn't be annoying if you use the marks for something else

> I think what youve written is perfectly fine performance wise, and is formatted well.

nice, in that case gnu emacs should get the credit where it's due ;)

szos:
> this would work with a fullscreen function on `*new-window-hook*`

Ah, so if I understand correctly you want to automatically fullscreen every window as its created, and then have the option of changing that "fullscreenness" from the frame to the head? That could be achieved by a separate command, like so: 

```lisp
(defvar *myvar* t)

(define-fullscreen-in-frame-rule myrule (w)
  (declare (ignore w))
  *myvar*)

(defcommand fs-in-head () ()
  (let ((*myvar* nil)
        (win (current-window)))
    (when (window-fullscreen win)
      (update-fullscreen win 0))
    (update-fullscreen win 1)))
``` 

Have I understood you correctly?

ghost-of-freedom:
> Ah, so if I understand correctly you want to automatically fullscreen every window as its created, and then have the option of changing that "fullscreenness" from the frame to the head? 

yeah, just have it all start maximized in a frame, and have the option to make it fullscreened to the head if required

> 
> ```lisp
> (defvar *myvar* t)
> 
> (define-fullscreen-in-frame-rule myrule (w)
>   (declare (ignore w))
>   *myvar*)
> 
> (defcommand fs-in-head () ()
>   (let ((*myvar* nil)
>         (win (current-window)))
>     (when (window-fullscreen win)
>       (update-fullscreen win 0))
>     (update-fullscreen win 1)))
> ```

i've tried this out, but i don't think it makes new windows fullscreen in the frame

szos:
Yes this is just the portion on fullscreening in the head. Fullscreening every window via new window hook is simple enough i didnt think to include it.

ghost-of-freedom:
right, my bad. if this is possible then i guess there's no need to decouple them after all
i still think it would be good to make the variable names more verbose though

szos:
Which variable names in particular? I dont know the maintainers stance on such changes, but you could try making the change in this PR if the variables are local to the function you patched.

dmb2:
Open another issue for the length of variable names with specific examples.  I think in this case @ghost-of-freedom wants 'fs' to be expanded to 'fullscreen'.  I don't have feelings one way or the other.  I've swung both ways over the years, but I tend towards verbose names unless they're well defined and contextual (x, y, i, j, k, etc).

Issue #669, opened 2019-12-13
Terminal borders, empty frame indicator behave oddly with compositing


paulbdavis:
I am using picom (compton) to add transparency to emacs and terminals, but for some reason the terminal compositing behaves differently than that of other windows. It presents on all terminals I've tried except xterm and urxvt

I'm not totally sure if this issue lies in stump or in compton/picom, but since the issue has not presented on i3 from my previous setup I think something is going on in stump that is causing them to be treated differently

If there is no rule in the compton conf for the terminal at all, it still causes the borders to go nearly transparent, just having the compositor enabled at all causes it.

I've also noticed that the dashed line indicating being in an empty frame also disappears when compositing is turned on 

Here you can see the border is given extra transparency 
![emacs](https://user-images.githubusercontent.com/2730634/70823911-3c2f6a00-1d9e-11ea-93af-0b1bd06ff1cb.jpg)
Versus the way the emacs border looks when inactive
![termite](https://user-images.githubusercontent.com/2730634/70823915-3cc80080-1d9e-11ea-97ee-16c26a0ffb02.jpg)
When it is not floating there is a "gap" caused by the overly transparent border (hard to see here, but the emacs window in the lower right has it's inactive color extending further than the terminal window above it)
![term-full](https://user-images.githubusercontent.com/2730634/70824684-fa072800-1d9f-11ea-939a-d5c0b7d34530.jpg)
Here are shots of the termial and emacs in i3 using the same `compton.conf` file
![i3-term](https://user-images.githubusercontent.com/2730634/70823912-3c2f6a00-1d9e-11ea-91ce-8de402945e54.jpg)
![i3-term-float](https://user-images.githubusercontent.com/2730634/70823913-3c2f6a00-1d9e-11ea-9cb1-ba1d0b04b57f.jpg)

equwal:
This is related to https://github.com/stumpwm/stumpwm/issues/532. Adding support for CLX composite could fix a lot of rendering issues from forcing the use of the hardware device.

paulbdavis:
Interesting, I'll have to hack on this when I have some "free time"...

equwal:
[CLX composite](https://github.com/sharplispers/clx/blob/master/extensions/composite.lisp)
[Xorg documentation](https://www.x.org/releases/X11R7.7/doc/compositeproto/compositeproto.txt)

dmb2:
[...]

PR #996, opened 2022-07-20
Add an alpha channel to some allocated colors

jorams:
This is a hacky, partial fix for #669, also experienced by elais[m] on IRC, in which enabling a compositor turns the StumpWM window background color transparent. It's based on [a similar fix](https://github.com/herbstluftwm/herbstluftwm/pull/1178) in herbstluftwm.

Since this patch changes the result of a call to `xlib:alloc-color`, which involves a request to X, I suspect it to break on systems that don't use 24-bit colors. I'm not sure how to properly test that.

It would probably be better to either do this in our `alloc-color` or to somehow fix this in CLX.

### DIFF
--- a/color.lisp
+++ b/color.lisp
@@ -106,9 +106,10 @@ then call (update-color-map).")
                           color))))
 
 (defun alloc-color (screen color)
-  (xlib:alloc-color 
-   (xlib:screen-default-colormap (screen-number screen))
-   (lookup-color screen color)))
+  ;; We add an alpha channel to the color returned by xlib:alloc-color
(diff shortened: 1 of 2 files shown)


dmb2:
the hole codebase is hacky in some sense.. Give me a good reason not to merge.  CLX isn't changing easily.

jorams:
I'm going to move the transformation from `set-any-color` to our `alloc-color`, and then I think this is ready for merge.

I'm still a bit worried since I don't know in what situation this would break and what the effect would be, but I guess we'll find out.

jorams:
This should be good now. The result is the same, but this is a better place to do it.

Issue #840, opened 2020-12-27
unhandled error: 


erjoalgo:
I don't have more details to reproduce but I ran into this crash sometime today:

```
("^B^1*StumpWM Crashed With An Unhandled Error!"
  "Copy the error to the clipboard with the 'copy-unhandled-error' command."
  "^bThe value" "    NIL" "  is not of type" "    STUMPWM::FRAME^B^n"
  "Backtrace for: #<SB-THREAD:THREAD \"main thread\" RUNNING {10005884C3}>"
  "0: ((:METHOD GROUP-LOST-FOCUS (STUMPWM::TILE-GROUP)) #<STUMPWM::TILE-GROUP {100574D253}>) [fast-method]"
  "1: ((:METHOD GROUP-REMOVE-HEAD (STUMPWM::TILE-GROUP T)) #<STUMPWM::TILE-GROUP {100574D253}> #S(frame 2 NIL 0 0 1360 768)) [fast-method]"
  "2: (STUMPWM::REMOVE-HEAD #S<screen #<XLIB:SCREEN :0.0 1600x900x24 TRUE-COLOR>> #S(frame 2 NIL 0 0 1360 768))"
  "3: (STUMPWM::SCALE-SCREEN #S<screen #<XLIB:SCREEN :0.0 1600x900x24 TRUE-COLOR>> (#S(frame 0 NIL 0 0 1600 900)))"
  "4: (STUMPWM::HEAD-FORCE-REFRESH #S<screen #<XLIB:SCREEN :0.0 1600x900x24 TRUE-COLOR>> (#S(frame 0 NIL 0 0 1600 900)))"
  "5: ((LAMBDA (&REST #:G0 &KEY :STACK-MODE :WINDOW :X :Y :WIDTH :HEIGHT :BORDER-WIDTH :VALUE-MASK &ALLOW-OTHER-KEYS) :IN \"/home/ealfonso/src/stumpwm/events.lisp\") :STACK-MODE NIL :WINDOW #<XLIB:WINDOW :0 111> :X 0 :Y 0 :WIDTH 1360 :HEIGHT 768 :BORDER-WIDTH 0 :VALUE-MASK NIL)"
[...]

Issue #763, opened 2020-04-07
stump crashes when switching between monitors with autorandr


m-g-r:
When switching between monitor setups with mutual exclusive screens, that is, I switch off the notebook's screen when plugging in an external monitor, StumpWM crashes. I currently use autorandr for that task but StumpWM is not happy with the way autorandr does his job.

Here is a screenshot of the crash message:
![2020-04-08-000233_2560x1440_stumpwm-crash-on-autorandr](https://user-images.githubusercontent.com/7143985/78724498-e32ba900-792d-11ea-973a-6024157eac5b.png)

I'd assume this happens when there is for a short time no head available but when I try to simulate the situation manually with just switching all screens of with xrand, nothing happens.

I can also try fixing it myself but I thought I ask here first. Maybe someone has fixed this already privately. :-)

m-g-r:
Ok, I looked at this again, and autorandr just uses xrandr. (Autorandr has it's own problems and seems not to switch screens as it promises (see https://github.com/phillipberndt/autorandr/issues/ 193) but even if it didn't its job as specified it would crash stumpwm.)

Long story short, this means I can cause the same crash with xrandr only:
> xrandr --output DP-2-2 --mode 2560x1440 --output DP-2-1 --off

results in the unhandled error: "The value NIL is not of type STUMPWM:FRAME" i method (GROUP-LOST-FOCUS (STUMPWM:TITLE-GROUP)) (see screenshot in first comment).

That is,
1. I have my internal screen DP-2-1 switched on and DP-2-2 switched off
2. I invoke above xrandr call to switch DP-2-1 off and DP-2-2 on _at the same time_
3. stumpwm crashes _reliably_.

Even if I do the screen flipping _one after the other_ without a delay like this:
> xrandr --output DP-2-2 --mode 2560x1440 ; xrandr --output DP-2-1 --off

stumpwm crashes _sometimes_.

But if I add a short `sleep .5` in between, stumpwm crashed _never_ for me.

So I can workaround this problem but stumpwm should be able to handle the situation of a clean switch from one screen to another, right?

PuercoPop:
>  stumpwm should be able to handle the situation of a clean switch from one screen to another, right?

Yes.

<hr >

@m-g-r Thanks for looking it, From your report and it seems the issue is that in the short time there is no monitor the windows have no frame and some code that is called by group-remove-head assumes windows always have a frame. It looks to be that the code in question is `group-lost-focus` by the way of `hide-window`. 

[...]

Issue #841, opened 2020-12-27
unhandled error: The value" NIL is not of type  STUMPWM::FRAME


erjoalgo:
I don't have more details to reproduce but I ran into this crash sometime today:

```
("^B^1*StumpWM Crashed With An Unhandled Error!"
  "Copy the error to the clipboard with the 'copy-unhandled-error' command."
  "^bThe value" "    NIL" "  is not of type" "    STUMPWM::FRAME^B^n"
  "Backtrace for: #<SB-THREAD:THREAD \"main thread\" RUNNING {10005884C3}>"
  "0: ((:METHOD GROUP-LOST-FOCUS (STUMPWM::TILE-GROUP)) #<STUMPWM::TILE-GROUP {100574D253}>) [fast-method]"
  "1: ((:METHOD GROUP-REMOVE-HEAD (STUMPWM::TILE-GROUP T)) #<STUMPWM::TILE-GROUP {100574D253}> #S(frame 2 NIL 0 0 1360 768)) [fast-method]"
  "2: (STUMPWM::REMOVE-HEAD #S<screen #<XLIB:SCREEN :0.0 1600x900x24 TRUE-COLOR>> #S(frame 2 NIL 0 0 1360 768))"
  "3: (STUMPWM::SCALE-SCREEN #S<screen #<XLIB:SCREEN :0.0 1600x900x24 TRUE-COLOR>> (#S(frame 0 NIL 0 0 1600 900)))"
  "4: (STUMPWM::HEAD-FORCE-REFRESH #S<screen #<XLIB:SCREEN :0.0 1600x900x24 TRUE-COLOR>> (#S(frame 0 NIL 0 0 1600 900)))"
  "5: ((LAMBDA (&REST #:G0 &KEY :STACK-MODE :WINDOW :X :Y :WIDTH :HEIGHT :BORDER-WIDTH :VALUE-MASK &ALLOW-OTHER-KEYS) :IN \"/home/ealfonso/src/stumpwm/events.lisp\") :STACK-MODE NIL :WINDOW #<XLIB:WINDOW :0 111> :X 0 :Y 0 :WIDTH 1360 :HEIGHT 768 :BORDER-WIDTH 0 :VALUE-MASK NIL)"
[...]

PR #1008, opened 2022-08-03
Made scale-screen add heads before removing heads

tiberious726:
This resolves the crash in #647 and #763 . 

Ultimately, this happens because `group-remove-head` can't handle there not being any frames in a group. If you try to remove the only head, `group-remove-head` will delete its frames from the group's `tile-group-frame-tree`, leaving it empty, then it will try to grab the group's first frame `;; just set current frame to whatever.`, which will, unfortunately, be nil, and proceed to set all of the windows' window-frame to nil, causing this problem.

https://github.com/stumpwm/stumpwm/blob/master/tile-group.lisp#L196-L208

This pull request solves the problem by simply preventing there from not being any heads during normal operation: adding new heads before removing the old ones if both happen in a single `:CONFIGURE-NOTIFY ` event.

I do, however, think that stumpwm ought be able to handle not having any actual heads attached, at least temporarily, and this patch is really papering over the issues with that (and, in fact, prevents `tile-window`'s frame slots from being set to nil if you have safety turned up on sbcl).

Furthermore, it doesn't looks like frames are properly adopted and scaled when moving between heads of different resolution.

### DIFF
--- a/head.lisp
+++ b/head.lisp
@@ -169,10 +169,10 @@
           (when (= (head-number oh) (head-number nh))
             ;; Same frame number, probably the same head
             (setf (head-number nh) (head-number oh))))))
-    (dolist (h (set-difference oheads heads :test '= :key 'head-number))
-      (remove-head screen h))
     (dolist (h (set-difference heads oheads :test '= :key 'head-number))
       (add-head screen h))
+    (dolist (h (set-difference oheads heads :test '= :key 'head-number))
+      (remove-head screen h))
     (dolist (h (intersection heads oheads :test '= :key 'head-number))
       (let ((nh (find (head-number h) heads  :test '= :key 'head-number))
             (oh (find (head-number h) oheads :test '= :key 'head-number)))
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -200,6 +200,7 @@
     (setf (tile-group-frame-tree group) (delete frames-to-delete group-frame-tree))
     ;; Just set current frame to whatever.
     (let ((frame (first (group-frames group))))
+      (unless (frame-p frame) (error "Couldn't locate a frame in group ~A" group))
       (setf (tile-group-current-frame group) frame
             (tile-group-last-frame group) nil)
       ;; Hide its windows.
--- a/tile-window.lisp
+++ b/tile-window.lisp
@@ -9,7 +9,7 @@
 like xterm and emacs.")
 
 (define-swm-class tile-window (window)
-  ((frame   :initarg :frame   :accessor window-frame)
+  ((frame   :initarg :frame   :accessor window-frame :type frame)
    (normal-size :initform nil :accessor window-normal-size)))
 
 (defmethod print-swm-object ((object tile-window) stream)


tiberious726:
I got a little bit further in tracking down the other issue tonight:

Visually, there is a frame that is correctly sized for the old head's resolution, but taking up the same visual space as the correctly sized frame (above or below them). You can create new frames incorrectly configured by generating them while this frame is highlighted, and cycle-through them with next-window. You can "jump" to a correctly configured frame by raising one of its windows.

`
STUMPWM>  (tile-group-frame-tree (current-group))
(#<FRAME 2 #<TILE-WINDOW "*sly-mrepl for sbcl*" #xA0013E> 0 0 2560 1440>)
STUMPWM> (window-frame (current-window))
#<FRAME 2 #<TILE-WINDOW "*sly-mrepl for sbcl*" #xA0013E> 0 0 2560 1440>
STUMPWM> (window-frame (current-window))
#<FRAME 0 #<TILE-WINDOW "*sly-mrepl for sbcl*" #xA002A2> 0 0 1920 2160>
`
The last two commands were run from repls different windows (the last being from the one in the messed-up frame). Notice the sizes.

Some windows are left associated with frames no longer in any group's frame-tree (and other than the per-group frame trees I don't think stump keeps a global list of frames anywhere, you could get it by walking calling window-frame on a list of all windows tho---probably from xlib too).

However, it's not just the frame list that's messed up:
`
STUMPWM> (current-screen)
#<SCREEN #<XLIB:SCREEN :0.0 6400x2160x24 TRUE-COLOR>>
STUMPWM> (screen-width (current-screen))
2560 (12 bits, #xA00)
STUMPWM> (screen-height (current-screen))
1440 (11 bits, #x5A0)
STUMPWM> (describe (screen-number (current-screen)))
#<XLIB:SCREEN :0.0 6400x2160x24 TRUE-COLOR>
  [structure-object]

Slots with :INSTANCE allocation:
  ROOT                           = #<XLIB:WINDOW :0 7C8>
  WIDTH                          = 6400
  HEIGHT                         = 2160
  WIDTH-IN-MILLIMETERS           = 1693
  HEIGHT-IN-MILLIMETERS          = 571
  DEPTHS                         = ((24 #<XLIB:VISUAL-INFO 8-bit TRUE-COLOR #1=:0 33>..
  ROOT-DEPTH                     = 24
  ROOT-VISUAL-INFO               = #<XLIB:VISUAL-INFO 8-bit TRUE-COLOR :0 33>
  DEFAULT-COLORMAP               = #<XLIB:COLORMAP TRUE-COLOR :0 32>
  WHITE-PIXEL                    = 16777215
  BLACK-PIXEL                    = 0
  MIN-INSTALLED-MAPS             = 1
  MAX-INSTALLED-MAPS             = 1
  BACKING-STORES                 = :WHEN-MAPPED
  SAVE-UNDERS-P                  = NIL
  EVENT-MASK-AT-OPEN             = 0
  PLIST                          = NIL
`
While the screen object knows what size it should be, the actual underlying xlib object is much too large, in fact, it's the exact size it should be if both monitors were on (xrandr confirms that only one is enabled (or even connected)).

This and a review of the code leads me to, unfortunately, believe that this isn't caused by inverting the order of adding and removing screens, but rather is a race in handling `:CONFIGURE-NOTIFY` events.

tiberious726:
Does Stumpwm ever tell xlib about the display size? If not, at least the second part of the previous comment is a cl-xlib bug:

```
STUMPWM> (xlib:display-roots *display*)
(#<XLIB:SCREEN :0.0 6400x2160x24 TRUE-COLOR>)
```

```
$ xrandr
Screen 0: minimum 320 x 200, current 2560 x 1440, maximum 16384 x 16384
eDP-1 connected primary 2560x1440+0+0 (normal left inverted right x axis y axis) 309mm x 174mm
```

```
$ xwininfo -root

xwininfo: Window id: 0x7c8 (the root window) (has no name)

  Absolute upper-left X:  0
  Absolute upper-left Y:  0
  Relative upper-left X:  0
  Relative upper-left Y:  0
  Width: 2560
  Height: 1440
  Depth: 24
  Visual: 0x21
  Visual Class: TrueColor
  Border width: 0
  Class: InputOutput
  Colormap: 0x20 (installed)
  Bit Gravity State: ForgetGravity
  Window Gravity State: NorthWestGravity
  Backing Store State: NotUseful
  Save Under State: no
  Map State: IsViewable
  Override Redirect State: no
  Corners:  +0+0  -0+0  -0-0  +0-0
  -geometry 2560x1440+0+0
```

The frame issue can be (hackily) resolved by:
```
(let ((a-correct-frame  (window-frame (current-window))))
  (mapcar #'(lambda (w) (setf (window-frame w) a-correct-frame)) (all-windows)))
```
and then getting all windows to refresh/redisplay.

I'm not sure how significant the xlib object being too large is, but it looks like things should function correctly if we stop that single window from continuing to be assigned to a frame that shouldn't exist anymore.

PuercoPop:
> Does Stumpwm ever tell xlib about the display size?
Do, we just query it and use xrandr to keep it synced

PuercoPop:
> This pull request solves the problem by simply preventing there from not being any heads during normal operation: adding new heads before removing the old ones if both happen in a single :CONFIGURE-NOTIFY event.

> I do, however, think that stumpwm ought be able to handle not having any actual heads attached, at least temporarily, and this patch is really papering over the issues with that (and, in fact, prevents tile-window's frame slots from being set to nil if you have safety turned up on sbcl).

I agree that ultimately having no monitors should be supported, at least as a transitory state but your workaround should cover most scenarios. Thanks!

tiberious726:
Did a bit more digging, it's actually the xlib "screen" size that is incorrect: it's just bigger than the display size.

```
STUMPWM> (setf (xlib:screen-height (screen-number (current-screen))) 1440)
1440 (11 bits, #x5A0)
STUMPWM> (setf (xlib:screen-width (screen-number (current-screen))) 2560)
2560 (12 bits, #xA00)
```
Fixes the xlib screen size and 
```
STUMPWM> (setf (xlib:screen-width-in-millimeters (screen-number (current-screen))) 677)
STUMPWM> (setf (xlib:screen-height-in-millimeters (screen-number (current-screen))) 318)
```
Interestingly, these are both about 2.19x larger than the values reported by xrandr, must be something being smart along the way.

So whatever should be resizing the screen on head-changes isn't working---tho so long as no windows were assigned to orphaned frames, I'm not sure anyone would have noticed their screen being larger than their actual display---since you were still constrained to a frame within your head, splitting and such would work correctly.

I'll make either a new issue or new pull request to keep track of this, depending on how quickly I figure it out.

tiberious726:
This should also resolve #840 and #841

Issue #994, opened 2022-07-17
Build fails on latest SBCL 2.2.6.118-bb245ef94


reltr:
As of [7882c92e6a5d6da86c2acd1242806aa05f0479f8](https://github.com/sbcl/sbcl/commit/856a6ac26f2e7ac74d5f26cd4390bc1bd7e63c4e) SBCL removes the symbol `sb-unix:unix-select` which results in a build failure on stumpwm/ioloop.lisp.
Given the verbiage in the commit, this is not an mistake, and won't be reversed.
This feature is replaced with `sb-unix:unix-simple-poll`, but I do not know how to alter the calls in ioloop.lisp to adapt the code to the new interface.

To reproduce:
1. Build latest SBCL from Git
2. Attempt to build StumpWM from source

Thank you,

PuercoPop:
I have to double check, but iirc sb-unix is considered an 'internal' package and sb-posix is the one we should be using

PuercoPop:
Had some time to look into, there is sb-unix:unix-fast-select which receives the same arguments but returns different values

cnoceda:
Same problem here whit SBCL 2.2.7

yusi1:
Same problem for me, can't build StumpWM anymore, it just fails with that.

yusi1:
```
; caught ERROR:
;   READ error during COMPILE-FILE:
;   
;     Symbol "UNIX-SELECT" not found in the SB-UNIX package.
;   
;       Line: 273, Column: 43, File-Position: 13683
;   
;       Stream: #<SB-INT:FORM-TRACKING-STREAM for "file /home/yaslam/.gitcloned/stumpwm/ioloop.lisp" {1003E8F223}>
; 
; caught ERROR:
;   READ error during COMPILE-FILE:
;   
;     Symbol "UNIX-SELECT" not found in the SB-UNIX package.
;   
;       Line: 273, Column: 43, File-Position: 13683
```

jl2:
I looked into this a little bit earlier this week, and the fix isn't very straightforward.

SBCL removed unix-select, but AFAICT the function they replaced it with in their code, sb-unix:unix-simple-poll, won't work for the way StumpWM is using it because StumpWM is waiting on mulitple file descriptors.

sb-unix:unix-fast-select should work, but the arguments and return values are encoded differently (directly as foreign (struct fd-set)) and when I tried it, StumpWM froze.  I'll have more time to tinker with it this weekend.

[...]

PR #1011, opened 2022-08-06
Fix 994: Replace sb-unix:unix-select with sb-unix:unix-fast-select

jl2:
This fixes [Issue 994](https://github.com/stumpwm/stumpwm/issues/994), caused by SBCL [removing `#'sb-unix:unix-select`](https://github.com/sbcl/sbcl/commit/856a6ac26f2e7ac74d5f26cd4390bc1bd7e63c4e)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -76,3 +76,4 @@ Andrin Bertschi     hi at abertschi dot ch
 Spenser Max Truex   web ate spensertruex.com
 Toby Worland        toby dot worland at gmail.com
 Tomasz Jeneralczyk  silicius at schwi dot pl
+Jeremiah LaRocco    jeremiah_larocco at fastmail com
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -200,91 +200,110 @@
                                            (format nil " (~A)" description)
                                            ""))
       (block io-loop
-        (loop
-           (with-simple-restart (:restart-ioloop "Restart at I/O loop~A"
-                                                 (if description
-                                                     (format nil " (~A)" description)
-                                                     ""))
-             (macrolet ((with-channel-restarts ((channel &optional remove-code) &body body)
-                          (let ((ch (gensym "CHANNEL")))
-                            `(let* ((,ch ,channel)
-                                    (*current-io-channel* ,ch))
-                               (restart-case
-                                   (progn ,@body)
-                                 (:skip-channel ()
-                                  :report (lambda (s)
-                                            (format s "Continue as if without channel ~S" ,ch))
-                                   nil)
-                                 (:remove-channel ()
-                                  :report (lambda (s)
-                                            (format s "Unregister channel ~S and continue" ,ch))
-                                   ,(or remove-code `(io-loop-remove info ,ch))
-                                   nil))))))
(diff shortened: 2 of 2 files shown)


jl2:
I checked [another PR build](https://github.com/stumpwm/stumpwm/runs/7702949228?check_suite_focus=true#step:3:5299), and it also gets this error, but since it's running interactively, the debugger comes up, and somehow the process exits normally and the error ends up ignored.

The problem appears to be line 81 of manual.lisp with UTF-8 characters:

`(ppcre:register-groups-bind (name) ("^€€€ (.*)" line)`

I don't get this error locally with SBCL 2.2.7, so it may be fixed by switching to a newer SBCL.

I think the best solution is to configure SBCL to use UTF-8 during the build, but I'm not sure how to configure that.  Another option is to replace "€€€" with something else, but that seems lame.

Any suggestions on how to move forward?  I'll have some time this evening to work on it.

sandric:
Just wanted to leave mb helpful note that I had the same issue with sbcl error during make, and after running make on your pr branch - stumpwm got installed without any errors :+1:

jl2:
I reverted the Makefile changes in this PR and [opened a new one ](https://github.com/stumpwm/stumpwm/pull/1018) that adds "--non-interactive" and sets SBCL's default encoding to UTF-8.

PuercoPop:
Thanks for taking the time to tackle this :pray: :tada:

cnoceda:
Thank you very much!!!!

tiberious726:
With this PR merged, some time after starting stumpwm, I get an error from the sbcl-specialized ioloop method:

```
debugger invoked on a TYPE-ERROR @52AE264D in thread
#<THREAD "main thread" RUNNING {10025E0293}>:
  The value
    NIL
  is not of type
    NUMBER
```

I'll post more details if I uncover them

shassard:
I'm experiencing the same crash over here, but hadn't had a chance to bisect it against this specific commit.

Here's the backtrace via `(copy-unhandled-error)`:

```
The value
  NIL
is not of type
  NUMBER
Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {1005698073}>
0: ((:METHOD IO-LOOP (STUMPWM::SBCL-IO-LOOP)) #<STUMPWM::SBCL-IO-LOOP {1003E95033}> :DESCRIPTION "StumpWM") [fast-method]
1: (STUMPWM::STUMPWM-INTERNAL-LOOP)
2: (STUMPWM::STUMPWM-INTERNAL ":0")
3: (STUMPWM ":0")
4: (COMMON-LISP-USER::MAIN)
5: (ROSWELL::ENTRY "COMMON-LISP-USER::MAIN")
6: (ROSWELL:RUN ((:ENTRY "COMMON-LISP-USER::MAIN")))
7: ((FLET SB-UNIX::BODY :IN SB-IMPL::START-LISP))
8: ((FLET "WITHOUT-INTERRUPTS-BODY-3" :IN SB-IMPL::START-LISP))
9: (SB-IMPL::%START-LISP)
```

tiberious726:
I've been running on f70c93ba (in addition to my pull-requests, and 6761f1f1 ) for a few hours now, no crashes, I could reliably get anything running on top of 9f476737 to crash in a few minuets.

tiberious726:
#1020

PR #1017, opened 2022-08-06
Grab buttons on managed window instead of parent

### LINKED ISSUE #527: Clicking on a :hover element in browser hides it
So, I know stumpwm is not a browser, but this behavior is only seen when using stumpwm.

You can see it here: http://jsfiddle.net/396hLkgv/embedded/result/

Hover over "foo", and try clicking on "bar". It'll disappear, without any reason.

It looks like the mouse click sends... _something_ to Chrome, making Chrome think the mouse is no longer there. Almost as if there's a `banish`, or something like that (the cursor isn't moving, to be clear).

Any idea of where I can look at fixing this?

### LINKED ISSUE #894: Odd interaction with some drop-down menu elements in Chromium-based web browsers
(For the record #527 looks really similar to this issue, but I'm not 100% comfortable ascribing this problem to the same GH issue...)

---

Steps to reproduce:
1. Run a Chromium-based web browser in stumpwm (Chromium, Chrome, Vivaldi)
2. Visit any of the following sites: www.aliexpress.com www.teksavvy.com www.trackitt.com (I'm sure there are 1000s more)
3. Attempt to interact with the drop-down navigational menus on the websites
4. Observe that clicking menu items results in the menu disappearing without following the link.
  The menu disappears seemingly before the click goes through -- often meaning something _behind_ the link gets clicked instead.

---

Stumpwm version: `master`, SBCL `"2.1.1"`, `xlib:*version* ;=> "MIT R5.02"`

I tested this issue across:
* Multiple Machines
* Input Methods (Mouse, TrackPoint)
* Linux distros (Arch, Ubuntu)
* Browsers (Vivaldi, Chrome), incognito modes included
* Environments (Gnome, i3, and even CLFSWM to see if it was a CLX issue -- it wasn't)

The only environment triggering this behaviour is stumpwm (with no config).

It's not a regression either -- I recompiled stumpwm from releases in 2014 and 2017 and it had exactly the same issue.

---

_Possibly unrelated, but these menus also disappear when I hover them, and then press the prefix key `C-t`..._

jorams:
This fixes an issue that is quite possibly a bug in Chromium. Given the following steps:

1. Run a Chromium-based web browser
2. Visit a site with a menu that opens on hover
3. Hover over the menu to open it
4. Click an item in the menu

The expectation is to be clicking the menu item. Instead the menu would close and the click would pass through to the element underneath.

This issue could be worked around by passing the following option to Chromium:

    --disable-features=SendMouseLeaveEvents

In the source code, this feature is explained as follows:

    This feature allows native ET_MOUSE_EXIT events to be passed through
    to blink as mouse leave events. Traditionally these events were
    converted to mouse move events due to a number of inconsistencies on
    the native platforms. crbug.com/450631

With the buttons grabbed on the parent window, Chromium would receive the following sequence of events (as reported by xev):

    LeaveNotify event, serial 36, synthetic NO, window 0x3600001,
      root 0x1dd, subw 0x0, time 1129083356, (10,677), root:(970,693),
      mode NotifyGrab, detail NotifyAncestor, same_screen YES,
      focus YES, state 256

    EnterNotify event, serial 36, synthetic NO, window 0x3600001,
      root 0x1dd, subw 0x0, time 1129083358, (10,677), root:(970,693),
      mode NotifyUngrab, detail NotifyAncestor, same_screen YES,
      focus YES, state 256

    KeymapNotify event, serial 36, synthetic NO, window 0x0,
      keys:  221 0   0   0   0   0   0   0   0   0   0   0   0   0   0   0
             0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0

    ButtonPress event, serial 36, synthetic NO, window 0x3600001,
      root 0x1dd, subw 0x0, time 1129083356, (10,677), root:(970,693),
      state 0x0, button 1, same_screen YES

    ButtonRelease event, serial 36, synthetic NO, window 0x3600001,
      root 0x1dd, subw 0x0, time 1129083460, (10,677), root:(970,693),
      state 0x100, button 1, same_screen YES

One of the events before ButtonPress triggers the problem. Considering the workaround this is almost certainly LeaveNotify.

These LeaveNotify and EnterNotify events are "pseudo-motion" events due to the activation and de-activation of StumpWM's grab on the parent window.

After these changes, the sequence of events becomes much cleaner:

    ButtonPress event, serial 36, synthetic NO, window 0x3600001,
      root 0x1dd, subw 0x0, time 1129083356, (10,677), root:(970,693),
      state 0x0, button 1, same_screen YES

    ButtonRelease event, serial 36, synthetic NO, window 0x3600001,
      root 0x1dd, subw 0x0, time 1129083460, (10,677), root:(970,693),
      state 0x100, button 1, same_screen YES

This brings StumpWM in line with other tested window managers.

Fixes #527
Fixes #894

### DIFF
--- a/events.lisp
+++ b/events.lisp
@@ -631,8 +631,19 @@ the window in it's frame."
   (let ((button (decode-button-code code))
         (screen (find-screen window))
         (mode-line (find-mode-line-by-window window))
-        (win (find-window-by-parent window (top-windows))))
+        (win))
     (run-hook-with-args *click-hook* screen code x y)
+    (cond
+      ((setf win (find-window window))
+       ;; We received the event at the parent window, but it occurred in
+       ;; the window itself. We need to adjust the coordinates to make them
+       ;; relative to the parent window.
+       (setf x (+ x (xlib:drawable-x window))
+             y (+ y (xlib:drawable-y window))))
+      ((setf win (find-window-by-parent window (top-windows)))
+       ;; The click happened on the parent window, so the coordinates are
+       ;; correct.
+       ))
     (cond
       ((and screen (not child))
        (group-button-press (screen-current-group screen) button x y :root)
--- a/window.lisp
+++ b/window.lisp
@@ -672,7 +672,7 @@ and bottom_end_x."
     (unless (eq (xlib:window-map-state (window-xwin window)) :unmapped)
       (incf (window-unmap-ignores window)))
     (xlib:reparent-window (window-xwin window) master-window 0 0)
-    (xwin-grab-buttons master-window)
+    (xwin-grab-buttons xwin)
     ;;     ;; we need to update these values since they get set to 0,0 on reparent
     ;;     (setf (window-x window) 0
     ;;          (window-y window) 0)


tiberious726:
This resolves the issue for me.

Thank you!

PR #1020, opened 2022-08-08
Handle nil values returned by sb-unix:unix-fast-select

ak-coram:
On my system (FreeBSD 13.1-RELEASE) #1011 caused errors when `sb-unix:unix-fast-select` returned `nil` (I'm not sure why this would happen in the first place). This change ignores these values and resolves the issue for me, please let me know if there's a better solution.

Thanks!

### DIFF
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -289,20 +289,19 @@
                                     (floor remaining-ms 1000000))
                                   (values nil nil))))
                        ;; Actually block for events
-                       (let ((rval (multiple-value-call
-                                       #'sb-unix:unix-fast-select (1+ maxfd)
-                                     (sb-alien:addr rfds)
-                                     (sb-alien:addr wfds)
-                                     (sb-alien:addr efds)
-                                     (compute-timeout))))
-                         (cond ((= rval -1)
-                                ;; ???
-                                (let ((errno rval))
-                                  (cond ((eql errno sb-unix:eintr)
-                                         nil)
-                                        (t (error "Unexpected ~S error: ~A"
-                                                  'sb-unix:unix-fast-select
-                                                  (sb-int:strerror errno))))))
+                       (multiple-value-bind (rval errno)
+                           (sb-unix:unix-fast-select (1+ maxfd)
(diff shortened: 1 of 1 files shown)


PuercoPop:
https://github.com/sbcl/sbcl/blob/master/src/code/unix.lisp#L86 You are right, the syscall will return nil and the the errno in the second value  when it is not successful

PuercoPop:
I think the fix would be to check the errno returned by the syscall but I still have to test it. @jl2 any thoughts?
```patch
modified   ioloop.lisp
@@ -289,13 +289,13 @@ (defmethod io-loop ((info sbcl-io-loop) &key description)
                                     (floor remaining-ms 1000000))
                                   (values nil nil))))
                        ;; Actually block for events
-                       (let ((rval (multiple-value-call
-                                       #'sb-unix:unix-fast-select (1+ maxfd)
-                                     (sb-alien:addr rfds)
-                                     (sb-alien:addr wfds)
-                                     (sb-alien:addr efds)
-                                     (compute-timeout))))
-                         (cond ((= rval -1)
+                       (multiple-value-bind (rval errno)
+                           (sb-unix:unix-fast-select (1+ maxfd)
+                                                     (sb-alien:addr rfds)
+                                                     (sb-alien:addr wfds)
+                                                     (sb-alien:addr efds)
+                                                     (compute-timeout))
+                         (cond ((or errno (= rval -1))
                                 ;; ???
                                 (let ((errno rval))
                                   (cond ((eql errno sb-unix:eintr)
```

jl2:
Yes, that looks good to me.

ak-coram:
@PuercoPop, @jl2: Thanks! I'll test this and will update the PR.

ak-coram:
Rebased this on current master

Phundrak:
I tested your branch on my Linux machine, it indeed fixes the issue.

ak-coram:
@Phundrak: thanks! @PuercoPop, @jl2 I've updated the PR to push these errors to the debug log instead (`*debug-level* >= 5`)

PuercoPop:
@ak-coram Thanks for taking the initiative

PR #1025, opened 2022-08-17
Fix the ioloop timeout

tiberious726:
1451b05 fixed the issues with unix-fast-select returning nil, however,
it replace a call to `multiple-value-call` to one with
`multiple-value-bind` to get the the errno. However, (compute-timeouts)
returns multiple values: seconds and milliseconds remanining in the
timeout. 1451b05 passed the seconds value, and set the milliseconds to
nil, resulting in waiting for 0 seconds and 0 milliseconds whenever
the timeout was supposed to be less than a second

I also made the syscall-level errors from this debug level 1 instead
of 5

### DIFF
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -290,15 +290,16 @@
                                   (values nil nil))))
                        ;; Actually block for events
                        (multiple-value-bind (rval errno)
-                           (sb-unix:unix-fast-select (1+ maxfd)
-                                                     (sb-alien:addr rfds)
-                                                     (sb-alien:addr wfds)
-                                                     (sb-alien:addr efds)
-                                                     (compute-timeout)
-                                                     nil)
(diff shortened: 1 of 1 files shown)


PuercoPop:
> t replace a call to multiple-value-call to one with
multiple-value-bind to get the the errno. However, (compute-timeouts)
returns multiple values: seconds and milliseconds remanining in the
timeout. 

So that is why we were using m-v-c! Thanks for the clear explanation of the issue. I didn't understand the purpose of using m-v-c!

ak-coram:
@tiberious726: thanks, great find! I didn't notice this either (and sorry for screwing this up).

tiberious726:
No worries! I was pretty relieved when I realized was just time timeout and not something deeper about unix-fast-select being too fast.

PR #1027, opened 2022-08-17
Allow specifying bright and dark colors separately

goose121:
I have been having difficulty reading some of the messages (especially from the `eval` command) because the colors are too dark. By default, the current implementation of `update-color-map` subtracts 25% brightness to create the normal colors and adds 25% brightness to create the bright colors. With this approach, no "normal" color can ever be brighter than #BFBFBF, even though those colors are often used by default by various components of the window manager and may be too dark to read easily.

This pull request adds the ability to use color specifications of the form `(dark-color light-color)` in `*colors*`. This allows the user to pick any two colors for the dark and light colors, while maintaining compatibility with older configuration files.

Another possibility would be to update the color-adjusting code, but the current documentation says that the code is meant to emulate VGA colors, which do have rather muddy normal colors by default, and doing so would also change the look of StumpWM for all users, which may not be desirable.

I am not sure how to update the manual, but it would also need to reflect this new possibility for color specifications.

### DIFF
--- a/color.lisp
+++ b/color.lisp
@@ -77,7 +77,8 @@ then call (update-color-map).")
   (labels ((max-min (x y) (max 0 (min 1 (+ x y)))))
     (setf (xlib:color-red color) (max-min (xlib:color-red color) amt)
           (xlib:color-green color) (max-min (xlib:color-green color) amt)
-          (xlib:color-blue color) (max-min (xlib:color-blue color) amt))))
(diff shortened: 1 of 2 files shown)


goose121:
Oh oops, I called them "dark" instead of "normal" in the commit message

goose121:
Ok I added this syntax to the manual

dmb2:
Thanks! This was well documented and easy to follow and a nice feature.

PR #1030, opened 2022-08-21
Improve resizing

### LINKED ISSUE #928: Bug: iresize breaks tiling
To reproduce:

`only, hsplit, vsplit, move-focus down, vsplit, iresize, left, left, left, left, left, enter.`

This creates 4 frames, one on the right side and three on the left, then attempts to decrease the width of the middle left frame.

Expected:
* The three frames on the left decrease in width while left border is stationary.
* The frame on the right increases in width while right border is stationary.
* The new layout is still tiled correctly.

Actual:
* The top left frame increases in width while left border is stationary.
* The two other frames on the left decrease in width while right border is stationary.
* The frame on the right remains unchanged.
* The result is no longer a tiled layout, as the top left and right windows now overlap.

As an aside, I believe the test suite I presented in #893 would be able to catch this kind of bug. I'd be willing to add fuzzing of iresize to it, if there is an interest to add the suite to upstream.

### LINKED ISSUE #1015: Not all splits can be moved using iresize
To reproduce: `only`, `hsplit`, `hsplit`, `move-focus right`, `move-focus right`, `hsplit`.

Now, the middle frame border cannot be changed by any combination of resize.

### LINKED ISSUE #1016: *min-frame-width* and *min-frame-height* not always respected
They are respected when splitting frames, but not always when resizing or changing resolution.

### LINKED ISSUE #1031: stumpwm sometimes crashes when changing resolution

### LINKED ISSUE #1032: Changing and restoring the resolution results in small layout differences
Rounding errors can propagate. Perhaps rounding should be deferred until displaying a frame or window?

svenssonaxel:
NOTE: This PR depends on PR #1023.
Only a few commits are unique to this PR.

Fixes #928.
Fixes #1015.
Fixes #1016.
Fixes #1031.
Fixes #1032.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -243,8 +243,11 @@
 (defmethod group-replace-head (screen (group float-group) old-head new-head)
   (declare (ignore screen old-head new-head)))
 
-(defmethod group-resize-head ((group float-group) oh nh)
+(defmethod group-before-resize-head ((group float-group) oh nh)
   (declare (ignore oh nh)))
+
+(defmethod group-after-resize-head ((group float-group) head)
+  (declare (ignore head)))
 
 (defmethod group-sync-all-heads ((group float-group)))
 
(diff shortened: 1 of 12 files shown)


dmb2:
fix merge conflicts and I'll merge

svenssonaxel:
@dmb2 ping

PR #1038, opened 2022-09-16
make stumpwm working with *print-case* :downcase

bymoz089:
fix generating of generic functions and methods for class `head`, when `*print-case*` is configured to `:downcase`. Otherwise methods like `HEAD-X` where not defined (they where named `|HEAD-x|`, etc.)


sdilts:
Thanks! To make this truly correct, I think something needs to be done on the constant string part of this depending on the value of `(readtable-case *readtable*)` but we will cross that bridge when someone falls off it.

PR #1042, opened 2022-10-18
Fix crash when switching screens with floating groups

### LINKED ISSUE #1041: Crash When Screen Swapping With Floating Groups
If you e.g. disable your built-in laptop display when plugged in to an external monitor, StumpWM crashes if you don't re-enable the built-in before unplugging the external.  It will also crash if you've got only one monitor and you just turn that monitor off.

sukaeto:
Fixes #1041.

If the number of screens drops to zero (e.g. when turning off the monitor, or when unplugging an external screen from a laptop before re-enabling a disabled built-in display) and floating groups exist, the window manager will crash.  This is a regression introduced in 7fa64230 - a very helpful new generic function was added to preserve the layout of tiling groups when switching screens, but no default implementation was given for floating groups.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -240,6 +240,9 @@
 (defmethod group-remove-head ((group float-group) head)
   (declare (ignore head)))
 
+(defmethod group-replace-head (screen (group float-group) old-head new-head)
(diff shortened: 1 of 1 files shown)


sdilts:
Thanks!

PR #1049, opened 2022-11-15
Improve workflow

svenssonaxel:
Note: Builds on top of PR #1030, only a few commits are new. Will be in draft status until #1030 is merged.

Improve the CI workflow in a number of ways:
* Create script for running GitHub workflow locally
* Use same docker image for both tests and integration-tests
* Expand .dockerignore
* Count and assert number of compile warnings and notes
* Check integrity of downloads
* Upgrade SBCL in docker image to 2.2.10


dmb2:
I can probably get a decent code review in place over the long weekend, gut reaction:
There's a decent amount of code here that seems separate from the main cause of improving the CI workflow.  I would prefer those get separated out.

Also: I (admittedly) haven't been tracking changes as close as I should have, but when did we add docker build/dependencies?? I'm hoping those are just for integration testing with travis/github.

svenssonaxel:
> when did we add docker build/dependencies?? I'm hoping those are just for integration testing with travis/github.

Docker is used to run tests and integration tests in a (semi-)reproducible manner, and nothing else. This PR, among other things, makes sure we don't build two separate docker images for those two test suites.

PR #1050, opened 2022-11-15
Fix compiler warnings

svenssonaxel:
Note: Builds on top of PR #1049, only a few commits are new:
* Fix compiler warning about redefining orphaned-frames
* Fix all compiler warnings about redefining functions
* Fix compiler warning about redefining print-swm-object method
* Fix compiler note about unknown types in handle-top-level-condition
* Fix compiler notes about unknown type dynamic-group
* Fix compiler note about unknown type tile-window

Will be in draft status until #1049 is merged. This fixes most, but not all, compiler warnings and notes. Only two problems remain, and I would need some help with them:
* REDEFINITION-WITH-DEFMACRO warning for every use of defmacro. It might have something to do with ASDF.
* In command.lisp, defun argument-pop, TYPE-STYLE-WARNING: The binding of I is not a REAL.


szos:
I havent taken a full look through everything here, but one thing i noticed is that youve removed several eval-when forms from the minor modes file. Unless theres a good reason for it, i think these should be kept in, as the wrapped functions are called at macroexpansion time.

svenssonaxel:
> youve removed several eval-when forms from the minor modes file. Unless theres a good reason for it, i think these should be kept in, as the wrapped functions are called at macroexpansion time.

I don't understand `eval-when` well enough to argue otherwise, but I'm curious why removing these forms solved the compiler warnings about function redefinition, while causing no other warnings or other conditions to appear.

svenssonaxel:
There are four `eval-when` forms in `minor-modes.lisp`, all of them on the form `(eval-when (:compile-toplevel :load-toplevel :execute)`, and they cause 13 warnings of type `sb-kernel:redefinition-with-defun`. These warnings can be solved without creating any other warnings by:
* Removing the `eval-when` forms
* Changing them to `(eval-when (:load-toplevel :execute)`
* Changing them to `(eval-when (:compile-toplevel :execute)`
* Changing them to `(eval-when (:load-toplevel)`
* Changing them to `(eval-when (:compile-toplevel)`

Not sure which option is correct. Any ideas, @szos?

szos:
Im curious whats causing these warnings as well; i just upgraded from sbcl 2.2.9 to 2.2.10, and took it as an opportunity to test this. I pulled from master and ran make. When compiling the minor modes file, there were no warnings issued. Do you have any declaimations in your sbclrc?

Im guessing theres no warnings issued when the forms are removed because stumpwm doesnt ship any minor modes by default, meaning that these functions never get called (during the define-minor-mode macroexpansion). If stump wants to ship minor modes by default, the eval-when forms will be needed¹. This can be checked by copying the evil-mode example from the docstring to the bottom of the file and compiling stump - it fails, saying parse-minor-mode-options is defined in the file but not available at compile time. 

In general i think its good form to keep all functions called by macros in an eval-when with :compile-toplevel. 

¹ after rereading (well, perusing) §3.2 (and specifically §3.2.2/§3.2.3) of the spec it seems that the eval-whens may be removable, given the file compilation semantics of CL. Someone more familiar with the spec would have to chime in though. Assuming they can be removed, a note should be placed at the lowermost portion of the file (or rather, immediately before or after the define-minor-mode macro definition) stating that no minor modes may be implemented in this file. 

I will say, im partial to keeping the eval-when forms. They signify to any readers that the following code is intended to be available at compile time. I am curious why you get errors from these forms when i dont.

svenssonaxel:
> I am curious why you get errors from these forms when i dont.

Probably because I am explicitly establishing a condition handler that prints and  declines to handle conditions signaled during loading. Check out the parent of the commit in question (`95b3136`), run `./ci`, and you should see the same. I did this because I wanted to count and assert the conditions raised during compilation. Turns out that SBCL does not print every condition raised, and rather somehow tells interesting and uninteresting conditions apart. It's possible I'm beating a dead cat here, but if so I'd rather find a way to programmatically decide what conditions are uninteresting.

> Im curious whats causing these warnings as well;

After reading up on ASDF operations [1] and the spec sections you mentioned, I believe what happens is the following:
* `make-image.lisp` loads `load-stumpwm.lisp` which tells ASDF to load the system using `asdf:load-op`.
* The load operation in ASDF has a dependency on the compile operation, so every file in the system will be first compiled and then loaded.
* During compilation, macro definitions and the functions enclosed by the `eval-when` forms will be evaluated, and as a side effect of compilation, the process doing the compilation will have these macros and functions defined.
* Loading is then done in the same process. Loading also has a side effect of defining the same functions and macros, so will signal the conditions `sb-kernel:redefinition-with-defmacro` and `sb-kernel:redefinition-with-defun`, respectively.

If this is correct, then it should be possible to avoid these warnings either by compiling and loading in different, subsequent sbcl processes, or by using `asdf:load-source-op` which will load the system from source, without creating `*.fasl` files. The former seems convoluted/undesirable and the latter fails with different errors for reasons I do not immediately understand.

If it is correct, it would also explain why removing either `:load-toplevel` or `:compile-toplevel` solves the warnings, but I agree with your reasons for wanting to keep them.

[1] https://asdf.common-lisp.dev/asdf/Predefined-operations-of-ASDF.html

szos:
> Probably because I am explicitly establishing a condition handler that prints and declines to handle conditions signaled during loading.

Ah so youre grabbing every condition signalled that isnt handled lower down the stack. 

> [...] I'd rather find a way to [...] decide what conditions are uninteresting

I think that will be fairly subjective; i find the compiler notes about inability to open code uninteresting. It also seems that the compiler is using signals to display info to the user, as notes arent (to my knowledge) errors or warnings. 

> It's possible I'm beating a dead cat here [...]

It may be; even if you could handle or fix all signals in Stump, what happens if sbcl changes the conditions signalled by the compiler at some point? What are you going to do about xlib (it throws up many many notes about performance), or about declarations that are correct and useful to the programmer but that the compiler cant contend with (such as declaring a value that cannot be stack allocated as having dynamic extent)? 

> After reading up on ASDF operations [1] and the spec sections you mentioned, I believe what happens is the following [...]

Thanks for doing some digging! From this i would think that the forms being evaluated at both compile and load time is the source of these signals. From reading §3.2.1 it seems we may be able to remove the load time flags to eval-when. This section has somewhat confusing (to me) uses of the term evaluation environment. However it is stated that it is unspecified whether a definition available in the compilation environment can be used in an evaluation initiated in the evaluation environment. This would imply that we at minimum should specify both compile and execute flags in the eval-when forms to stay compliant with the spec, and its especially important if asdf compiles all files before loading them.

All that being said, i dont know if someone will be impacted by removing the load time flags. I dont think so, but im unsure of all the different ways people build stump.

svenssonaxel:
> I think that will be fairly subjective;

I was referring to emulating SBCL's behavior, but maybe that too is subjective.

> what happens if sbcl changes the conditions signalled by the compiler at some point?

Then the version of SBCL used for the CI workflow should be changed in the same commit as the number of expected signals. The number of warnings are only asserted when running the CI workflow.

> What are you going to do about xlib (it throws up many many notes about performance)

Nothing; for some reason those notes aren't caught as it is now.

> declarations that are correct and useful to the programmer but that the compiler cant contend with

Manually decide that they shouldn't be fixed and leave them among the number of expected conditions.

>  i dont know if someone will be impacted by removing the load time flags.

Perhaps it's best to simply revert this commit. We know the cause of the signals and that they are not actually a problem. The fact that the problem comes from using ASDF in a fairly ordinary way, and that SBCL doesn't even print these signals, indicate that one might do best not caring too much about them.

dmb2:
I think we need to be careful about blindly making the compiler stop printing warnings... In general I've tried to keep it clean, but if there are good reasons why a warning is showing up (ie there was a design decision that led to them) we should tolerate/suppress the warning rather than damaging the design decision.

szos:
Ive done some more reading and according to a post by fare [0] we should keep compile load and execute in the eval-when forms. It seems eval-when is rather complicated when considering what is allowed by the spec.

0 https://fare.livejournal.com/146698.html?

adlai:
Compiler notes, warnings, and various other barf are pretty much universally there for good reason. Silencing warnings might be reasonable when there's nothing else urgent on the plate, although I agree that it should never be any sort of stability metric for the deliverable project.

dmb2:
@svenssonaxel If you can revert the eval-when changes, the rest of these look pretty reasonable and I'm happy to merge.

Issue #997, opened 2022-07-23
Incorrect key name - ' is reported as >


tobyWorland:
Using no config and the latest version currently (aaf1569056b80d8d4a095f4b26a556cc784ed037): `C-t C-h k '` stump responds with the wrong key name, `> is not bound.`. Looks like the bug was introduced with commit f335b8c618b5eb1e198525a9179386bc0d654204 due to the duplicate keysym names. Before this commit stump would report `C-quoteright`.

What is with the `quoteright` and `quoteleft` names? They both have comments showing them as deprecated. Is it intended that they are used over `grave` and `apostrophe`?

svenssonaxel:
Fixed by `abb93fe`, merged in #1231. This issue should be closed.

Issue #972, opened 2022-05-11
clx call `xlib:get-wm-class` causes stumpwm to crash


mariari:
While trying to run [a game](https://en.wikipedia.org/wiki/Embodiment_of_Scarlet_Devil) Stumpwm crashes upon the game launching. I was able to get a picture of the stack trace.

[Photo of the stack trace](https://pomf2.lain.la/f/7g03szke.jpg)

It seems that stumpwm calls `xlib:get-wm-class` from `clx` which can't handle non latin characters. 

```lisp
(xlib:char->card8 #\latin_small_letter_AE)
The value
  #\LATIN_SMALL_LETTER_AE
is not of type
  BASE-CHAR
when binding CHAR
   [Condition of type TYPE-ERROR]
```

which stumpwm passes it from the game's information.

This issue seems related to https://github.com/sharplispers/clx/issues/182 who also seems to have run into a similar issue for his own stumpwm code.

It might be good to have a check on `stumpwm::xwin-class` or `stumpwm::xwin-to-window` to prevent stumpwm from crashing upon this.

dmb2:
Sorry we ruined your game session! This is a relatively well known issue in clx. Maybe we can work around it, but really its upstream of us.

PR #1051, opened 2022-11-15
Reproduce #997 and #972

svenssonaxel:
This PR adds test cases for reproducing issues #972 and #997. Both test cases are marked as expected to fail, awaiting a fix.


svenssonaxel:
Will be included in another PR.

PR #1053, opened 2022-11-25
patch dynamic-group-resize-head to not map over window

szos:
change one line to not attempt to map over a window. fixes issues with retiling heads in a dynamic group.


szos:
Apologies, this was premature. The fix is to map over windows not only call on one window. I'm closing this PR and will opening another with the appropriate fix.

Issue #1058, opened 2022-12-06
refresh-heads doesn't work with float group


sanel:
When there is a float group, running `C-t ; refresh-heads` (or `C-t : (refresh-heads)` will throw `There is no applicable method for the generic function .... TILE-GROUP-FRAME-TREE`).

The only way I could find to properly switch monitors is to kill float group and then call `refresh-heads`.

PR #1060, opened 2022-12-07
patch the orphaned frames function to ignore float groups

szos:
This addresses issue #1058. The problem is in the function `orphaned-frames`, which assumes all groups in a screen are tiling groups. This PR removes that assumption and ignores floating groups. 

This PR ignores all floating windows in the function `screen-frames`. It also makes `group-adopt-orphaned-windows` ignore float windows, as it appears that floating windows can never be orphaned; orphan status seems to refer to frames that do not belong to a group and the windows which belong to those frames.

### DIFF
--- a/floating-group.lisp
+++ b/floating-group.lisp
@@ -257,8 +257,9 @@
 (defmethod group-adopt-orphaned-windows ((group float-group) &optional (screen (current-screen)))
   (let ((orphaned-frames (orphaned-frames screen)))
     (loop for window in (list-windows screen)
-          when (member (window-frame window) orphaned-frames)
-          do (group-add-window group window))))
+          when (and (not (float-window-p window))
+                    (member (window-frame window) orphaned-frames))
+            do (group-add-window group window))))
 
 (defvar *last-click-time* 0
   "Time since the last click occurred")
--- a/head.lisp
+++ b/head.lisp
@@ -228,5 +228,6 @@
   "Returns a list of frames on a screen not associated with any group.
   These shouldn't exist."
   (let ((adopted-frames (loop for group in (screen-groups screen)
-                              append (group-frames group))))
+                              unless (typep group 'float-group)
+                                append (group-frames group))))
     (set-difference (screen-frames screen) adopted-frames)))
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -494,6 +494,7 @@ T (default) then also focus the frame."
   "Returns a list of frames on a screen not associated with any group.
   These shouldn't exist."
   (let ((adopted-frames (loop for group in (screen-groups screen)
+                              unless (typep group 'float-group)
                               append (group-frames group))))
     (set-difference (screen-frames screen) adopted-frames)))
 
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -490,14 +490,6 @@ T (default) then also focus the frame."
   (remove-duplicates (mapcar #'(lambda (window) (window-frame window))
                              (list-windows screen))))
 
-(defun orphaned-frames (screen)
-  "Returns a list of frames on a screen not associated with any group.
-  These shouldn't exist."
-  (let ((adopted-frames (loop for group in (screen-groups screen)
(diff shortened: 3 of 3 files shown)


sanel:
Hi @szos ,

Thank you for your work! I've tested it with empty float group, it works. However, if there is a window present in it (in my case I just started gvim), it will throw:

```
There is not applicable method for the generic function
   #<STANDARD-GENERIC-FUNCTION STUMPWM::WINDOW-FRAME (1)>
when called with arguments
   (#<FLOAT-WINDOW "[No Name] - GVIM" #x1800007>).
```

I've applied https://github.com/stumpwm/stumpwm/pull/1060.patch on clean stumpwm git tree.

sanel:
Also to add, floating window doesn't have to be part of a floating group. If floating window existst it any of other groups, stumpwm will throw above error as well.

szos:
Hmm, are you able to get a backtrace? Im fairly busy today but will try to get to this later in the day if i have time.

sanel:
Sure, I'm attaching a stacktrace. Also, do you know any better way to capture stacktraces, other than taking screenshots? Does stumpwm logging capability works, as my `stumpwm.log` is empty?

Also, maybe unrelated, how to attach to stumpwm repl? I could not find anything in documentation.

stumpwm

szos:
Thanks for the backtrace! Regarding getting backtraces, thats kinda the best way i know of aside from evaluating something in sly/slime, but some things dont signal errors or freeze when called in sly/slime. Im currently working on a minor mode for debugging stump which invokes the clim debugger on any unhandled serious conditions. Ive got the debugger working but interactive restarts currently fail if they read any values cause the clim debugger is separate from the clim listener and doesnt dynamically bind `*query-io*`. 

To attach to a stumpwm repl you will want to load slynk or swank into the stumpwm image (i just use quicklisp for that) and then open a server. Then you can connect using emacs (or vim, atom, vscode, whatever has a slime-esq package for it will work). You could also spin off a thread that launches the clim listener, though it uses clx and thus has the same problems with alt graph and whatnot that stump has.

szos:
This should be fixed now. There were a few additional places where it was assumed that all windows are tile windows.

szos:
@sanel you were asking about getting backtraces. Over the past 2 days I've hacked together a minor mode I'm calling SDB for StumpWM DeBugger. It wraps the CLIM debugger and makes it play nice with StumpWM. One of the functionalities I've added is the ability to print the backtrace to a file. [Here is the link to the repo, with install instructions in the readme.](https://github.com/szos/SDB)

sanel:
Thank you! Just applied fresh 1060.patch and `refresh-heads` is working with floating windows in non-floating groups. However, if there are floating windows in floating group, it will show down below stacktrace.

> Over the past few days I've hacked together a minor mode I'm calling SDB for StumpWM DeBugger.

Great, will check it! I'm curious, is there a way to capture all exceptions and just log them in a file/stdout? Something like that in Clojure:

```clojure
;;; wm code
(try
  (main-event-loop)
  (catch Exception e
    (log/error e "Something happened, stacktrace is printed")
    (continue-main-loop)))
```

stumpwm

szos:
I fixed up the other method for group-adopt-orphaned-windows to also not assume all windows are tile windows. 

Regarding logging all exceptions, you can setup the debugger to log the condition its invoked on and then exit, and then set `*break-on-signals*` to be true. This will mean you wont be able to use the debugger tho. A (pseudocode) example:

```
(defun mylog (c e)
  (Format *my-file-stream* "encountered ~a, backtrace:~%~a" c (backtrace-string)))

(Setf *debugger-hook* #'mylog
      *break-on-signals* t)
```

dmb2:
if this is ready I'll merge.

sanel:
@dmb2 let me test if over the weekend and I'll get back to you

sanel:
Tested, looks good :) Tested with floating group, floating window only and dynamic group. Thank you guys!

szos:
@sanel thank *you* for being a tester for this!

PR #1075, opened 2023-01-26
Fix %d and %e time formats

### LINKED ISSUE #1074: %d and %e time formats exchanged
`%d` and `%e` are switched, compared to what they typically refer to.

svenssonaxel:
Fixes #1074

### DIFF
--- a/time.lisp
+++ b/time.lisp
@@ -64,9 +64,9 @@
     (#\B time-month-name)
     (#\c time-date-and-time)
     (#\C time-century)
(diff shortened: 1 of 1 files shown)


szos:
This PR should probably change the default time formatters as well, no?

svenssonaxel:
> This PR should probably change the default time format strings as well, no?

The documentation for `*time-format-string-default* ` indicates an intention to use space padded by default. I believe that will be the default behavior after this is merged, but I haven't tested it.

PR #1137, opened 2023-09-06
fix modeline bar

Junker:


PuercoPop:
Thanks for taking the time to submit the PR.

Can you add more details to the commit? What problem does it fix? How to reproduce the issue it is fixing. If possible can you attach before and after screenshots?

Junker:
```lisp
(stumpwm:bar 70 5 #\X #\=)
```
this should draw bar with length 5 and 70% of it should be "X" symbols.

**Before:** `"^B^======="`
**After:** `"^B^3*XXX=="`

As you can see,  color definition "^B^3*" was broken before and the bar length is wrong (must be 5)..

PuercoPop:
Thanks for explaining. Updated the commit message and added some tests. Merging

Issue #1134, opened 2023-09-03
Paste text into eval prompt


MaxGyver83:
When I open the `eval` prompt (`C-t :`), I can't paste text there. Is this a bug? Or was this never intended to work?

(I'm using the Arch Linux `stumpwm-git` package.)

dmb2:
I'm sorry this isn't obvious, but how are you pasting?

PuercoPop:
https://github.com/stumpwm/stumpwm/blob/63adb4ed3c2f8118e9a32fadf4371d938b0a8b62/input.lisp#L172-L173

You should be able to paste with `C-y`

dmb2:
Maybe we should ad C-v for paste? I mean, I get that this comes from some gnu/emacs heritage but its not very intuitive.  Bonus points if we could also support middle-click.

MaxGyver83:
@PuercoPop: Thank you! `C-Y` works.

@dmb2: I had only tried `C-v` and `C-V`.

I know `Ctrl-y` for pasting in the shell (both bash and fish). But I didn't think about trying it because (in the shell) `Ctrl-y` works only for text that was cut with `Ctrl-w`, not for the system clipboard. Maybe `Ctrl-y` makes sense for Emacs users. But adding `Ctrl-v` as an alternative pasting shortcut sounds reasonable to me.

PuercoPop:
> Maybe Ctrl-y makes sense for Emacs users.

Yes

>  But adding Ctrl-v as an alternative pasting shortcut sounds reasonable to me.

I don't see any harm in doing so. C-v is bound to scroll up, which I doubt is a desirable action in the text input area. Feel free to open the PR. Can we close this issue?

MaxGyver83:
I'll open a pull request!

PR #1138, opened 2023-09-06
Enable pasting from the clipboard with Ctrl-v

MaxGyver83:
For non-Emacs users it's not obvious that they can paste into the eval prompt only with `C-y` or `C-Y`. They might assume that they can paste from the clipboard with `C-v`. Add this binding as an alternative to `C-Y` (no conflicts).

Also see issue #1134.


PuercoPop:
Thanks for following up! :rocket:

PR #1147, opened 2023-10-01
Do not change focus on scroll

### LINKED ISSUE #1145: Mouse focus policy :click also changes focus on scroll
I have `(setf *mouse-focus-policy* :click)` in my config file. I would expect this to make focus switch on click only, but focus also changes when I scroll.

lrustand:
Fixes #1145

### DIFF
--- a/tile-group.lisp
+++ b/tile-group.lisp
@@ -168,10 +168,15 @@
   (declare (ignore x y))
   (when (typep where 'float-window)
     (call-next-method))
-  (when (member *mouse-focus-policy* '(:click :sloppy))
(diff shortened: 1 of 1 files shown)


sdilts:
This is great! Reading the definition of the `:sloppy` mouse focus policy, it sounds like it should change focus on scroll, but I'll let someone else chime in too. This MR modifies the behavior for both `:sloppy` and `:click`.

lrustand:
Okay, changed it so the behaviour for `:sloppy` is still the same.

sdilts:
Thanks!

Issue #1152, opened 2023-11-01
Resizing is extremely slow


rustand2:
Resizing frames is sometimes extremely slow, and it also accumulates a queue of resize events so it is very easy to end up with Stump freezing for a long time while slowly performing a lot of resizes. An then after a few minutes when it is done one frame is way too small, since there is no way to abort once you've started to resize.

The way I usually resize frames is to hold down the keybinding until I achieve the correct size. This works fine normally, but in the cases where the resize is slow for some reason I end up with a massive queue of resize events that need to be processed before Stump starts responding to anything else.

I think the slowness could be related to number of windows, since it is usually fast and responsive when first starting, but after a while this issue arises.

rustand2:
A simple way to fix it might be to ignore resize events when processing the previous event. But it is probably better to find the reason that resizing slows down in the first place.

rustand2:
One thing I see that is slowing things down is that the modeline is updated every time a resize happens, but I think this is only part of the problem.

lrustand:
> One thing I see that is slowing things down is that the modeline is updated every time a resize happens, but I think this is only part of the problem.

I figured out this part at least, it happens because `resize-direction` calls `resize` through `eval-command`, which on line 586 updates all modelines. I don't really see why `eval-command` should always cause modelines to be updated though. This should instead be done inside of the commands themselves if the command actually does something that requires the modeline to be updated.

(Yes, I'm the same guy as @rustand2, that is my alt account)

sdilts:
> because resize-direction calls resize through eval-command

This sounds like the real problem. Why is it going through `eval-command` instead of calling `resize` directly? If we don't want to directly call something defined with `defcommand`, the relevant code should be refactored out to something that we can call directly.

lrustand:
I tried rewriting `resize-direction` to call directly `resize`, but the modeline is still updated. Seems like all keybinds are actually processed through `eval-command`. I still had to disable the `(update-all-mode-lines)` line in order to not have it update for every resize event.

I still am of the opinion that there is no need to always update the modeline inside of `eval-command`, instead the specific commands themselves should update the modeline if necessary. I don't think there are actually that many commands that would need to update the modeline?

lrustand:
> If we don't want to directly call something defined with defcommand, the relevant code should be refactored out to something that we can call directly.

Should I have refactored out the code from `resize` into a defun also?

dmb2:
[...]

PR #1154, opened 2023-11-10
Refactor resize-direction

lrustand:
Rewrites `resize-direction` to call `resize` directly instead of through `eval-command`.`

See #1152 for details


dmb2:
I'm happy to merge the parts of this that call resize directly, that is clearly the more correct solution in this specific case. See the other discussion about removing mode-line-update.

lrustand:
I've removed the part that you requested and updated the PR description

Issue #1127, opened 2023-06-30
Mode line updates on every click?


MIvanchev:
Could it just be me or does the mode line update after every click anywhere for the rest of you? Right now, whenever I click somewhere in a window, an update is always issued.

ethanxxxl:
If you have `*mouse-focus-policy*` set to `:click` or `:sloppy`, and either of those events occur, [`group-buton-press`](https://github.com/stumpwm/stumpwm/blob/master/tile-group.lisp#L167) is called. This method calls two functions with side effects (for tiled windows):

1. [`(focus-all window)`](https://github.com/stumpwm/stumpwm/blob/master/events.lisp#L596) updates `*screen-list*` by moving the current screen to the front of the list, and modifies the (new) current screens `heads` and `groups` slots, moving the new current head/group to the front of their respective lists.
2. [`(update-all-modelines)`](https://github.com/stumpwm/stumpwm/blob/master/mode-line.lisp#L350) calls `redraw-mode-line` on every mode line in `*mode-lines*`

Perhaps a click in the current active window isn't distinguished from a click in any window.

MIvanchev:
OK, that would explain it, I guess this logic needs a bit of refined because updating the mode line is a costly operation.

lrustand:
I see how the intended behaviour could be helpful if one has a window list in the modeline. However if one does not, then there is no point in triggering the modeline on any sort of focus or click event. Shouldn't this be updated to only trigger if there is a window list in the modeline?

PR #1155, opened 2023-11-10
Don't update mode-line when changing focus, unless window is used in mode-line format

lrustand:
Related: #1152, #1127


dmb2:
Unfortunately this is doesn't adequately address the issue since contrib modules add mode-line format directives, some of which may need to update the modeline. I don't want to couple the codebase to contrib modules that may or may not be loaded.  Instead see my comments in the other issue.

PR #1159, opened 2023-12-02
stumpwm.asd: Bump :version to 23.11

Jehops:


dmb2:
I always forget to do this, thanks!

PR #1189, opened 2024-04-17
Ignore the hidden (dot-prefixed) executables in PATH.

aartaka:
This makes the `.`-prefixed executables hidden in `run-shell-command`/`exec`. Guix (I'm sure, among many other package managers) generates dot-prefixed executables when wrapping things into proper environments. So there are lots of `.program-real` types of executables that `exec` lists. 

This patch ignores the hidden executables by checking if they are prefixed with a dot. Makes things much clearer, at least on Guix.

### DIFF
--- a/user.lisp
+++ b/user.lisp
@@ -118,7 +118,8 @@ seperated by a colon."
            nconc (loop for file in (directory (merge-pathnames (make-pathname :name :wild :type :wild) dir)
                                               :resolve-symlinks nil)
                        for namestring = (file-namestring file)
-                       when (pathname-is-executable-p file)
+                       when (and (pathname-is-executable-p file)
+                                 (not (uiop:string-prefix-p #\. namestring)))
                          collect (if full-path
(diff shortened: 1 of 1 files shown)


sdilts:
This makes sense to me, but are there reasons against doing this? Do shells like Bash have the same behavior? I'm just trying to figure out if this would be unexpected behavior for anyone.

Hiding this behind a preference might be a good idea, but if we can prove other programs do this, there's no reason to.

aartaka:
My reasoning is that it's UNIX typical way of hiding things. But I don't
have proof beyond that, unfortunately.

dmb2:
Unless you've got some really good evidence in the desktop standards that this is desired behavior, I think its more a pain point of GUIX than stumpwm writ large. 

I think this might be better suited for your stumpwmrc than for behavior in stumpwm proper.

aartaka:
Alright, fair. Closing 😌

PR #1194, opened 2024-05-26
Avoid source registry init and add (require "asdf") where it's missing.

### LINKED ISSUE #1191: Help required: ASDF system in standard ASDF load paths (like `/usr/share/common-lib/source`) not considered unless I recompile StumpWM
Basically the title. If I add a system to the standard path like `/usr/share/common-lib/source` or `~/common-lisp` it's not discoverable from within StumpWM, i.e. `echo "(asdf:load-system \"iterate\")" | stumpish -e eval` unless I recompile StumpWM. Does anyone know how to remedy this situation?

MIvanchev:
Fixes #1191 and adds a missing `(require 'asdf)` because at least for me it wasn't working otherwise. About #1191, I found out that `initialize-source-registry` was causing issues with libraries not being found anymore until StumpWM is recompiled. Once I changed the code to avoid the source registry (re-)initialization it seems the problem has been fixed and StumpWM is running normally.


MIvanchev:
HmmmHmmm... maybe didn't quite fix #1191 from what I observe...

dmb2:
Do we need this PR anymore or not?

MIvanchev:
We don't **need** it but IMO it's a good idea to have it.

PR #1208, opened 2024-09-14
fix typo in conf-rc pathname

zmyrgel:
The config is filename so no trailing slash should be used.

### DIFF
--- a/stumpwm.lisp
+++ b/stumpwm.lisp
@@ -45,7 +45,7 @@ further up. "
          (dir-rc
            (probe-file (merge-pathnames #p".stumpwm.d/init.lisp" (user-homedir-pathname))))
          (conf-rc
-           (probe-file (uiop:xdg-config-home #p"stumpwm/config/")))
+           (probe-file (uiop:xdg-config-home #p"stumpwm/config")))
          (etc-rc (probe-file #p"/etc/stumpwmrc"))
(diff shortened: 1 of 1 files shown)


dmb2:
Config in this case should be a directory so this is not a typo. Additionally the PR has whitespace changes.

zmyrgel:
I find it odd to rely on probe-file to convert directory pathname to file pathname.

PuercoPop:
@zmyrgel imho directories are files. I am assuming you don't? Do you have an alternative approach in mind? Perhaps there is a stat/directoryp function in `sbcl-posix`?

zmyrgel:
I'm not that well versed in Common Lisp but isn't "directory == file" unix thing and not Common Lisp one. This is quite low hanging fruit in any case.

Codewise use something simple as : 
`(merge-pathnames "stumpwm/config" (uiop:xdg-config-home))`

Another thing I noticed, the StumpWM manual mentions it loads the configuration from file in XDG path ~/.config/stumpwm/config but it fails to mention that it only does that if the user has not set the XDG_CONFIG_HOME environment variable pointing to some other directory.

dmb2:
@zmyrgel If you'd like to contribute another MR addressing the documentation issue, I'm happy to review it.  For now I'm closing this without merging.

PR #1209, opened 2024-09-15
Better default data dir

zmyrgel:
Set the *data-dir* to directory where the user configuration file is instead of always using "~/.stumpwm.d".

### DIFF
--- a/debug.lisp
+++ b/debug.lisp
@@ -68,20 +68,3 @@ do:
         *standard-output* *redirect-stream*
         *trace-output*    *redirect-stream*
         *debug-stream*    *redirect-stream*))
-
-(defun rotate-log ()
-  (let ((log-filename (merge-pathnames "stumpwm.log" *data-dir*))
-        (bkp-log-filename (merge-pathnames "stumpwm.log.1" *data-dir*)))
-    (when (probe-file log-filename)
-      (rename-file log-filename bkp-log-filename))))
-
-(defun open-log ()
-  (rotate-log)
-  (let ((log-filename (merge-pathnames "stumpwm.log" *data-dir*)))
-    (setf *debug-stream* (open log-filename :direction :output
-                                            :if-exists :supersede
-                                            :if-does-not-exist :create))))
-(defun close-log ()
-  (when (boundp '*debug-stream*)
-    (close *debug-stream*)
-    (makunbound '*debug-stream*)))
--- a/debug.lisp
+++ b/debug.lisp
@@ -70,17 +70,14 @@ do:
         *debug-stream*    *redirect-stream*))
 
 (defun rotate-log ()
-  (let ((log-filename (merge-pathnames "stumpwm.log"
-                                      (data-dir)))
-        (bkp-log-filename (merge-pathnames "stumpwm.log.1"
-                                          (data-dir))))
+  (let ((log-filename (merge-pathnames "stumpwm.log" *data-dir*))
+        (bkp-log-filename (merge-pathnames "stumpwm.log.1" *data-dir*)))
     (when (probe-file log-filename)
       (rename-file log-filename bkp-log-filename))))
 
 (defun open-log ()
   (rotate-log)
-  (let ((log-filename (merge-pathnames "stumpwm.log"
-                                      (data-dir))))
+  (let ((log-filename (merge-pathnames "stumpwm.log" *data-dir*)))
     (setf *debug-stream* (open log-filename :direction :output
                                             :if-exists :supersede
                                             :if-does-not-exist :create))))
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -1539,9 +1539,22 @@ sync-all-frame-windows to see the change.")
 (defvar *data-dir* nil
   "The directory used by stumpwm to store data between sessions.")
 
+(defun ensure-data-dir ()
+  (ensure-directories-exist *data-dir* :mode #o700))
+
+(defun default-data-dir ()
+  "Return the default data dir pathname based on the loaded StumpWM configuration file."
+  (let ((rc-file (or
+                  (probe-file (merge-pathnames #p".stumpwmrc" (user-homedir-pathname)))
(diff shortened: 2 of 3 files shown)


zmyrgel:
Ah, should have properly tested this. Let's see.

MIvanchev:
Can you elaborate on this MR? I see a lot more going on than a new default dir.

zmyrgel:
Sure thing. The goal is to have choose good location for the *data-dir* instead of hard coding it to ~/.stumpwm.d. 
I have moved my stumpwm configuration to "XGD" location in ~/.config/stumpwm but noticed that stumpwm creates ~/.stumpwm.d directory with log file even though I'm not using that path myself. I'd argue that the data-dir should be set based on the user configuration directory location. The data-dir is set and created early in stumpwm.lisp so the directory is created before the user configuration is read so changing it in the configuration is too late.

The PR attempts to set the data-dir based on the user configuration file location to either ~/.config/stumpwm or ~/.stumpwm.d.

The data-dir functions etc were moved to primitives.lisp so they get loaded early. The debug.lisp was merged to primitives as well as it uses the data-dir but debug.lisp was loaded before primitives.lisp.

Now that I'm looking at the XDG stuff more closely, there is also XDG_DATA_HOME env variable which defaults to ~/.local/share/ so I'd say ~/.local/share/stumpwm would probably be possible candidate for data-dir location in case the user configuration is located at XGD_CONFIG_HOME or the XDG_DATA_HOME env variable is set.

MIvanchev:
Thanks. So I think it's better to leave debug.lisp as it was and if necessary rearrange the file list of the ASD file or use `depends-on` tags. This change makes the PR really hard to follow. About the data dir I think you're right that `XDG_CONFIG_HOME` is better suited for the the config dir, however when need to discuss it with someone else as we'll use backwards compatibility.

dmb2:
Hi,

I welcome this type of change to stumpwm however there are two things we need to merge:
1. It cannot change the current default behavior, (ie preferring XDG based variables has to be opt-in not opt-out)
2. There needs to be minimal disruption to the code to achieve the goals (ie don't move/delete/merge files)
3. No whitespace changes.

If you can address the above three points I'll merge the PR.

zmyrgel:
I've got a better version for this, I'll push it here once I figure out how to properly skip the whitespaces in it with emacs vc.

The data dir location is other thing, should stumpwm follow the XDG here and use the XDG_DATA_DIR so ~/.local/share/stumpwm?

sdilts:
> The data dir location is other thing, should stumpwm follow the XDG here and use the XDG_DATA_DIR so ~/.local/share/stumpwm?

Yes, that's correct. Here's the spec if someone else wants to reference it: https://specifications.freedesktop.org/basedir-spec/latest/index.html

To be clear, the init file is a configuration file, so it should go into the folder under `$XDG_CONFIG_HOME/stumpwm/`. Data files should go under `$XDG_DATA_HOME/stumpwm/`. So with the default locations, the configuration file should be at `~/.config/stumpwm/init.lisp`, and the data directory at `~/.local/share/stumpwm/`.

I don't know why we are expecting the configuration file at `$XDG_DATA_HOME/stumpwm/config/`. The code in [`load-rc-file`](https://github.com/stumpwm/stumpwm/blob/master/stumpwm.lisp#L37-L59) doesn't look right to me, but unless it affects how this code functions, it's outside the scope of what is trying to be done here and should be left alone.

After fixing the the path to point to the data directory, the only thing I would change would be to unify the code finding the init file between this new code and what's in the `load-rc-file` function so we only need to change it in one place if modifications are needed. That might require more effort beyond the scope of this MR though, so maybe just leaving a FIXME comment in both locations would be good enough.

sdilts:
Reading the spec again, it suggests that logs should be placed under `$XDG_STATE_HOME`, but it doesn't look like a corresponding `uiop` function exists for that particular directory.

Looking at the [relevant uiop code](https://gitlab.common-lisp.net/asdf/asdf/-/blob/master/uiop/configuration.lisp#L291-301), the implementation should be something like this (untested):

``` lisp
(defun xdg-state-home (&rest more)
  "Returns an absolute pathname for the directory containing user-specific data files
that aren't important or portable enough to be stored in the xdg-data directory.
MORE may contain specifications for a subpath relative to this directory: a
subpathname specification and keyword arguments as per RESOLVE-LOCATION \(see
also \"Configuration DSL\"\) in the ASDF manual."
  (uiop:resolve-absolute-location
   `(,(or (uiop;getenv-absolute-directory "XDG_DATA_HOME")
          (uiop:subpathname (user-homedir-pathname) ".local/state/"))
     ,more)))
```

zmyrgel:
> I don't know why we are expecting the configuration file at `$XDG_DATA_HOME/stumpwm/config/`. The code in [`load-rc-file`](https://github.com/stumpwm/stumpwm/blob/master/stumpwm.lisp#L37-L59) doesn't look right to me, but unless it affects how this code functions, it's outside the scope of what is trying to be done here and should be left alone.

The probe-file call in the load-rc-file converts the directory pathname into to file pathname so configuration is loaded from file called config without extension. I don't know why it is not using init.lisp there as well though.

But back to matter at hand, I agree the XDG_STATE_HOME would be target for data-dir.

dmb2:
I re-ran the build integration test, but it failed again.  I can't merge this until the CI test passes.

dmb2:
Can you provide the reasoning for moving the log functions from debug to primitives?

zmyrgel:
The reasoning is in the commit message and the above message. What else is needed?

dmb2:
Apologies, I should've read closer.

fstamour:
Finally 😍

anticomputer:
@zmyrgel would like to draw your attention to https://github.com/stumpwm/stumpwm/issues/1225#issuecomment-2675376252 and the fact that your merged changes in this PR break StumpWM users that manage their configuration files through symlinks, which is a reasonably popular thing to do. Examples include Guix, Nix, and users of stow.

PR #1214, opened 2024-10-08
Reset uiop/asdf after load an image

catap:
When stumpwm is built in some sort of fake environment the resulted image contains cached path to that fake enviroment which migth broke quicklisp.

Here I call a standard callback which is used by uiop for :toplevel function, and, additionally clear asdf configuration from build system and reset it.


catap:
@dmb2 thanks!

PR #1218, opened 2024-11-29
Fix building without quick-lisp

catap:
Before 24.11 it was possible to build stumpwm without quick-lisp, by adding a symlink to a few .asd into it's root.

https://github.com/stumpwm/stumpwm/pull/1194 brokes it, and here I had recovered that part which is needed to make an update for OpenBSD port at least.

### DIFF
--- a/load-stumpwm.lisp.in
+++ b/load-stumpwm.lisp.in
@@ -5,5 +5,9 @@
 
 (require 'asdf)
 
-(asdf:load-asd #p"@STUMPWM_ASDF_DIR@/stumpwm.asd")
+(asdf:initialize-source-registry
+ '(:source-registry
+   (:directory "@STUMPWM_ASDF_DIR@")
+   :inherit-configuration))
+
 (asdf:load-system "stumpwm")


MIvanchev:
@catap Keep in mind I don't use Quicklisp as well and I removed it from the file because it's a **horrible** way of doing things and it's much better to put your Common Lisp packages in `~/common-lisp` or another well-known location in ASDF (https://asdf.common-lisp.dev/asdf.html).

catap:
@MIvanchev yep, but this patch is required for OpenBSD ports.

catap:
See here: https://github.com/openbsd/ports/blob/master/x11/stumpwm/Makefile

MIvanchev:
Thanks, so the `DISTFILES` are unpacked into the build dir? Because it could be possible to specify in the `pre-xyz` step that they're unpacked to `<fake-home>/common-lisp` instead which is the correct place.

MIvanchev:
@catap So the ASDF docs say one could use `CL_SOURCE_REGISTRY` to specify the repository so maybe the Makefile needs to set just that to WRKDIST, worst case in conjuction with `SEPARATE_BUILD=yes` or something similar.

catap:
@MIvanchev indeed it make things cleaner.

PR #1222, opened 2025-01-28
Reworked IO loop, simplification & removing restarts.

### LINKED ISSUE #1204: High CPU usage
After updating (a lot of stuff: sbcl, hyperlisp, stumpwm..., I think the latter from 23.11 to master) the WM started using a lot of CPU time, around 80% of a single core. Sniffing around made me conclude it was most likely the io-loop, and while I don't know why yet, something causes selects to be called with no timeout (directly visible in strace). On a whim, I replaced `(values null null)` for when no timeouts are listed with `(values 0 100)` and got a much better behaved version, but it smells off (hence no PR) - if I add some printing to that part of the code, there's always an entry on the timeouts list. 

Mostly posting this, therefore, to help track things and to think out loud, maybe someone else has seen this behaviour and/or maybe I'm reading the code all wrong. 

```
$ git diff
diff --git a/ioloop.lisp b/ioloop.lisp
index 207df86..b553629 100644
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -287,7 +287,7 @@
                                                                   s-to-ms))
                                                         0)))
                                     (floor remaining-ms 1000000))
-                                  (values nil nil))))
+                                  (values 0 100))))
                        ;; Actually block for events
                        (multiple-value-bind (rval errno)
                            (multiple-value-call #'sb-unix:unix-fast-select

```

MIvanchev:
Fixes #1204.

### DIFF
--- a/ioloop.lisp
+++ b/ioloop.lisp
@@ -66,9 +66,8 @@
 ;;;; multiplexer to actually run it.
 
 (export '(io-channel-ioport io-channel-events io-channel-handle
-          *default-io-loop* *current-io-loop* *current-io-channel*
           io-loop io-loop-add io-loop-remove io-loop-update
-          callback-channel callback-channel-stream callback-channel-events))
+          *default-io-loop* *current-io-loop*))
 
 ;;; General interface
 (defgeneric io-channel-ioport (io-loop channel)
@@ -84,6 +83,7 @@
 
   An I/O channel may also return NIL to indicate that it is only
   interested in purely virtual events, such as :TIMEOUT or :LOOP."))
+
 (defgeneric io-channel-events (channel)
   (:documentation
    "Returns a list of events that CHANNEL is interested in. An event
@@ -118,6 +118,7 @@
   other point in time, it should use the IO-LOOP-UPDATE function to
   notify the I/O loop of such changes. The I/O loop may also update
   spuriously at any time, but such updates are not guaranteed."))
+
 (defgeneric io-channel-handle (channel event &key &allow-other-keys)
   (:documentation
    "Called by the I/O loop to notify a channel that an event has
@@ -129,12 +130,15 @@
 
 (defgeneric io-loop-add (io-loop channel)
   (:documentation "Add a channel to the given I/O multiplexer to be monitored."))
+
 (defgeneric io-loop-remove (io-loop channel)
   (:documentation "Unregister a channel from the I/O multiplexer."))
+
 (defgeneric io-loop-update (io-loop channel)
   (:documentation "Make the I/O loop update its knowledge of what
   events CHANNEL is interested in. See the documentation for
   IO-CHANNEL-EVENTS for more information."))
+
 (defgeneric io-loop (io-loop &key &allow-other-keys)
   (:documentation "Run the given I/O multiplexer, watching for events
   on any channels registered with it. IO-LOOP will return when it has
@@ -147,11 +151,6 @@
 (defvar *current-io-loop* nil
   "Dynamically bound to the I/O loop currently running, providing an
   easy way for event callbacks to register new channels.")
-
-(defvar *current-io-channel* nil
-  "While processing an I/O channel, this variable is dynamically bound
-  to the channel in question. This is provided primarily for
-  error-handling code.")
 
 ;; Default methods for the above
 (defmethod io-channel-handle (channel event &key &allow-other-keys)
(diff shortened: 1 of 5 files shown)


ioloop.lisp
```diff
@@ -66,9 +66,8 @@
 ;;;; multiplexer to actually run it.
 
 (export '(io-channel-ioport io-channel-events io-channel-handle
-          *default-io-loop* *current-io-loop* *current-io-channel*
           io-loop io-loop-add io-loop-remove io-loop-update
-          callback-channel callback-channel-stream callback-channel-events))
+          *default-io-loop* *current-io-loop*))
 
 ;;; General interface
 (defgeneric io-channel-ioport (io-loop channel)
```
sdilts, line 73: It looks like you changed the expected return type of this method? Can you update the documentation string if so?

sdilts (commented):
This looks great, although I admit to not fully understanding what the original code did or being very familiar with how IO loops work in general. This new code is much easier to understand and is much simpler, which I really like. If it solves the high IO usage problem, I'm all for it.

It's marked WIP. What else needs to be done on it?

I see that you remove several exported symbols/classes. I have no problem with that considering how esoteric adding IO channels seems to be and the severity of the problem. I also don't see a problem doing a major version bump for this change if @dmb2 disagrees. Was the removed class actually used anywhere in our codebase?

ioloop.lisp
```diff
@@ -193,197 +192,103 @@
...
+                                   (sb-alien:addr efds)
+                                   secs
+                                   usecs)
+              (declare (ignore count))
+              (if (and errno (plusp errno))
+                (unless (eql errno sb-unix:eintr)
+                  (dformat 1
+                           "Unexpected ~S error: ~A~%"
+                           'sb-unix:unix-fast-select
+                           (sb-int:strerror errno)))
```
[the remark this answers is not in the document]
MIvanchev: I dunno, it was in the original code so I just copied it over, but it seems a little bit logical. A signal is likely not an error.

ioloop.lisp
```diff
@@ -193,197 +192,103 @@
 (defmethod io-loop-update ((info sbcl-io-loop) channel)
   (declare (ignore info channel)))
 
-(defmethod io-loop ((info sbcl-io-loop) &key description)
+;;; Calculates the maximum blocking time that can be spend  waiting for
+;;; IO activity before channels interested in timeout events (i.e. timers)
+;;; need to be notified.
+(defun get-max-blocking-time (lowest-timeout)
```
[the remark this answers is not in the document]
MIvanchev: It's (likely) the return type of `get-internal-real-time` but I'm not sure if it's part of the standard.

ioloop.lisp
MIvanchev: Huh, what method?

MIvanchev:
Thanks a lot @sdilts, your review is invaluable!

> It's marked WIP. What else needs to be done on it?

Well, for us to have at least one discussion, I'm already using it for couple of hours without issues otherwise :)

ioloop.lisp
sdilts: A couple of implementations of this method changed, and I was wondering if there was an update to how this method is used that required a documentation update.

MIvanchev:
Hey took me a while to answer sorry!

So, @mtlll is right that some clarification would be great! First, StumpWM uses a [`select(2)`](https://linux.die.net/man/2/select) mechanism to efficiently block between X11 events. The documentation in the code clearly states that `epoll` or `kqueue` is preferred and indeed they are but there are no easy ways to use them from SBCL while select is usable through an [internal package](https://koji-kojiro.github.io/sb-docs/build/html/sb-unix/) of SBCL which we really shoudn't be using but we do. It's called `SB-UNIX:UNIX-FAST-SELECT`.

When using `select(2)` StumpWM waits on 2+ file descriptors being:

- X11's own file descriptor used to notify waiting listeners of events. This is called the display channel.
- StumpWM's request pipe file descriptor used to awaken the event loop from other threads if necessary. The is called the request channel.
- Some weird thing called a callback channel which is something the user can use...

How long `select(2)` blocks is calculated by taking the nearest point in time at which a StumpWM timer needs to be processed.

Why was the CPU spiking? Because the timeouts to `select(2)` were miscalculated so it never blocked anymore. The code was so chaotic I couldn't figure out why so I decided to reorganize it a bit.

About the removed restarts... they were intended to allow the user to "gracefully continue" in case something goes wild in the loop (an error occurs). "Gracefully continue" was understood as quitting the event loop or removing a channel the handler of which caused an error. This I think was nonsense because

a) errors in the event loop are catastrophic and should never happen,
b) a graphical session is active so there's no access to the terminal, StumpWM has a graphical facility for CL's debugger (popup selection menus) but with erroneous event loop there won't be any way to use it unless I'm missing something.

Actually I left 1 restart intact, https://github.com/stumpwm/stumpwm/blob/master/stumpwm.lisp#L190, the one just before the loop is launched. I don't believe it's useful.

@sdilts 
>Was the removed class actually used anywhere in our codebase?

No, I double checked, callback channels were only exported, not documented and not used internally. I truly believe nobody out there is using them because they're quite exotic. We'll see...

mtlll:
Thanks for elucidating. Now that you did, this change makes sense to me, and I support it, for whatever that's worth(not much :P)

The reliance on select(2) is interesting. Perhaps this would be a good time to address this? While SBCL doesn't expose epoll/kqueue, there is iolib(https://github.com/sionescu/iolib), which seems reasonably maintained. Doesn't support whatever the equivalent is on illumos, but it does support select(2), so it shouldn't break stumpwm on illumos anyway.

sdilts:
> The reliance on select(2) is interesting. Perhaps this would be a good time to address this? While SBCL doesn't expose epoll/kqueue, there is iolib(https://github.com/sionescu/iolib), which seems reasonably maintained. Doesn't support whatever the equivalent is on illumos, but it does support select(2), so it shouldn't break stumpwm on illumos anyway.

That's too much to ask right now, although I definitely think that it's worth investigating. `cl-async` might be the better choice due to the slightly odd C dependency that iolib has.

I think this PR ready to go, but I'll wait for @dmb2  to take a look.

mtlll:
I didn't know about cl-async. That looks better, although that also depends on a non-lisp library, libuv. Libuv is probably easier to install on various OS than libfixposix. And is pretty much guaranteed to be well maintained. And it supports illumos IOCP as well, unlike iolib.

Though I agree, it makes sense to wait. The reason select(2) replaced by epoll/kqueue/iocp in the first place was performance issues in things like web servers, when handling a huge amount of connections. We're just waiting on a handful of fds, should be fine, performance wise.

MIvanchev:
Yes, performance-wise I think we're good with `select(2)`. According to https://nrk.neocities.org/articles/x11-timeout-with-xsyncalarm there seems to be a correctness issue and one needs to call pending before block-waiting but I don't quite understand the point yet. Maybe there's nothing to it.

I hope with time we can refactor more of StumpWM.

dmb2:
This looks much better, thank you for your hard work! This new code is much easier to follow.

Happy to YOLO this into main.

fstamour:
> b) a graphical session is active so there's no access to the terminal, StumpWM has a graphical facility for CL's debugger (popup selection menus) but with erroneous event loop there won't be any way to use it unless I'm missing something.

If one is working on the event loop itself, they probably have swank running, which means they can connect to it from a virtual terminal.

I've done that a few times (connect from a VT).

I don't think I've ever had to use those restarts, but I can see why someone would want them.

PR #1224, opened 2025-01-31
stumpwm.asd: Bump :version to 24.11

DerGuteMoritz:
This was apparently forgotten when tagging 24.11 which means the `version` command would still show `23.11` when not building from git (as is the case with the nixpkgs package).

Is there some script / task for tagging releases, perhaps? Then we could automate the bump to prevent it from happening again in the future.

### DIFF
--- a/stumpwm.asd
+++ b/stumpwm.asd
@@ -7,7 +7,7 @@
 (defsystem :stumpwm
   :name "StumpWM"
   :author "Shawn Betts <sabetts@vcn.bc.ca>"
-  :version "23.11"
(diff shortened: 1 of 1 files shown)


dmb2:
Hi,

There used to be, but lately I've been doing it through the github gui which doesn't trigger the old tag make scripts.  I will try to remember in the future.

Thanks for taking the time to submit a PR, @catap beat you to it and I try to merge roughly chronologically if I can.  Sorry if this has been messing up anything related to packaging stumpwm for your distro.

PR #1227, opened 2025-02-23
Add a setf function for window-property, also FIX issue #997

Luminous-99:
As far as i could find quoteleft and quoteright have no relation with the symbols < and > , the bepo layout can output ` and ’ which are already defined as grave and apostrophe.


dmb2:
Please separate these into two separate PRs, the fix for #997 is uncontroversial.  What is the purpose for window-property and why does it need to be exported?

PR #1228, opened 2025-02-24
Use non-canonical paths when resolving default data dir

### LINKED ISSUE #1225: init.lisp cannot be a symlink?
I'm using guix home, and ~/.stumpwm/init.lisp is a symlink into a read-only filesystem.
When upgrading from 24.11 to master to get the "rework-io-loop" branch and reduce CPU usage, my system no longer starts.
Replacing the symlink file with a regular file solves the issue.

Could this be related to https://github.com/stumpwm/stumpwm/commit/019e44a110fff6d600802901c30d5be7a16da8d0 ?

Anyone else using guix home who can verify this?

anticomputer:
Resolving the data dir path based on probe-file's canonical path lookup of pathspec breaks users that manage their StumpWM configuration through symlinked dotfiles. The data dir would incorrectly be set to wherever the root of the symlinked file is located on the file system. This may be a read only location on e.g. systems such as Guix and NixOS, or just some random location in the case of stow. StumpWM would then incorrectly treat this random location as its data directory. This patch ensures we operate on the pathspec argument and not the truename result from probe-file.

Fixes #1225

### DIFF
--- a/primitives.lisp
+++ b/primitives.lisp
@@ -1544,9 +1544,12 @@ sync-all-frame-windows to see the change.")
 (defun default-data-dir ()
   "Return the default data dir pathname based on the loaded StumpWM configuration file."
   (let ((rc-file (or
-                  (probe-file (merge-pathnames #p".stumpwmrc" (user-homedir-pathname)))
-                  (probe-file (merge-pathnames #p".stumpwm.d/init.lisp" (user-homedir-pathname)))
-                  (probe-file (uiop:xdg-config-home #p"stumpwm/config")))))
+                  (let ((pathspec (merge-pathnames #p".stumpwmrc" (user-homedir-pathname))))
+                    (and (probe-file pathspec) pathspec))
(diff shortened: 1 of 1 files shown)


simendsjo:
Should we add someone as reviewers or ping someone to get some eyes on this?

dmb2:
I'm not entirely sure this is the correct solution. It sounds like some systems set the data-dir to a read only area but stumpwm assumes it can write to the .stumpwm.d directory. 

If a user is symlinking their init files to manage them, its not clear to me what stumpwm should do in that case.  Is there something in the XDG spec we can refer to for guidance?

sdilts:
@dmb2, here's how I understand it. It probably has the details wrong, but here goes:

The problem is that the actual file that they want read is in `/some/location/` that is read-only. However, they have a symlink in `other/location` in a place that is writable, which points to the first file. `probe-file` returns the canonical path, which gives you the first location, when you wanted the second one. The code then assumes that the first incorrect location is where the data files are located, which is incorrect.

Or it might be the other way around or involve relative paths, I'm not sure. Either way, `probe-file` is transforming the path to something it shouldn't, and this avoids that.

PR #1232, opened 2025-03-17
Add a setf function for window-property

Luminous-99:
I found xlib:change-property too inconvinient to use so i made this abstraction over it


Luminous-99:
One of my uses is to add a window property called _STUMPWM_FLOATING when calling float-this, with value 1 for floating or 0 for tiling.
this way i can add effects to the windows with picom like this:
```javascript
rules = ({
    match = "_STUMPWM_FLOATING@ = 1";
    corner-radius = 10;
})
```
So a simple setf  seems nicer than searching CLX for change-property to then try to make sense of how x11 handles window properties (which is what happened for me).

dmb2:
Happy to merge this in contrib as its own module, but I won't be including it in the main stumpwm api.

PR #1241, opened 2025-04-29
Add shutdown, suspend, etc. from wishlist

BaalHadad:
Made a new file for them since they didn't really fit in with any other file.
These functions don't require root access.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,6 @@ StumpWM universe (in no particular order):
   is involved)
 * Wallpapers! (support pulling from remote sources, changing based on
   timers, and other hacky features)
-* Shutdown, restart, suspend, and hibernate functions that don't
-  require root access
 * Revamped, mouse-friendly mode-line. 
   * Support fixed number of chars for window titles
   * Dynamically trim window titles to fit them all on the mode-line
--- a/poweroff.lisp
+++ b/poweroff.lisp
@@ -1,0 +1,80 @@
+(in-package :stumpwm)
+
+(export '(shutdown cancel-shutdown reboot halt suspend hibernate
+          *alternative-suspend* *alternative-hibernate*))
+
+(defun path-directories (&optional (name "PATH"))
+  (let ((path-dirs (uiop:split-string (uiop:getenv name) :separator '(#\:))))
+    (mapcar (lambda (x) (uiop:ensure-pathname x :ensure-directory t)) path-dirs)))
+
+(defun file-in-path-p (name)
+  (dolist (dir (path-directories))
+    (when (probe-file (merge-pathnames dir name))
+      (return dir))))
+
+(defcommand shutdown (&optional (now nil) (time 30)) ((:y-or-n "Now? "))
+  "Poweroff the system. If NOW is not NIL poweroff immediately, otherwise asks for a date."
+  #+(or bsd linux)
+  (let* ((time (if (and %interactivep% (not now))
+                   (read-one-line (current-screen) "When? ")
+                   (if (and %interactivep% now) "" time)))
+         (cmd (format nil "shutdown ~a ~a" time (if now "now" ""))))
+    (run-shell-command cmd)))
+
+(defcommand cancel-shutdown () ()
+  "Cancels any shutdown process."
+  #+(or bsd linux) (run-shell-command (format nil "shutdown -c")))
+
+(defcommand reboot (&optional (now nil) (time 30)) ((:y-or-n "Now? "))
+  "Reboot the system. If NOW is not NIL reboot immediately, otherwise asks for a date."
+  #+(or bsd linux)
+  (let* ((time (if (and %interactivep% (not now))
+                   (read-one-line (current-screen) "When? ")
+                   (if (and %interactivep% now) "" time)))
+         (cmd (format nil "shutdown -r ~a ~a" time (if now "now" ""))))
+    (run-shell-command cmd)))
+
+(defcommand halt (&optional (now nil) (time 30)) ((:y-or-n "Now? "))
+  "Halt the system. If NOW is not NIL halt immediately, otherwise asks for a date."
+  #+(or bsd linux)
+  (let* ((time (if (and %interactivep% (not now))
+                   (read-one-line (current-screen) "When? ")
+                   (if (and %interactivep% now) "" time)))
+         (cmd (format nil "shutdown -H ~a ~a" time (if now "now" ""))))
+    (run-shell-command cmd)))
+
+(declaim (type (or function string null) *alternative-suspend*))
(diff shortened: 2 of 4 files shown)


Junker:
I notice a direct connection to systemd here. In certain Linux distributions that do not use systemd, such as Void, Alpine, and Guix, it will not function properly. I don't believe it is a good idea to tie StumpWM to systemd.

Luminous-99:
good point, i added variables for this use case so that the user can change their behavior.

MIvanchev:
Can confirm as a Void user, this is not how we do stuff but also what is the point of these commands that anyone can quickly define for themselves? I.e. from my config

```
(defcommand shutdown () ()
  (play-sound :shutting-down)
  (run-async "sudo poweroff"))

(defcommand reboot () ()
  (run-async "sudo reboot"))

(defcommand suspend () ()
  (run-async "sudo zzz"))
```

BaalHadad:
IIRC When i made the PR i couldnt find a portable way do to suspend and hibernate, ill try to see if a better way is possible

BaalHadad:
I changed it to also check for the kernel interface (/sys/power/state) for suspend/hibernate but stumpwm needs to run as root; as long as it runs as root it should work.

BaalHadad:
@MIvanchev thanks for mentioning Void; it helped a lot, i mostly translated their zzz shell script to lisp

MIvanchev:
@Luminous-99 I really think you're investing a lot of good research and effort into a non-existing problem and its just wasteful. The wishlist item

```
Shutdown, restart, suspend, and hibernate functions that don't require root access
```

is absolute nonsense that no StumpWM user ever stumbles upon and should be removed from the README ASAP. Half of the wishlist is nonsense to be honest. We can dedicate the time for more significant issues like non-blocking modeline updates etc. A very relevant thing is

```
Float windows within parent applications (specifically dialogs in gimp or firefox).
```

which affects tons of programs (anything opening a file selection dialog).

BaalHadad:
@MIvanchev 

> @Luminous-99 I really think you're investing a lot of good research and effort into a non-existing problem and its just wasteful. The wishlist item
> 
> ```
> Shutdown, restart, suspend, and hibernate functions that don't require root access
> ```
> 
> is absolute nonsense that no StumpWM user ever stumbles upon and should be removed from the README ASAP. Half of the wishlist is nonsense to be honest. We can dedicate the time for more significant issues like non-blocking modeline updates etc. A very relevant thing is

i would assume most people end up writing their own functions/commands for poweroff functions, i believe it's more convenient to have zero-trouble built-in functions to avoid wasting time like that, if they don't work on some systems users can submit a PR to permanently fix it and save others time instead of keeping the fix solely in their config

> ```
> Float windows within parent applications (specifically dialogs in gimp or firefox).
> ```
> 
> which affects tons of programs (anything opening a file selection dialog).

this is very important too, too many programs end up breaking because of this problem. realistically it should not be hard to solve since basically every WM/DE implements it

BaalHadad:
Also other significant issues are the mode-line and clx-truetype, the mode-line is mostly fine for it's purpose but it could really use a performance update (for example multiple threads for formatters and :evals).

As for clx-truetype it's simply too slow when rendering text, when i open the help message window it takes around 1-2 seconds to render anything despite not having much to render, last time i profiled it with sbcl the main offender seemed to be a function involving iterators and bezier curves (or something like that), there has to be a more performant way.

MIvanchev:
> i believe it's more convenient to have zero-trouble built-in functions to avoid wasting time like that, if they don't work on some systems users can submit a PR to permanently fix it and save others time instead of keeping the fix solely in their config

OK, yes, I agree with you, however I would at least soften the requirement of a sudo or at least show a message if it's not possible to execute the command due to permissions. I appreciate your enthusiasm for StumpWM a lot, btw, thank you!

> As for clx-truetype it's simply too slow when rendering text..

You might wanna try `sdl-fonts` which could be a bit faster, however the help window does take a lot of layouting so on my PC it's also ~500ms. We could try to streamline the code a bit there too!

BaalHadad:
> OK, yes, I agree with you, however I would at least soften the requirement of a sudo or at least show a message if it's not possible to execute the command due to permissions. 

 yeah the permissions are the biggest problem, i really dont know how to get around that besides building a separate executable alongside stumpwm to call sudo on. ill add that message, makes sense.
 
 > I appreciate your enthusiasm for StumpWM a lot, btw, thank you!
 
 Thanks! it's been my daily driver for a year so i spend a lot of time adding/fixing stuff (my config has a lot of features (like a simple window decorator) where i'm not sure if i should push some as PRs)

> You might wanna try `sdl-fonts` which could be a bit faster, however the help window does take a lot of layouting so on my PC it's also ~500ms. We could try to streamline the code a bit there too!

rendering message strings chunk by chunk might be better, at least for responsiveness.

dmb2:
I added the 'shutdown, restart, and suspend' to the wishlist because stumpwm is a generally disorienting experience. I was hoping that there was a standard way to do this, but seeing the implementations, I feel like this would be best as a small contrib module. 

There's probably not a non-systemd non-root way to do this.

Anything goes in contrib.

sdilts:
> I don't want to pollute the codebase with OS specific flags.

I was planning to have "official" contribs for stuff like this in Mahogany. The idea would be to put it in an asdf system on an installed path (like `/usr/share/mahogany/`), and configure asdf on startup to look for systems in that directory. That way, it would be included in the default distribution but not loaded by default for the people who don't need it.

> There's probably not a non-systemd non-root way to do this.

~~There is, but it still involves systemd 😆. Systemd controls sessions via the [logind daemon](https://www.freedesktop.org/wiki/Software/systemd/logind/), which accepts commands via dbus. There's a non-systemd implementation of it called [elogind](https://github.com/elogind/elogind) that runs on other systems.~~

Sorry, I misremembered. It's [`seatd`](https://git.sr.ht/~kennylevinsen/seatd), which integrates with different init systems. You need to bind the libseat C library and use its functions.

sdilts:
I created #1251 to discuss system-specific things like this. I also created #1252 to look at a generic way to do session management.

Since this discussion in the context of a PR has derailed a bit, I'm closing this; we can resume the  relevant discussion(s) in one of those two issues. @Luminous-99, it might be worth creating an issue talking about slow font rendering; aside from sdl, there are lisp bindings to the cairo library that might also work.

Some code that accomplishes what is in the PR is [already in the contrib repo](https://github.com/stumpwm/stumpwm-contrib/tree/master/util/end-session). It's missing hibernate and a bit wonky in other places.

Apparently I wrote it, but I don't remember doing so, and don't currently use it in my setup. I would have piped up sooner if I remembered, sorry.

PR #1245, opened 2025-07-31
add window classes for input, message and mode-line windows

BaalHadad:
i saw that [https://github.com/stumpwm/stumpwm/discussions/1105#discussioncomment-6017456](url) requested this 
so i decided to add them

### DIFF
--- a/mode-line.lisp
+++ b/mode-line.lisp
@@ -23,6 +23,7 @@
           *mode-line-border-color*
           *mode-line-border-width*
           *mode-line-foreground-color*
+          *mode-line-window-class*
           *mode-line-pad-x*
           *mode-line-pad-y*
           *mode-line-position*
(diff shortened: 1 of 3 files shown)


dmb2:
Hi,

I'm getting a 404 for your link.  Without additional context, its hard to understand what exactly this is trying to address beyond exporting some symbols which I am wary about doing without strong justification.

BaalHadad:
For some reason the link broke, its this one: https://github.com/stumpwm/stumpwm/discussions/1105#discussioncomment-6017456

dmb2:
Can you help me understand what this does exactly? The post you linked seems to just state its annoying to have to write something for their config.  I'm not understanding what the advantage of this PR is over having add some lines to your config for specific compositing issues.

BaalHadad:
> I wonder if we can add WM_CLASS by default for modeline, message windows, and other windows made by StumpWM itself, so it is easier for users to add compositor exclude rules for these windows.

is what they said, since the message windows don't have a WM_CLASS or other discernible window properties currently compositors can't detect them to add some visual effects (stuff like corner rounding or transitions)

MIvanchev:
I think this is a useful addition. Being able to interact with those windows has many potential advantages.

PR #1246, opened 2025-08-12
Added more mouse interactivity and support for fixed number of chars for window titles in the mode line

BaalHadad:


dmb2:
While the code is relatively straightforward, next time a description of what you are implementing would help a lot.

PR #1255, opened 2026-01-20
Revised the poweroff functions from the prior PR

BaalHadad:
For Linux instead of systemctl only logind is used, if either logind or elogind is available it should work fine, if neither exists the kernel interface will be used as long as stumpwm runs as root  instead, otherwise an error message will be shown.

For BSD ConsoleKit2 is used and since the interface is identical with logind the logic is the same.

This is a good compromise from my point of view but i'd appreciate more suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,8 +179,6 @@ StumpWM universe (in no particular order):
   is involved)
 * Wallpapers! (support pulling from remote sources, changing based on
   timers, and other hacky features)
-* Shutdown, restart, suspend, and hibernate functions that don't
-  require root access
 * Revamped, mouse-friendly mode-line. 
   * Support fixed number of chars for window titles
   * Dynamically trim window titles to fit them all on the mode-line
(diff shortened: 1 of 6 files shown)


BaalHadad:
very annoying to have a build error out of any other kind of error

BaalHadad:
finally fixed it

sdilts:
>  I remember thinking shutdown, reboot, and hibernate were something we should support. Now I feel like its something no window manager should bother with.

This is a weird one. It absolutely makes sense to be able to do these things via a key binding and have stumpwm-native prompts to do it, but triggering these things can be a bit non-standard and you'd think there would be a tool floating around already that handles all the details.

Does Stumpwm need to do anything so that it exits cleanly? For example, save the locations of all the open windows? If so, it would make sense to have these commands built in so that the appropriate shutdown hooks are ran before the shutdown sequence is triggered; I don't think we can guarantee the state of the system otherwise. This is why I'd lean towards including shutdown and reboot, especially because we can use the shutdown command.

Other than that, I don't have much to add.

Issue #1264, opened 2026-08-22
Sticky windows leave orphaned records in other groups after closing.


yvm:
`$ sbcl --version`
> SBCL 2.6.7.167-023755eb0

`STUMPWM> *version*`
> "24.11-39-g49b9779 Compiled On Fri Aug 21 2026 22:51:19"

Steps to reproduce:
```
(gnew-float "test")
(run-commands "exec xclock")
(toggle-always-show)
; at this step vgroups command shows xclock window in each group as expected
(run-commands "exec killall xclock")
; xclock window removed only from current-gtoup
```

Suggested fix:
```
diff --git a/window.lisp b/window.lisp
index b9551be..433ebd7 100644
--- a/window.lisp
+++ b/window.lisp
@@ -892,6 +892,8 @@ needed."
 (defun withdraw-window (window)
   "Withdrawing a window means just putting it in a list til we get a destroy event."
   (declare (type window window))
+  (if (find window *always-show-windows*)
+      (toggle-always-show window))
   ;; This function cannot request info about WINDOW from the xserver as it may not exist anymore.
   (let ((group (window-group window))
         (screen (window-screen window)))

```

sdilts:
Closed by #1265.

PR #1265, opened 2026-08-22
Unstick sticky windows before removing to prevent orphaned records

yvm:
Issue #1264


sdilts:
Thanks! When looking into this more, I found a call to `(xlib:change-property (window-xwin window) :_NET_WM_DESKTOP)`, which could mean changing a property on a non-existent window.

Have you been running with this patch for a while? I'm not sure if this is a problem or not.

yvm:
You're right. Updated the fix.
I'm still experiencing an issue with always-show'n window: it causes focus-next-window to cycle only between two windows instead of the whole group. Since this seems unrelated to the current topic, I can move it to a new issue if needed.

sdilts:
Thanks!

sdilts:
> I'm still experiencing an issue with always-show'n window: it causes focus-next-window to cycle only between two windows instead of the whole group

Ya, that sounds like a different issue.

